PR #42, opened 2026-04-03
fix(ui): prevent bug report dialog freeze when gh CLI is missing

carmelp16:
openGitHubIssue returned nil when gh was not installed, leaving submitting=true permanently and freezing the dialog. Return a proper bugReportOpenErrMsg instead, and clear submitting on error so the dialog stays usable.

## Summary

When the `gh` CLI is not installed, pressing Enter in the bug report dialog sets `submitting = true` but `openGitHubIssue` returns `nil` - no `tea.Cmd` runs, no message is sent back, and the dialog is permanently frozen (can't type, can't submit, only Esc works).

**Two fixes:**
- **`bugreport.go`**: `openGitHubIssue` now returns a `bugReportOpenErrMsg` when `gh` is missing instead of `nil`
- **`app.go`**: the `bugReportOpenErrMsg` handler now clears `submitting` so the dialog unfreezes (also fixes the case where `gh` exists but the API call fails mid-flight)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist

- [ ] Documentation updated (if applicable)
- [ ] `CHANGELOG.md` updated under `[Unreleased]` (if user-facing change)

## Test plan
- [x] \`TestBugReportDialog_EnterWithGhMissing_ReturnsCmd\` — core regression test
- [x] \`TestBugReportDialog_EnterWithEmptyDesc_Noop\` — empty description does nothing
- [x] \`TestBugReportDialog_EnterWhileSubmitting_Noop\` — double-submit guard
- [x] \`TestBugReportDialog_Esc_Hides\` — Esc always works, even while submitting

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Bug report dialog freezing permanently when `gh` CLI is not installed
 - Session stuck at "waiting" status after user interrupts/escapes a permission prompt
 
 ## [1.1.0] - 2026-03-21
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -332,6 +332,7 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return h, nil
 
 	case bugReportOpenErrMsg:
+		h.bugReport.submitting = false
 		h.setError(msg.err)
 		return h, nil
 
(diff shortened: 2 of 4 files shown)


hayke102:
@carmelp16  Thanks for catching this! Clean fix with solid test coverage. 
After you patch in that changelog, I will merge this 🚀

carmelp16:
@hayke102 There you go my friend ❤️‍🔥

PR #45, opened 2026-04-08
fix(ui): clean up temp file on bug report write failure

carmelp16:
## Summary

- Fix temp file leak in `openGitHubIssue` when `WriteString` fails: `defer os.Remove()` was registered after the write check, so an early return would skip cleanup. Moved it to right after `CreateTemp` succeeds.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist

- [ ] Documentation updated (if applicable)
- [x] `CHANGELOG.md` updated under `[Unreleased]` (if user-facing change)

## Test plan
- [x] Code review — the fix is a one-line move, easy to verify by inspection

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Session stuck at "waiting" status after user interrupts/escapes a permission prompt
 - Agent team sessions showing idle/running instead of waiting when sub-agent needs approval
 - Status oscillating between idle and finished when stale waiting hook is present
+- Temp file not cleaned up when bug report body write fails
 
 ## [1.1.0] - 2026-03-21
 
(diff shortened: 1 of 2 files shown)


carmelp16:
@hayke102 Another one at your service 🫡

hayke102:
Hey @carmelp16, great find on the temp file leak! The fix is correct.

There's a merge conflict on CHANGELOG.md because we recently switched to a fragment-based changelog. Instead of editing CHANGELOG.md directly, add a fragment file:

```
changelog/unreleased/fix-bugreport-tempfile.md
```
```yaml
---
type: fixed
---
Temp file not cleaned up when bug report body write fails
```

Rebase on master, drop the CHANGELOG.md change, and add the fragment instead. See `changelog/README.md` for details. CI will check for it automatically.

denniszag (approved):

PR #69, opened 2026-05-06
feat(ui): add PgUp/PgDn navigation in sidebar

hayke102:
## Summary
- Adds `PgUp` / `PgDn` to jump the sidebar cursor by a full page; clamps at list bounds (no wrap), matching `j`/`k` semantics.
- Fixes a viewport bug where the highlighted row could fall outside the rendered window at the list edges — `syncViewport` was ignoring the rows reserved for top/bottom scroll indicators in `RenderSidebar`. Extracted `sidebarVisibleRows()` so PgUp/PgDn page size and viewport math share the same height.
- Registers both keys in `allKeyBindings` so they show up in the `?` help overlay (no bottom-bar slot, to keep the bar uncluttered).

## Test plan
- [ ] Build and run with enough sessions to overflow the sidebar.
- [ ] Press `PgDn` repeatedly: cursor jumps by ~one page, stops at the last item and remains highlighted.
- [ ] Press `PgUp` repeatedly: cursor jumps back, stops at the first item.
- [ ] Hold `j` past the bottom: highlighted row stays inside the visible window (was previously slipping off-screen by 1 row).
- [ ] Open `?` overlay: `PgUp` / `PgDn` listed in Navigation.
- [ ] Resize terminal small enough that visible rows clamp to 1: PgDn moves by 1 with no panic.

## Summary by CodeRabbit

* **New Features**
  * PgUp/PgDn now move the sidebar by a full page, clamping selection to valid items and keeping the cursor visible; preview refreshes for the newly selected item.
  * Sidebar sizing logic improved to compute stable visible rows so page navigation behaves consistently across window sizes.

* **Documentation**
  * Help overlay updated with Page Up / Page Down entries; unreleased changelog added.

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -2684,6 +2684,26 @@ func (h *Home) rebuildSessionMap() {
 	}
 }
 
+// sidebarListHeight returns the height of the sidebar panel in the current
+// layout — i.e. the value View() passes to RenderSidebar as `height`, before
+// chrome rows (title + underline) and scroll indicators are subtracted.
+// Stacked mode gives the sidebar ~55% of the content area; single/dual give
+// it the full content area. Mirrors the arithmetic in View() (app.go:794+).
+func (h *Home) sidebarListHeight() int {
+	contentHeight := h.height - 2 - helpBarHeight
+	if contentHeight < 1 {
+		contentHeight = 1
+	}
+	if h.layoutMode() == "stacked" {
+		sh := (contentHeight * 55) / 100
+		if sh < 3 {
+			sh = 3
+		}
+		return sh
+	}
+	return contentHeight
+}
+
 // sidebarMinVisibleRows is the conservative lower bound on visible session rows
 // in the sidebar — it assumes both scroll indicators are drawn, so the value
 // holds even mid-scroll. Used by syncViewport to anchor the cursor before
@@ -2691,7 +2711,7 @@ func (h *Home) rebuildSessionMap() {
 // Subtracts panel chrome (title + underline = 2) and reserves 2 rows for the
 // indicators RenderSidebar may add at top/bottom.
 func (h *Home) sidebarMinVisibleRows() int {
-	v := h.height - 2 - helpBarHeight - 4
+	v := h.sidebarListHeight() - 4
 	if v < 1 {
 		v = 1
 	}
@@ -2703,7 +2723,7 @@ func (h *Home) sidebarMinVisibleRows() int {
 // page-jump moves a full panel; syncViewport handles anchoring if the target
 // would otherwise sit on an indicator row.
 func (h *Home) sidebarPanelRows() int {
-	v := h.height - 2 - helpBarHeight - 2
+	v := h.sidebarListHeight() - 2
 	if v < 1 {
 		v = 1
 	}
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -1020,7 +1020,7 @@ func (h *Home) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		h.syncViewport()
 		return h, h.fetchPreviewForSelected()
 	case "pgdown":
-		target := h.cursor + h.sidebarVisibleRows()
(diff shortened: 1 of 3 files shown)


internal/ui/app.go
```diff
@@ -2665,6 +2684,18 @@ func (h *Home) rebuildSessionMap() {
 	}
 }
 
+// sidebarVisibleRows returns the number of session rows visible in the sidebar.
+// Subtracts the panel chrome (title + underline = 2) and reserves 2 rows for
+// the scroll indicators RenderSidebar (sidebar.go) may add at top/bottom — so
+// the cursor stays in view even at the list edges.
+func (h *Home) sidebarVisibleRows() int {
+	v := h.height - 2 - helpBarHeight - 4
```
hayke102: 🤔 reserve here is always 2, but `RenderSidebar` only draws indicators when scrolling (0 when the list fits, 1 at edges, 2 mid-scroll). Means pgdn is 2 rows short when the list fits. Intentional?

internal/ui/app.go
```diff
@@ -2665,6 +2684,18 @@ func (h *Home) rebuildSessionMap() {
 	}
 }
 
+// sidebarVisibleRows returns the number of session rows visible in the sidebar.
+// Subtracts the panel chrome (title + underline = 2) and reserves 2 rows for
+// the scroll indicators RenderSidebar (sidebar.go) may add at top/bottom — so
+// the cursor stays in view even at the list edges.
+func (h *Home) sidebarVisibleRows() int {
```
hayke102: Name reads like "current visible rows" but returns the lower bound (assumes both indicators are on). Maybe `sidebarPageStep` or `sidebarMinVisibleRows` makes the conservativeness obvious?

internal/ui/app.go
hayke102: Fixed — split helpers. PgUp/PgDn now use sidebarPanelRows (no indicator reserve), so a page jump moves a full panel even when the list fits. syncViewport keeps the worst-case anchor via sidebarMinVisibleRows.

internal/ui/app.go
hayke102: Fixed — renamed to sidebarMinVisibleRows and updated the doc comment to lead with the conservative lower-bound nature.

internal/ui/app.go
```diff
@@ -1019,6 +1019,25 @@ func (h *Home) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
...
+		h.syncViewport()
+		return h, h.fetchPreviewForSelected()
+	case "pgup":
+		target := h.cursor - h.sidebarPanelRows()
+		if target < 0 {
+			target = 0
+		}
+		h.cursor = target
+		h.syncViewport()
+		return h, h.fetchPreviewForSelected()
```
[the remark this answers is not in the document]
hayke102: Repo headers are intentionally selectable — see the doc comment on NextSelectableItem in sidebar.go ("repo headers are now selectable") and the enter handler at app.go:1041 which expects the cursor to land on a header to toggle the group. j/k already lands on headers; PgUp/PgDn matching that is the correct behavior, not a regression. Skipping the proposed fix.

internal/ui/app.go
```diff
@@ -2665,6 +2684,32 @@ func (h *Home) rebuildSessionMap() {
...
+// scroll indicators are drawn. Used as the PgUp/PgDn page step so a single
+// page-jump moves a full panel; syncViewport handles anchoring if the target
+// would otherwise sit on an indicator row.
+func (h *Home) sidebarPanelRows() int {
+	v := h.height - 2 - helpBarHeight - 2
+	if v < 1 {
+		v = 1
+	}
+	return v
+}
```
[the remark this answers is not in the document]
hayke102: Fixed — extracted sidebarListHeight() mirroring View()'s per-layout sidebar height (stacked: 55%, clamped >=3; single/dual: full contentHeight). sidebarMinVisibleRows and sidebarPanelRows now subtract chrome from that, so PgUp/PgDn and syncViewport behave correctly in stacked mode.

PR #80, opened 2026-05-23
fix(git): default worktree base to origin/<branch>

hayke102:
## Summary
- `GetDefaultBranch` now returns `origin/<branch>` (e.g. `origin/master`) when an origin remote is configured, instead of the local branch name.
- The worktree picker dialog (`w` key) pre-fills the base-branch input with this value, so `git worktree add -b <new> <base>` starts the new branch from the remote-tracking ref rather than a possibly-stale local branch.
- Local-only fallback (no `origin`) still returns plain `main`/`master` since there's no remote-tracking ref to use.

Note: nothing here runs `git fetch`, so `origin/master` is only as fresh as the last fetch.

## Test plan
- [ ] Open worktree dialog (`w`) on a repo with an `origin` remote → base branch input pre-fills `origin/master` (or `origin/main`)
- [ ] Submit dialog → new worktree branch starts at `origin/<branch>` commit
- [ ] On a local-only repo (no `origin`) → base branch input pre-fills `main` or `master` as before


hayke102:
Superseded by #82, which contains only the git.go change on a clean branch off master. PR #80 accidentally bundled the unrelated Codex commit (`e32f201`) that belongs to #78.

PR #121, opened 2026-06-07
fix(workspace): stop leftover dev processes before removing a worktree

hayke102:
## Problem

Deleting a git worktree sometimes fails with `git worktree remove: ... 'Directory not empty'` and leaves an orphaned directory that fleet has *forgotten*.

Root cause (verified with `lsof` on a real orphan): it's **not** a dirty repo (`--force` already handles that). It's **live dev-stack daemons from the session that outlived the tmux kill** — `process-compose` (cwd = the worktree), `air` (Go hot-reload, holding open fds), and `node` (vite/esbuild) keep re-creating files faster than git can delete them, so the final `rmdir` hits `ENOTEMPTY`. They detached from the tmux pane's process group, so killing the tmux session didn't stop them.

It also silently orphaned: the session row is deleted from SQLite immediately, then `Destroy` runs later in the background; on failure the user only got a 5s toast while the worktree dir + git registration remained — so leftovers piled up.

## What changed

**A — kill holders before removing** (`internal/proc`, new)
- `FindHolders` runs one fast system-wide `lsof -Fpcn` (~0.5s, no `+D` tree-walk) and prefix-matches; `Kill` does SIGTERM → 2s grace → SIGKILL.
- **Denylist spares editors, language servers (gopls/terraform-ls), shells, tmux, ssh, and fleet itself** (plus the configured editor) — so it kills the dev stack, not your VS Code / shell.
- `destroyWorktree` (app.go) kills holders then removes, wired into both the finalize path and the empty-worktree path.

**B — no silent orphans**
- `GitWorktreeProvider.Destroy` falls back to `os.RemoveAll` + `git worktree prune` (package mutex serializes concurrent removes of the same main repo).
- If it still fails, the worktree is **re-pinned + flagged `✕ removal failed — d to retry`** in the sidebar with an error naming the blockers; `d` retries.

**C — pre-flight warnings in the delete dialog**
- Instant (cache): "Uncommitted changes will be discarded", "N session(s) here are still running".
- Async (gen-guarded `ConfirmDialog.StartScan`/`SetScan`): "Will terminate N process(es): process-compose, air, node".

## Testing
- `make build` ✓, full `go test ./...` with `-race` ✓
- New `internal/proc` unit tests cover the lsof parse, prefix boundary, self-skip, and denylist
- gofmt clean

## Summary by CodeRabbit

* **New Features**
  * Worktree removal now detects and terminates blocking processes before deletion
  * Confirmation dialog displays which processes will be terminated
  * Failed removals can be retried with visual indicators in the sidebar

* **Bug Fixes**
  * Fixed "Directory not empty" errors during worktree deletion

* **Documentation**
  * Added documentation on the enhanced worktree deletion process


internal/ui/app.go
```diff
@@ -888,11 +901,22 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				break
 			}
 		}
-		if msg.workspaceErr != nil {
+		if msg.repoPath != "" {
```
hayke102: **Routing key is too broad — normal session deletes flow through the worktree-failure handler.**

`finalizeDelete` always sets `deleteCleanupDoneMsg.repoPath` (it copies `pd.RepoPath`, which `deferDelete` populates unconditionally at app.go:2954), regardless of `DestroyWS`. So this `if msg.repoPath != ""` branch is taken for *every* session delete, not just worktree destroys.

Usually harmless, but: a worktree whose removal previously failed has `failedWorktreeRemovals[wt]=true` and the `✕ removal failed` marker. The dir still exists, so the user starts a new session in it and later deletes that session normally → `handleWorktreeDestroyResult` hits the `workspaceErr == nil` branch and runs `delete(failedWorktreeRemovals[wt])` + rebuild, silently dropping the retry marker even though the worktree was never removed.

Suggest gating on an explicit flag (e.g. a `destroyAttempted bool` on the msg, set only when `destroyWorktree` actually ran) rather than `repoPath != ""`.

internal/ui/app.go
```diff
@@ -2987,13 +3062,98 @@ func (h *Home) finalizeDelete(pd PendingDelete) tea.Cmd {
...
+}
+
+// handleWorktreeDestroyResult applies part B. On a failed worktree destroy it
+// re-pins the worktree (so its header reappears as an empty checkout to retry
+// with d) and flags it for a persistent sidebar marker + error; on success it
+// clears any prior failure flag.
+func (h *Home) handleWorktreeDestroyResult(msg deleteCleanupDoneMsg) {
+	if msg.workspaceErr == nil {
+		if h.failedWorktreeRemovals[msg.repoPath] {
+			delete(h.failedWorktreeRemovals, msg.repoPath)
```
hayke102, line 3164: **Fan-out race can clear the failure flag set by a failed destroy.**

A worktree with N≥2 sessions deletes via N concurrent `finalizeDelete` goroutines; only the last carries `DestroyWS`. All N emit `deleteCleanupDoneMsg{repoPath: wt}` — the N-1 non-destroy siblings with `workspaceErr == nil`. If the last (destroy) fails it sets `failedWorktreeRemovals[wt]=true`, but goroutine completion order is nondeterministic: a sibling's nil-error message processed *after* the failure reaches this `delete(...)` and clears the flag. Result: worktree still on disk and re-pinned, but the `✕ removal failed — d to retry` marker is gone.

Same root cause as the routing comment above — only run this handler when a destroy was actually attempted.

internal/ui/app.go
```diff
@@ -2987,13 +3062,98 @@ func (h *Home) finalizeDelete(pd PendingDelete) tea.Cmd {
...
+}
+
+// destroyWorktree terminates the dev-stack processes still holding a worktree
+// (they detached from the session's tmux pane, so killing the pane didn't stop
+// them) and then removes the worktree. Editors, language servers, and shells
+// are spared via the proc denylist plus the configured editor. On failure it
+// re-scans and returns the remaining holder names so the caller can surface
+// what's still blocking removal. Runs on a background goroutine — must not
+// touch Home.
+func destroyWorktree(repoPath, workspaceName, editor string) (remaining []string, err error) {
```
hayke102: **Quit path bypasses this helper, so holders aren't killed on quit.**

`finalizeAllPendingDeletes` (~app.go:3179) still calls `workspace.ResolveProvider(...).Destroy(...)` directly instead of `destroyWorktree`. If the user deletes a worktree (dev daemons still pinning the dir) and quits within the 5s undo window before the cleanup goroutine runs, the quit path removes the worktree with no `proc.FindHolders`/`Kill` — so `git worktree remove` hits "Directory not empty", falls through to RemoveAll+prune (or fails), the daemons keep running orphaned after fleet exits, and no `failedWorktreeRemovals` flag is set (state is being torn down). The feature is inert on exactly the path it was meant to fix.

Route the quit path through `destroyWorktree` too (it's already goroutine-safe / doesn't touch `h`).

internal/ui/dialogs.go
```diff
@@ -421,6 +428,23 @@ func (d *ConfirmDialog) ShowDanger(title, subject string, details []string, onYe
 	d.subject = subject
 	d.details = details
 	d.onYes = onYes
+	d.scanLine = ""
```
hayke102, line 431: **`scanGen` is never reset here, so an in-flight scan can leak its kill-list onto a later dialog.**

`ShowDanger` (and `Show`) clear `scanLine` but leave `scanGen` at its previous value, and `nextHolderScanGen` only bumps for worktree dialogs. Sequence: `d` on a worktree → `StartScan(gen=N)` launches the ~0.5s lsof scan → user presses `n`, moves to a real-repo header, presses `d` → `ShowDanger("Remove repo from fleet?")` leaves `scanGen=N` and starts no scan. The gen=N result arrives; `SetScan` sees `d.visible && N == d.scanGen` and renders `Will terminate 3 process(es): node, air` onto the repo-forget dialog, which kills nothing — a misleading prompt that may scare the user into aborting.

Reset `scanGen` (or bump to an unused sentinel) here alongside `scanLine`.

internal/ui/app.go
```diff
@@ -2106,13 +2137,11 @@ func (h *Home) confirmDeleteHeader(item SidebarItem) tea.Cmd {
 	switch {
 	case isWorktree && count > 0:
 		title = "Remove Worktree?"
-		details = []string{
-			fmt.Sprintf("Deletes %d session(s) + the worktree directory", count),
-			"Press u to undo within 5s",
-		}
+		details = append([]string{fmt.Sprintf("Deletes %d session(s) + the worktree directory", count)}, h.worktreeDeleteWarnings(repoPath)...)
```
hayke102, line 2143: **The `d`-to-retry path can silently no-op if the cache stops classifying the path as a worktree.**

The retry affordance depends on `repoIsWorktree(repoPath)` staying true. But the fast-path above (app.go:2130) is `if count == 0 && !isWorktree { return h.unpinRepoHeader(repoPath) }`. If `destroyWorktree`'s RemoveAll removed the worktree's `.git` linkage but failed to fully rmdir (a holder recreated a file), the next git refresh flips `repoIsWorktree` to false. Pressing `d` on the flagged header then takes the instant-unpin branch — the leftover directory is never removed, the `failedWorktreeRemovals` entry is left dangling, and the retry becomes a no-op.

Consider keying the retry branch off `failedWorktreeRemovals[repoPath]` rather than the (mutable) worktree classification.

internal/workspace/provider.go
```diff
@@ -247,10 +254,31 @@ func (g *GitWorktreeProvider) Destroy(repoPath, name string) error {
...
+		// "Directory not empty" when something re-created files under the path.
+		// Fall back to force-removing the directory ourselves, then prune the
+		// now-missing worktree from the main repo's admin metadata (prune only
+		// deregisters worktrees whose dir is already gone — hence RemoveAll first).
+		debuglog.Logger.Warn("git worktree remove failed; falling back to RemoveAll+prune", "name", name, "path", wtPath, "err", rmErr)
+		if rmAllErr := os.RemoveAll(wtPath); rmAllErr != nil {
+			debuglog.Logger.Error("git worktree destroy: RemoveAll failed", "name", name, "path", wtPath, "err", rmAllErr)
+			return fmt.Errorf("git worktree remove: %s", rmErr)
+		}
+		if pruneOut, pruneErr := exec.CommandContext(ctx, "git", "-C", mainPath, "worktree", "prune").CombinedOutput(); pruneErr != nil {
```
hayke102: **`prune` reuses the shared 60s ctx, so a ctx-expiry-induced failure leaves dangling metadata while Destroy reports success.**

If the `git worktree remove` above failed because the shared 60s `ctx` expired (slow disk / large worktree), `os.RemoveAll` (no ctx) still wipes the directory, then this `prune` runs on the already-cancelled `ctx` and fails immediately — logged at warn and swallowed. `os.Stat` then finds the dir gone and returns `nil`, so the caller is told removal succeeded while the worktree's admin entry under `.git/worktrees` is left dangling.

Use a fresh context for the RemoveAll+prune fallback (the remove may well have failed *because* the original ctx is dead).

hayke102 (commented):
Code review of the worktree-holder-kill change — 6 inline findings, most-severe first. Mostly in the new delete/retry routing.

internal/ui/app.go
hayke102: Fixed in 11b2b84 — added an explicit `destroyAttempted` flag on `deleteCleanupDoneMsg`, set only where `destroyWorktree` actually runs, and the handler now gates on it instead of `repoPath != ""`. Normal session deletes no longer touch the worktree-failure handler.

internal/ui/app.go
hayke102: Fixed in 11b2b84 — same root cause. The N-1 non-destroy sibling messages now carry `destroyAttempted=false`, so they never enter `handleWorktreeDestroyResult` and can't clear a flag set by the last goroutine's failed destroy.

internal/ui/dialogs.go
hayke102: Fixed in 11b2b84 — `ShowDanger` and `Show` now reset `scanGen = 0` (an unused sentinel, since `nextHolderScanGen` yields ≥1) alongside `scanLine`, so a stale worktree-scan result fails the gen guard on a non-worktree dialog.

internal/workspace/provider.go
hayke102: Fixed in 11b2b84 — the RemoveAll+prune fallback now runs `prune` on a fresh 30s context instead of the (possibly already-expired) shared ctx, so a ctx-expiry-induced remove failure no longer leaves a dangling `.git/worktrees` entry while Destroy reports success.

internal/ui/app.go
hayke102: Fixed in 11b2b84 — `confirmDeleteHeader` now treats a path in `failedWorktreeRemovals` as a worktree (so the retry routes to destroy, not instant-unpin), and `destroyWorktree` falls back to a direct `os.RemoveAll` for an orphaned dir. That fallback is guarded on `git.IsWorktree(repoPath)==false`, so a healthy worktree (which keeps .git intact) can never be force-removed on a transient git error.

internal/ui/app.go
hayke102: Fixed in 11b2b84 — `finalizeAllPendingDeletes` now routes worktree cleanup through `destroyWorktree` (killing holders before removal) with grace=0, i.e. immediate SIGKILL, so the quit-within-undo-window path is covered without adding the 2s kill grace to app exit.

PR #122, opened 2026-06-07
fix(status): adopt rotated Claude session ids so finished sessions don't stick on "running"

hayke102:
## Problem

Two `/debug-status` snapshots (16:00 and 16:30, different repos) showed sessions stuck on **running** when the agent had clearly **finished**. Same root cause both times.

When Claude Code rotates its conversation `session_id` mid-life — compaction, `/clear`, or a continue/resume handoff — the new id (`1ba70e78`, `7fbd4947`) differs from the one fleet claimed at launch (`acff4e56`, `c12a0691`). The transcripts confirm a continuation: the new file's first entry lands **3–6 ms before** the old file's last entry.

`UpdateHookStatus`'s ownership guard treated the new id as a **nested/foreign Claude** and `return false`'d **every** hook from it — including the `Stop`/finished — **silently**. So:

1. The in-memory hook froze at the old session's last event (a stale `waiting`).
2. The pane fallback (`applyHookWaiting`) drove status from there. Its overridden fast-path only flipped to **running** on a spinner and did nothing on a finished pane, so once the agent finished it stayed pinned to the last `running` it saw.
3. Collateral: `ClaudeSessionID` stuck on the old id (restart/resume would resume the wrong conversation), and auto-naming froze.

The drop produced **no log line**, which is why this took a transcript dig to find.

## Fix

- **Adopt legitimate rotations.** When a non-owner `session_id` reports, distinguish a rotation from a concurrent nested child via **transcript continuity** (`isSessionRotation`): the rotated transcript's first entry lands within 10 s of the owner transcript's last entry. Rotations are adopted (ownership + resume id follow); concurrent children are still ignored. Transcript I/O runs on the rare non-owner path only, outside `s.mu`.
- **Symmetric fast-path** in `applyHookWaiting`: a stale overridden-waiting hook now settles to finished/idle when the pane shows the idle prompt, not just resumes on a spinner. Self-corrects the symptom even if a hook is dropped.
- **Visibility:** log ownership transfer (info) and foreign drops (debug).

## Tests

- `TestUpdateHookStatusAdoptsRotatedSession` / `TestUpdateHookStatusIgnoresNestedChild` — ownership handoff vs. nested-child rejection (hermetic, `HOME`-rooted transcript fixtures).
- `TestScenarioStaleOverriddenWaitingSettlesFinished` — the stuck-running transition.
- Verified each test **fails without its corresponding fix** and passes with it. Existing `TestScenarioNestedClaudeIgnored` / `TestScenarioOwnerEndStillReported` still pass. Full `go test -race ./...` green; `golangci-lint` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sessions no longer remain stuck in "running" after the assistant rotates its session ID mid-conversation (including during compaction, clearing, or continue), so session status updates now reflect actual conversation state.

* **Tests**
  * Added tests validating session rotation detection, transcript continuity handling, and correct ownership resolution between rotated and nested sessions.

### DIFF
--- a/internal/session/claude_name.go
+++ b/internal/session/claude_name.go
@@ -206,32 +206,48 @@ func CopyClaudeForkTranscript(parentSessionID, parentProjectPath, destProjectPat
 	return nil
 }
 
-// rotationHandoffWindow bounds how far apart a rotated session's first transcript
-// entry and the previous (owner) session's last entry may be while still counting
-// as a continuation. Claude Code's session-id rotations (compaction, /clear,
-// continue) hand off within milliseconds; the window absorbs clock/flush skew but
-// stays far tighter than the gap a concurrent nested child shows.
+// rotationHandoffWindow bounds how far apart a /clear-rotated session's first
+// transcript entry and the previous (owner) session's last entry may be while
+// still counting as a continuation. A /clear hands off within milliseconds; the
+// window absorbs clock/flush skew but stays far tighter than the gap a concurrent
+// nested child shows. (Linked rotations — continue/resume/fork — use the
+// deterministic uuid signal below instead, which has no time dependence.)
 const rotationHandoffWindow = 10 * time.Second
 
 // isSessionRotation reports whether newSessionID is ownerSessionID's conversation
 // continued under a fresh id (a Claude session-id rotation) rather than a separate
 // nested child claude that merely inherited the same FLEET_INSTANCE_ID.
 //
-// The signal is transcript continuity: a rotation's new transcript begins right
-// where the owner's ends, so the new file's first entry and the owner file's last
-// entry land within rotationHandoffWindow of each other. A concurrent child starts
-// mid-way through the owner's life and the owner keeps appending past it, so the
-// two never line up.
+// Claude rotates a session id three ways, and they leave different traces:
+//   - continue / resume / fork: the new transcript opens with a
+//     `compact_boundary`/`summary` entry whose `logicalParentUuid` points at the
+//     prior session's tail uuid. Preserved history means the new file's first
+//     timestamp is OLD (minutes-to-days before the handoff), so we match the
+//     uuid against the owner transcript instead — deterministic, time-independent,
+//     and a nested child can't match (its link, if any, points at its own parent).
+//   - /clear: a fresh conversation (first user entry `parentUuid:null`, no link)
+//     that hands off instantly, so the new file's first entry lands within
+//     rotationHandoffWindow of the owner's last entry.
+//   - in-place compaction keeps the same session id, so it never reaches here.
 //
-// When either transcript can't be read we return false — the safe default, since
-// wrongly adopting a real nested child would clobber the owner's status and resume
-// id.
+// When transcripts can't be read we return false — the safe default, since
+// wrongly adopting a real nested child would clobber the owner's status and
+// resume id.
 func isSessionRotation(projectPath, ownerSessionID, newSessionID string) bool {
 	if projectPath == "" || ownerSessionID == "" || newSessionID == "" {
 		return false
 	}
-	newFirst := firstTranscriptTimestamp(ClaudeTranscriptPath(newSessionID, projectPath))
-	ownerLast := lastTranscriptTimestamp(ClaudeTranscriptPath(ownerSessionID, projectPath))
+	newPath := ClaudeTranscriptPath(newSessionID, projectPath)
+	ownerPath := ClaudeTranscriptPath(ownerSessionID, projectPath)
+
+	// Primary: deterministic parent link (continue/resume/fork).
+	if link := transcriptParentLink(newPath); link != "" && transcriptContainsUUID(ownerPath, link) {
+		return true
+	}
+
+	// Fallback: instant timestamp handoff (/clear).
+	newFirst := firstTranscriptTimestamp(newPath)
+	ownerLast := lastTranscriptTimestamp(ownerPath)
 	if newFirst.IsZero() || ownerLast.IsZero() {
 		return false
 	}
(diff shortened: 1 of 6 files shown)


internal/session/ownership_test.go
```diff
@@ -0,0 +1,97 @@
...
+		t.Fatal(err)
+	}
+	dir := filepath.Join(home, ".claude", "projects", ClaudeProjectDirName(projectPath))
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	var b strings.Builder
+	for _, ts := range stamps {
+		b.WriteString(`{"type":"user","timestamp":"`)
+		b.WriteString(ts.UTC().Format(time.RFC3339Nano))
```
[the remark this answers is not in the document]
hayke102: Fixed in 2f9b843 — switched scanTranscriptTimestamp to parse `time.RFC3339Nano`. Note this wasn't actually a parse failure: Go's `time.Parse` accepts a fractional-seconds field even with the RFC3339 layout, and TestUpdateHookStatusAdoptsRotatedSession passes (it could only pass if the RFC3339Nano test timestamps parsed). Changed anyway since the Nano layout names the real transcript format (ms precision) and matches the test helper.

internal/session/session.go
```diff
@@ -237,20 +237,44 @@ func (s *Session) UpdateHookStatus(hs *HookStatus) bool {
...
+	// Claim ownership (first hook after launch), adopt a verified rotation, or
+	// no-op for the current owner.
 	if hs.SessionID != "" {
-		if s.ownerSessionID == "" {
-			s.ownerSessionID = hs.SessionID // claim ownership
-		} else if hs.SessionID != s.ownerSessionID {
-			return false // foreign (nested) Claude — ignore entirely
-		}
+		s.ownerSessionID = hs.SessionID
 	}
```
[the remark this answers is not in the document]
hayke102: Added a defensive re-check in 2f9b843: after taking the write lock, UpdateHookStatus re-reads ownerSessionID and drops the hook if it changed to a different non-empty id than the validated snapshot (the next worker cycle re-evaluates). Worth noting the described clobber is unreachable today — all UpdateHookStatus callers are serialized by h.workerMu, and the only writer outside it (clearHookState) only ever sets "", so owner can't become a different id mid-call. Kept the guard as defense-in-depth in case a future caller skips workerMu. No retry loop needed since dropping self-heals on the next cycle.

internal/session/session.go
```diff
@@ -237,20 +237,55 @@ func (s *Session) UpdateHookStatus(hs *HookStatus) bool {
...
+	//     resume id, so it must be ignored.
+	// isSessionRotation distinguishes them via transcript continuity.
+	var owner string
+	if hs.SessionID != "" {
+		s.mu.RLock()
+		owner = s.ownerSessionID
+		projectPath := s.ProjectPath
+		s.mu.RUnlock()
+		if owner != "" && hs.SessionID != owner {
+			if isSessionRotation(projectPath, owner, hs.SessionID) {
```
hayke102: **Blocking transcript I/O on the Bubble Tea `Update()` loop.** `isSessionRotation` does blocking file I/O (two `os.Open` + full-file JSONL scans), but `UpdateHookStatus` is reached from the UI goroutine, not only the worker: `hookChangedMsg`/`statusUpdateMsg` call `syncHookStatuses` inside `Update()` (`app.go:483`, `app.go:494`). So every foreign/rotated hook blocks the render loop on disk reads — the invariant CLAUDE.md calls out ("all blocking I/O … never in Bubble Tea `Update()`").

Also heads-up: the re-validation comment at L279-281 says callers are "serialized by workerMu" — they aren't. The worker calls `syncHookStatuses` **without** `workerMu` (`app.go:3452`; the lock is released after the snapshot at L3435), while the UI path holds it, so the worker and UI run `UpdateHookStatus` on the same session concurrently. The re-validation is therefore load-bearing, not just defensive — worth fixing the comment.

Suggest: skip rotation resolution on the UI path — return early for a foreign id and let the worker cycle adopt it (≤500ms later, off the loop).

internal/session/claude_name.go
```diff
@@ -204,3 +205,90 @@ func CopyClaudeForkTranscript(parentSessionID, parentProjectPath, destProjectPat
...
+// two never line up.
+//
+// When either transcript can't be read we return false — the safe default, since
+// wrongly adopting a real nested child would clobber the owner's status and resume
+// id.
+func isSessionRotation(projectPath, ownerSessionID, newSessionID string) bool {
+	if projectPath == "" || ownerSessionID == "" || newSessionID == "" {
+		return false
+	}
+	newFirst := firstTranscriptTimestamp(ClaudeTranscriptPath(newSessionID, projectPath))
```
hayke102: **Timestamp proximity misses compaction/continue rotations.** This works for `/clear` (fresh conversation → the new file's first entry ≈ now), but not the other two triggers named in the doc comment. A compaction/continue rotation's new transcript opens with *preserved/summarized history*, so its first timestamped entry carries an **old** timestamp.

I measured real linked child→parent transcripts in `~/.claude/projects` (child's first parent-link uuid resolving into the parent file): first-vs-parent-last gaps were ~30 min, ~27 min, ~63 min, ~4 hr, ~4 days — never < 10s. So `isSessionRotation` returns false and the rotation is dropped → the stuck-running bug persists for exactly those cases.

Claude records a direct link instead: the new transcript's `compact_boundary`/`summary` entry has `logicalParentUuid` pointing at the parent's tail uuid (verified in real data). Matching that against the owner transcript is deterministic, immune to handoff latency, and also eliminates the nested-child false-positive. Worth either switching to that signal, or scoping this fix to `/clear` and documenting the limitation. Which trigger actually caused the reported bug?

internal/session/claude_name.go
```diff
@@ -204,3 +205,90 @@ func CopyClaudeForkTranscript(parentSessionID, parentProjectPath, destProjectPat
...
+
+// firstTranscriptTimestamp returns the timestamp of the earliest timestamped JSONL
+// entry in the transcript at path, or the zero time if it's missing/unreadable or
+// has no timestamped entries.
+func firstTranscriptTimestamp(path string) time.Time { return scanTranscriptTimestamp(path, false) }
+
+// lastTranscriptTimestamp returns the timestamp of the latest timestamped JSONL
+// entry in the transcript at path, or the zero time. It scans the whole file;
+// transcripts are small and this runs only on the rare session-id-change path.
+func lastTranscriptTimestamp(path string) time.Time { return scanTranscriptTimestamp(path, true) }
```
hayke102, line 337: **Uncached whole-file rescans on the fast path.** `lastTranscriptTimestamp` rescans the entire (growing) owner transcript, and the rotation check runs on *every* `UpdateHookStatus` call — before the `changed` test — so a persistent foreign id (a nested-child eval harness, the exact case this guards against) drives two full-file scans every ~500ms fast cycle indefinitely, on the path the split-cadence worker deliberately keeps to in-memory lookups. Consider negative-caching the `(owner, foreign)` verdict, or using `os.Stat` ModTime as the owner's last-write proxy instead of a full scan.

internal/session/session.go
```diff
@@ -621,13 +656,24 @@ func (s *Session) applyHookRunning(oldStatus Status, paneContent string, paneSta
...
+	// stale overridden-waiting hook from staying pinned to the last "running" it
+	// saw — e.g. when the real Stop hook was dropped after a session-id rotation.
 	if !s.hookOverriddenAt.IsZero() && s.hookOverriddenAt.Equal(s.hookUpdatedAt) {
-		if paneStatus == StatusRunning {
+		switch paneStatus {
+		case StatusRunning:
 			s.Status = StatusRunning
 			s.Acknowledged = false
 			log.Info("overridden waiting hook but pane shows running, resuming")
+		case StatusFinished:
```
hayke102, line 699: **Overridden-finished branch: missing reset + unhandled pane states.** Two gaps vs the non-overridden finished path below: (1) this `case StatusFinished` doesn't clear `lastContentHash`/`lastContentChangeAt` the way the normal settle does; (2) the `switch` only handles `Running`/`Finished` — if the pane settles to anything else (a non-prompt screen, an empty capture) it matches no case and returns, leaving the session pinned at its last status (e.g. Running) with no stability fallback. A `default` that falls through to the content-stability heuristic would close that.

internal/session/session.go
hayke102: Fixed in 152bf25. You're right on both counts — verified the worker runs syncHookStatuses *without* workerMu (it's released after the session-slice snapshot), so worker + UI race and the re-check is load-bearing, not defensive. Threaded a `resolveRotation` flag through syncHookStatuses → UpdateHookStatus: the worker passes true (transcript I/O runs off the Update() loop), and the two UI paths (hookChangedMsg/statusUpdateMsg) pass false — a foreign id is deferred to the next worker cycle (~500ms), which adopts it. Rewrote the misleading workerMu comment.

internal/session/claude_name.go
hayke102: Fixed in 152bf25 — switched to the logicalParentUuid signal you suggested, with proximity kept only as the /clear fallback. To answer your question: the reported bug was a /clear-type rotation — I checked the two stuck transcripts (1ba70e78, 7fbd4947) and both open with a fresh user entry (parentUuid:null) at the handoff instant (3–6ms), no compact_boundary, so proximity is the right signal there. Confirmed your finding for the other triggers: in my ~/.claude/projects, continue/resume/fork rotations open with {type:system, subtype:compact_boundary, logicalParentUuid:<parent tail>} and old first timestamps. isSessionRotation now matches that logicalParentUuid against the owner transcript first (deterministic, nested-child-proof), then falls back to proximity for /clear (which has no link). One note from the data: in-place compaction keeps the same session_id (mid-file compact_boundary), so it never reaches this path — only cross-file continue/resume/fork do. Added a linked-rotation adopt test + a nested-child-with-foreign-compaction reject test.

internal/session/claude_name.go
hayke102: Fixed in 152bf25. Added a per-session negative cache of the (owner, foreign) verdict: once isSessionRotation rejects a foreign id it's recorded, and subsequent hooks from that same id short-circuit before any transcript I/O — so a persistent nested-child harness no longer drives full-file scans every fast cycle. The cache self-invalidates when the owner or foreign id changes and is cleared in clearHookState. Combined with the resolveRotation gate from the other thread, the rotation I/O now only happens on the worker path and at most once per distinct foreign id.

internal/session/session.go
hayke102: Fixed 4a in 152bf25: the overridden→finished settle now clears lastContentHash/lastContentChangeAt, matching the non-overridden path below. For 4b I kept the early return rather than falling through: this branch's whole purpose is to short-circuit an already-resolved stale hook, and falling through re-enters the unconditional `Status = StatusWaiting` + hash logic, which risks re-introducing the oscillation the override prevents. The unhandled-pane case (empty/non-prompt capture under a stale overridden hook) is rare and self-heals on the next real hook, so I'd rather not widen the blast radius here. Happy to revisit with a dedicated default→stability path if you've seen it pin in practice.

PR #126, opened 2026-06-07
feat(ui): contextual tips in bottom-right corner

hayke102:
## What

Adds a small, **extensible tips engine** that shows an actionable hint box in the **bottom-right** when a condition makes it relevant. Dismiss any tip with **Shift+X**.

Motivation: after a reboot every session is dead and the sidebar fills with `✕ error` rows — the fix (Ctrl+K → "Reload All Sessions") already exists but is undiscoverable. This teaches it, and gives us a home for future hints (slot hotkeys, approve-waiting, eventually announcement-style tips like Linear/Jira).

## Tips shipped

1. **Failed → Reload All** (`tipRecurring`, priority 100) — fires at **≥4** `StatusError` sessions → *"N sessions stopped. Press Ctrl+K → \"Reload All Sessions\" to restart them."*
2. **Command-palette discovery** (`tipOnce`, priority 10) — one-time hint once you have ≥3 sessions.

## Design

- **`internal/ui/tips.go`** — `tipRegistry` catalog + engine. Two dismissal policies:
  - `tipRecurring`: condition-driven; Shift+X hides the current occurrence, returns when the condition recurs (and on a fresh launch). **Not persisted.**
  - `tipOnce`: no recurrence; shows while relevant, auto-expires after `tipOnceBudget` (45s), and once dismissed/expired is recorded in config `seen_tips` so it never returns.
- `refreshTips` runs on the existing ~2s tick (render path stays side-effect free); `dismissActiveTip` handles Shift+X.
- `renderTip` mirrors `renderToast`; the box stacks above any toast and reuses the existing `overlay.Composite` bottom-right anchor. Suppressed while a modal owns the screen (new `modalOpen` helper).
- `internal/config/config.go` — `SeenTips []string` (`seen_tips`) + `IsTipSeen`, mirroring the `DisplayOnboardingSeen`/`Save` pattern.
- `internal/ui/keybindings.go` — `X → Dismiss on-screen tip`.

## Adding tips later

Append a `Tip{}` to `tipRegistry` with `ID`, `Policy`, `Priority`, `active(h)`, and `text(h)`. That's the whole surface.

## Notes

- No global on/off setting in v1 — per-tip dismissal covers it; easy to add a Behavior toggle later.

## Testing

- `internal/ui/tips_test.go` covers threshold, episode reset/recurrence, tipOnce seen-respect + persisted dismissal, priority ordering, and `IsTipSeen`.
- `make build`, `make fmt`, `make lint` (0 issues), `make test` (race) all green.
- Manual: start fleet with ≥4 sessions, `tmux kill-server` → tip appears within ~2s; Ctrl+K → Reload All clears it; Shift+X dismisses (recurs on next failure). Command-palette tip persists to `seen_tips` once dismissed.


internal/ui/tips.go
```diff
@@ -0,0 +1,224 @@
...
+				continue
+			}
+		case tipOnce:
+			if !act || h.cfg.IsTipSeen(t.ID) {
+				continue
+			}
+			// Start the visibility timer on first display, then retire the tip
+			// once it has had its time on screen.
+			if _, ok := h.tipShownSince[t.ID]; !ok {
+				h.tipShownSince[t.ID] = now
```
hayke102: **The `tipOnce` budget runs even when the tip is never shown.** `tipShownSince` is started and the budget is checked *before* the priority pick below decides which tip actually renders — so a lower-priority `tipOnce` that loses (or is hidden behind a modal, since `View` suppresses it while the budget keeps ticking) is still retired by `markTipSeen` after 45s.

Concretely: with ≥4 error sessions the reload tip (Priority 100) renders, but ≥4 also satisfies `cmdPaletteMinSessions=3`, so `command_palette` (Priority 10) is `active`. Its timer starts and counts down unseen; after 45s of the error condition it's written to `seen_tips` — the feature's only discovery tip is consumed without ever appearing on screen.

Suggest advancing/starting the budget only for the tip `refreshTips` actually **chose**, not for every active `tipOnce`.

internal/ui/tips.go
```diff
@@ -0,0 +1,224 @@
...
+		case tipRecurring:
+			if !act {
+				// Episode ended — clear dismissal so the tip can recur.
+				delete(h.tipEpisodeDismissed, t.ID)
+				continue
+			}
+			if h.tipEpisodeDismissed[t.ID] {
+				continue
+			}
+		case tipOnce:
```
hayke102, line 128: **Stale `tipShownSince` retires a `tipOnce` early on a flapping condition.** When the condition goes inactive before the budget elapses this branch just `continue`s and never clears `tipShownSince[t.ID]` (unlike `tipRecurring`, which `delete()`s its flag on the inactive path).

So: user briefly hits 3 sessions (timer set), drops to 2 for >45s, returns to 3 — now `now.Sub(staleStart)` already exceeds `tipOnceBudget`, so the tip is `markTipSeen`'d immediately without a meaningful appearance. A `delete(h.tipShownSince, t.ID)` on the inactive path fixes this and the issue above shares the same root: the timer is wall-clock from first eligibility, decoupled from actual display.

internal/ui/app.go
```diff
@@ -1192,13 +1199,47 @@ func (h *Home) View() string {
...
+			stack = tip
+		}
+	}
+	return overlay.Composite(stack, base, overlay.Right, overlay.Bottom, -1, -1)
+}
+
+// modalOpen reports whether any full-screen view currently owns the screen.
+// Mirrors the early-returns in renderBody (plus the command-palette overlay) so
+// a sticky bottom-right tip is never composited on top of a dialog.
+func (h *Home) modalOpen() bool {
```
hayke102, line 1229: **Altitude:** `modalOpen()` hand-mirrors `renderBody`'s 12 `IsVisible()` early-returns (plus command palette + launchpad) with no shared source of truth. The list is correct today, but a future dialog wired into `renderBody`'s early-return and forgotten here will let the sticky tip composite on top of it — exactly the bug this function exists to prevent. Consider a single ordered predicate consumed by both `renderBody` and `modalOpen` so the two can't drift.

internal/ui/tips.go
```diff
@@ -0,0 +1,224 @@
...
+	t := findTip(h.activeTipID)
+	if t == nil {
+		return ""
+	}
+	return renderTip(t.text(h), h.width)
+}
+
+// tipMaxWidth bounds the tip column; matches the toast column for visual
+// consistency when both stack in the bottom-right.
+const tipMaxWidth = 44
```
hayke102: **Reuse:** `tipMaxWidth = 44` duplicates `toastMaxWidth = 44` (toast.go:22) even though the comment promises the tip “matches the toast column,” and `renderTip` below is a near-verbatim copy of `renderToast` (innerWidth/bodyWidth math, wrap, rounded-border/`Width(width-toastBorderCols)` close). If someone widens or fixes the box math in `renderToast`, the stacked tip silently won't follow and the two boxes misalign. Reuse `toastMaxWidth` and extract the shared box body (e.g. `renderHintBox(lines, footer, accent)`).

hayke102 (commented):
Review of the contextual-tips feature — 2 correctness issues (shared root cause) plus an altitude and a reuse note.

internal/ui/tips.go
hayke102: Fixed — refreshTips now picks the tip first, then ages the display budget only for the chosen + visible tipOnce (accumulated tipVisibleFor gated on visibility). command_palette no longer counts down while the reload tip renders or while a modal covers it.

internal/ui/tips.go
hayke102: Fixed — the inactive (and non-chosen) path now deletes the tip's accumulated time, and the budget is on-screen time rather than wall-clock-from-first-eligibility, so a flapping condition starts fresh instead of retiring the tip immediately. Same restructure as the sibling thread.

internal/ui/app.go
hayke102: Addressed via a reflection-based regression test (internal/ui/modal_test.go) rather than a shared predicate, to avoid restructuring the hot renderBody path. It enumerates every Home dialog field exposing IsVisible()+visible, flips each visible, and asserts modalOpen() covers it — so a future dialog auto-joins the check and fails CI if modalOpen forgets it. modalOpen itself is unchanged.

internal/ui/tips.go
hayke102: Fixed — dropped tipMaxWidth (now uses toastMaxWidth) and extracted a shared renderHintBox(sigil, accent, body, footer, width) in toast.go. renderToast calls it with no footer (output is byte-identical) and renderTip with the ✦ sigil + dismiss footer, so the stacked boxes can't drift.

PR #127, opened 2026-06-08
fix(status): time out git/gh calls so a hang can't freeze every session's status

hayke102:
## Problem

A session showed `running` for ~12 minutes when it was genuinely `waiting`. Tracing it revealed this is **not a status-detection bug** — the detection was correct, it just wasn't being called.

The decisive evidence: three unrelated sessions all flushed their pending status within **11ms** of each other, with hook ages of **1m45s, 17m23s, and 12m16s**. A single burst draining a multi-minute backlog across sessions = the background status worker goroutine **stalled, then resumed**. The UI stayed responsive the whole time (the `D`-key snapshot worked).

## Root cause

`UpdateStatus()` runs only on the `statusWorker` goroutine. Its heavy cycle calls `refreshAllGitAndPR(repos, 4, 0, nil)` with `deadline=0`, which runs `dispatchAndWait()` **synchronously**, ending in `wg.Wait()`. The per-repo goroutines run `git`/`gh` execs that had **no timeouts** — so one hung `gh` call (network stall, blocked auth) meant `wg.Done()` never fired, `wg.Wait()` blocked forever, and the entire worker (every session's status) froze until the OS finally tore the connection down (~18 min).

## Fix

- **git/gh timeouts (the actual fix):** wrap all git execs at 8s (`git.go` `gitOutput`/`gitRun`, `remote.go`, `session.go` `GetRepoRoot`) and gh execs at 15s (`pr.go` `ghTimeout`). A hung call now errors in bounded time, so the cycle always completes. `tmux` calls already had timeouts; git/gh were the gap.
- **Worker heartbeat in snapshots:** `lastWorkerCycleNano`/`workerCycleStartNano` atomics stamped in `statusWorkerCycle`. The `D`-key snapshot now emits a `worker` block in `snapshot.json` (`last_cycle_ago`, `stalled`) + a `goroutines.txt` dump — so a worker stall is a one-glance diagnosis next time.
- **Dev-only watchdog:** `workerWatchdog` launches **only when `version=="dev"`** (i.e. `make run`; released builds get a real semver and never run it). If a cycle doesn't complete for 30s it auto-writes a `<ts>_worker-stall/` snapshot dir with a goroutine dump — no need to be quick on the `D` key.

## Testing

- `make build` clean, `go vet` clean
- `go test -race ./internal/session/...` (status detection suite) + `github`/`ui` tests pass

## Not changed (flagged)

The worker still does its git/PR fan-out with `deadline=0` (synchronous `wg.Wait()`). The per-call timeouts now bound it; a whole-fan-out deadline would be belt-and-suspenders but wasn't needed to fix the stall.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session statuses no longer freeze when `git` or `gh` hang; subprocess calls now time out (8s for git, 15s for gh) to keep the background worker responsive.
  * PR info refresh now preserves cached PR data on transient GitHub errors instead of clearing it.

* **New Features**
  * Status snapshots now include worker liveness and produce goroutine dump files for improved diagnostics; a dev-only watchdog captures stall information.

* **Chores**
  * Updated changelog for unreleased changes.

### DIFF
--- a/changelog/unreleased/status-worker-timeouts.md
+++ b/changelog/unreleased/status-worker-timeouts.md
@@ -1,0 +1,4 @@
+---
+type: fixed
+---
+Session statuses no longer freeze across the whole app when a `git` or `gh` call hangs — those calls now time out (8s/15s) instead of wedging the background status worker indefinitely.
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -1,6 +1,7 @@
 package git
 
 import (
+	"context"
 	"fmt"
 	"os/exec"
 	"path/filepath"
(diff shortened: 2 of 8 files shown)


hayke102:
Superseded by #128, which bundled this exact status-worker fix (its squash commit 07ed71c carries the `fix(status): time out git/gh calls...` commit verbatim) and then refined it — master now has all of this work with a better watchdog threshold (90s vs 30s) and first-cycle wedge detection. Closing as already-shipped; nothing here is missing from master.

PR #128, opened 2026-06-08
fix(tmux): copy selections to system clipboard via copy-command pbcopy

hayke102:
## Problem
Selecting text in an attached session didn't reach the system clipboard. The selection went only to tmux's paste buffer, or rode OSC 52 — which **iTerm2 blocks by default** and **Apple Terminal doesn't support** — so a drag/double-click/keyboard copy silently did nothing.

## Fix
Set **`copy-command pbcopy`** (server option, **only when unset** so a user's own `copy-command` is respected). tmux's default copy bindings run `copy-pipe-and-cancel` *with no argument*, which pipes to `copy-command`, so this routes **every** copy gesture — drag, double-click, triple-click, keyboard — to `pbcopy`. `pbcopy` is a local pipe, so it works on every terminal regardless of OSC 52.

Applied from `Start` (server guaranteed up) and the startup bootstrap (for users with existing sessions). `copy-command` is a server option, so it touches the shared tmux server — but it's one benign option, set once, only when unset.

## Warp
Warp **can't drag-copy at all** — it doesn't deliver mouse drag-selection to terminal apps, so tmux never enters copy-mode and no tmux-side change can help. The status bar now shows an amber `⚠ Warp doesn't support drag-to-copy (Warp bug)` when `TERM_PROGRAM=WarpTerminal`.

## Verified
- ✅ iTerm2: drag-select → clipboard (was the original failure — OSC 52 blocked)
- ✅ `pbcopy` pipe confirmed reaching the macOS clipboard from the tmux server
- `make build` + `go vet` pass

## Summary by CodeRabbit

* **New Features**
  * Improved copy-to-clipboard functionality for iTerm2 and Apple Terminal
  * Added Warp terminal detection with helpful notification about drag-copy limitations

* **Bug Fixes**
  * Fixed session status freezing when git or GitHub operations stall by enforcing command execution timeouts

* **Documentation**
  * Added guides covering terminal behavior, clipboard configuration, and status timeout handling


internal/github/pr.go
```diff
@@ -84,7 +94,9 @@ func GetPRForBranch(repoPath, branch string, ignorePatterns []string) (*PR, erro
 
 	debuglog.Logger.Debug("PR fetch: start", "path", repoPath, "branch", branch)
 
-	cmd := exec.Command("gh", "pr", "view", branch,
+	ctx, cancel := context.WithTimeout(context.Background(), ghTimeout)
+	defer cancel()
+	cmd := exec.CommandContext(ctx, "gh", "pr", "view", branch,
```
hayke102, line 99: **A `gh` timeout now silently clears the PR badge.** When this `gh pr view` hits the 15s deadline, `cmd.Output()` errors with an empty stderr, so `classifyGHError("")` returns nil and the function falls through to `return nil, nil` at line 116 — indistinguishable from "no PR exists for this branch".

`RefreshPRInfo` (`internal/git/repo_info.go`) treats `(nil, nil)` as a successful round-trip and sets `info.PR = nil`, so any 15s+ network blip during the 60s refresh flickers a real PR's badge off until the next cycle. The rate-limit path already preserves the cached PR; the timeout path should too — e.g. detect `ctx.Err()`/`errors.Is(err, context.DeadlineExceeded)` and return that error so `RefreshPRInfo` leaves `info.PR` untouched.

internal/tmux/tmux.go
```diff
@@ -56,6 +58,37 @@ func IsTmuxAvailable() error {
...
+func EnsureCopyCommand() {
+	if copyCommandDone.Load() {
+		return
+	}
+	out, err := exec.Command("tmux", "show-options", "-sv", "copy-command").Output()
+	if err != nil {
+		return // No server yet (or tmux error) — retry on a later call.
+	}
+	if strings.TrimSpace(string(out)) == "" {
+		_ = exec.Command("tmux", "set-option", "-s", "copy-command", "pbcopy").Run()
```
hayke102, line 88: **This writes a *server-global* tmux option with no opt-out, and the "only when unset" guard can't tell intent from default.** `copy-command` empty is also the state of a user who deliberately relies on OSC 52 (e.g. copying from a remote tmux to their local clipboard over SSH). Setting it server-wide forces *all* their tmux sessions — including non-fleet ones on the shared server — through the local `pbcopy` pipe, silently breaking that workflow, with no config knob to disable it.

Worth at least a config escape hatch (a `copy_command` setting, or skip when an env like `FLEET_NO_COPY_COMMAND` is set). The CLAUDE.md note acknowledges the server scope is intentional, so this is a flag for discussion rather than a hard bug.

internal/tmux/tmux.go
```diff
@@ -56,6 +58,37 @@ func IsTmuxAvailable() error {
...
+	}
+	copyCommandDone.Store(true) // Either we set it, or the user has their own.
```
hayke102: **`copyCommandDone` latches `true` even when the `set-option` failed.** The error from line 87's `set-option` is discarded (`_ =`), then this line marks the feature resolved unconditionally. If that `set-option` fails transiently (server momentarily busy / signalled) after `show-options` succeeded, copy-to-clipboard is silently disabled for the rest of the run and never retried.

Only latch when the set actually succeeded (or the value was already non-empty):
```go
if strings.TrimSpace(string(out)) != "" {
    copyCommandDone.Store(true) // user has their own
    return
}
if exec.Command("tmux", "set-option", "-s", "copy-command", "pbcopy").Run() == nil {
    copyCommandDone.Store(true)
}
```

internal/git/git.go
```diff
@@ -11,11 +12,31 @@ import (
 	"github.com/brizzai/fleet/internal/debuglog"
 )
 
+// gitTimeout bounds every git subprocess. git is normally local and fast, but a
+// hung call (a held .git/index.lock, a dead NFS mount, a wedged filesystem)
+// must not freeze the caller. The status worker waits synchronously on these
+// during its per-cycle git+PR fan-out, so a single indefinite git call would
+// stall every session's status update.
+const gitTimeout = 8 * time.Second
```
hayke102, line 20: **The timeout campaign missed a git call on the startup path.** `internal/analytics/analytics.go:161` `readGitIdentity` still runs two *unbounded* `git config --global` calls, and it's invoked synchronously via `DiscoverIdentity()` in `cmd/fleet/main.go:142` **before** `p.Run()`. A hung `git config` (dead/NFS `$HOME`, held lock on the global config, wedged credential helper) freezes fleet at startup before the TUI renders — exactly the freeze class this PR set out to bound, just on the launch path instead of the worker. Worth wrapping those two calls in the same `gitTimeout`/`CommandContext` pattern.

internal/git/git.go
```diff
@@ -163,14 +181,12 @@ func GetDefaultBranch(repoPath string) string {
 
 // IsWorktree returns true if the given path is a git worktree (not the main repo).
 func IsWorktree(repoPath string) bool {
-	gitDir := exec.Command("git", "-C", repoPath, "rev-parse", "--git-dir")
-	gitDirOut, err := gitDir.Output()
+	gitDirOut, err := gitOutput("-C", repoPath, "rev-parse", "--git-dir")
```
hayke102, line 184: **Timeout collapses into "not a worktree," which changes a destructive action.** On an 8s deadline this returns `false` — identical to a genuine non-worktree result. Per CLAUDE.md, the delete-scope decision for an empty/uncached header falls back to `git.IsWorktree`. If that git call hangs (held `.git/index.lock`, slow NFS) and trips the deadline, the header is treated as a real repo, so `d` runs the **"forget repo"** path instead of **`git worktree remove`** — a different destructive action than the user would have gotten before the timeout was added. Low likelihood, but it's on a destructive path, so a timeout probably shouldn't be silently squashed to `false` here.

internal/ui/app.go
```diff
@@ -3794,6 +3821,68 @@ drainPriority:
 	h.refreshTmuxStatusBars(sessions)
 }
 
+// workerStallThreshold is how long the status worker may go without completing a
+// cycle before the dev-only watchdog treats it as wedged and dumps goroutine
+// stacks. Set well above the worst-case bounded cycle (git/gh now time out at
+// 8s/15s, fanned out 4-wide), so it fires only when a call ignores its deadline
+// or a real deadlock occurs — not on a merely slow cycle.
+const workerStallThreshold = 30 * time.Second
```
hayke102: **30s is below the real worst-case healthy cycle, so the watchdog/`stalled` flag can false-fire.** A single repo's git refresh is serial (`GetBranchName` 8s + `HasUncommittedChanges` 8s + `IsWorktree` = two 8s calls + `GetOriginKey` 8s ≈ 40s), and `GetPRForBranch` chains **two independent 15s `gh` calls** (`gh pr view` + the `gh api graphql` in `getUnresolvedThreadCount` ≈ 30s). When gh/git are slow-but-bounded, a legitimate cycle exceeds 30s; the dev watchdog then dumps goroutines and the D-key snapshot reports the worker "wedged" though it's healthy.

Two follow-ups: (1) raise this above the true fanned-out worst case (or give `GetPRForBranch` a single shared budget), and (2) the `status-worker-timeouts.md` changelog and the comment here say "15s" — `GetPRForBranch`'s real ceiling is ~30s because of the two chained calls.

internal/tmux/tmux.go
```diff
@@ -263,6 +308,15 @@ func (s *Session) ApplyStatusBar(o StatusBarOpts) {
 	}
 }
 
+// isWarpTerminal reports whether fleet is running inside Warp. Warp has a
+// long-standing bug where it doesn't hand mouse drag-selection to terminal
+// apps, so drag-to-copy can't work in an attached session there and no
+// tmux-side fix exists (see the warning surfaced in the status bar). Detected
+// from the fleet process's own env — the outer terminal it was launched from.
+func isWarpTerminal() bool {
+	return os.Getenv("TERM_PROGRAM") == "WarpTerminal"
```
hayke102, line 333: **Warp is detected from fleet's *launcher* env, but the constraint depends on the terminal the session is *attached* from.** These can differ: launch fleet from iTerm2 and attach to a session inside Warp → drag-copy silently fails with **no** warning (the bar was painted Warp-free); launch from Warp and attach in iTerm2 → the persistent "Warp doesn't support drag-to-copy" banner is branded onto a session where drag-copy actually works. The bar is also written once and never repainted on env change, so the flag is stale. Hard to fully solve from tmux, but worth a comment acknowledging it only reflects the launch terminal — or detecting at attach time if feasible.

internal/git/git.go
```diff
@@ -128,7 +146,9 @@ func GetUserEmail(repoPath string) string {
 // CheckoutBranch checks out the given branch in the repo.
 // For remote-only branches, git auto-creates a local tracking branch.
 func CheckoutBranch(repoPath, branch string) error {
-	cmd := exec.Command("git", "-C", repoPath, "checkout", branch)
+	ctx, cancel := context.WithTimeout(context.Background(), gitTimeout)
```
hayke102, line 149: **Nit (consistency): the timeout is centralized into `gitOutput`/`gitRun`, but several call sites bypass them.** `CheckoutBranch` here, `session.GetRepoRoot` (which even re-hardcodes its own `8*time.Second` rather than `gitTimeout`), and all of `internal/github/pr.go` inline their own `context.WithTimeout`/`CommandContext`. The 8s/15s deadlines are now duplicated literals that can drift if the policy changes, and you can't tell which calls are bounded by grepping for the helper. Consider a `gitCombined(args...)` helper for the `CombinedOutput` case, and a `ghOutput`/`ghRun` pair mirroring `gitOutput`/`gitRun` for pr.go.

internal/github/pr.go
hayke102: Fixed — `GetPRForBranch` now returns the `DeadlineExceeded` error on timeout, and `RefreshPRInfo` preserves the cached PR on a timeout (mirroring the rate-limit path) instead of falling through to `(nil, nil)` and clearing the badge.

internal/tmux/tmux.go
hayke102: Fixed — added a `FLEET_NO_COPY_COMMAND` opt-out (truthy) to skip setting `copy-command` entirely, for users who deliberately rely on OSC 52 (e.g. remote tmux → local clipboard over SSH). It still only sets `copy-command` when unset otherwise.

internal/tmux/tmux.go
hayke102: Fixed — removed the `copyCommandDone` latch entirely. `EnsureCopyCommand` now re-checks the live server on every call, so a failed `set-option` is retried (and a fresh server after a restart still gets `pbcopy`). This also resolves the cross-server-restart caching concern.

internal/git/git.go
hayke102: Fixed — wrapped both `git config --global` calls in `readGitIdentity` in a 5s `context` timeout, so a hung `git config` on the startup path can't freeze fleet before the TUI renders.

internal/git/git.go
hayke102: Acknowledged — leaving as-is. Likelihood is very low (a local `git rev-parse --git-dir` hanging 8s), and the invasive `(bool, error)` signature change isn't worth it here: the mis-routed "forget repo" path only unpins + drops fleet's session records, leaving the worktree folder untouched and fully recoverable. Noted for a future pass if it ever shows up in practice.

internal/ui/app.go
hayke102: Fixed — raised `workerStallThreshold` to 90s (above the true fanned-out worst case) and corrected the comment to note `GetPRForBranch`'s ~30s ceiling from its two chained gh calls (`gh pr view` + the `gh api graphql` thread-count query).

internal/tmux/tmux.go
hayke102: Fixed — expanded the `isWarpTerminal` comment: it reflects fleet's launch terminal, which IS the attach terminal in the normal PTY-attach flow (fleet attaches inside its own process). The one case it can't see is a manual `tmux attach` from a different terminal, which is out of fleet's control; attach-time detection from the status bar isn't feasible.

internal/git/git.go
hayke102: Fair — leaving as-is for now. pr.go already centralizes on its own `ghTimeout` const, and `CheckoutBranch` needs `CombinedOutput` (not `gitOutput`'s `Output`), so the only real drift is `GetRepoRoot`'s hardcoded 8s. Noting it for a future cleanup pass rather than expanding this PR's scope.

changelog/unreleased/status-worker-timeouts.md
```diff
@@ -0,0 +1,4 @@
+---
+type: fixed
+---
+Session statuses no longer freeze across the whole app when a `git` or `gh` call hangs — those calls now time out (8s/15s) instead of wedging the background status worker indefinitely.
```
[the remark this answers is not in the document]
hayke102: Fixed — rephrased the fragment to be user-facing: "Session statuses no longer freeze across the app when a background status check gets stuck." (dropped the git/gh and 8s/15s internals).

internal/tmux/tmux.go
```diff
@@ -4,9 +4,11 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
+	"os"
 	"os/exec"
 	"strings"
 	"sync"
+	"sync/atomic"
```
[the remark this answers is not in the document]
hayke102: Fixed — removed the process-global `copyCommandDone` cache. `EnsureCopyCommand` now re-checks the live tmux server on each call (cheap `show-options`), so a server created fresh after a restart still gets `pbcopy`.

internal/ui/app.go
```diff
@@ -3794,6 +3821,68 @@ drainPriority:
...
+		if stalledFor < workerStallThreshold {
+			continue
+		}
+		// One dump per stall episode: lastWorkerCycleNano is frozen while wedged,
+		// so dump once and wait for it to advance (worker recovered) before arming
+		// again.
+		if last == dumpedForCycle {
+			continue
+		}
+		dumpedForCycle = last
```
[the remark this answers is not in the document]
hayke102: Fixed — the watchdog now falls back to `workerCycleStartNano` when no cycle has completed yet (`last == 0`), so a wedge on the very first cycle is caught instead of being skipped.

PR #133, opened 2026-06-16
fix(status): false-finish on stale waiting hook; fork-of-fork re-forking ancestor

hayke102:
Two independent session-tracking fixes, both rooted in how fleet reconciles Claude hooks against the conversation transcript.

## 1. Session shows "finished" while the agent is actively running

`fix(status): keep session running when transcript advances past a stale waiting hook` (57437f0)

A granted permission prompt fires no resume hook, so the hook lingers at `waiting`. Modern Claude keeps the `❯` input box on screen while working, so a between-bursts frame (activity line momentarily gone) reads as a finished idle prompt — and `applyHookWaiting` overrode `waiting → finished` while the agent was actively running. Verified from a snapshot: at both false-finished moments the transcript's last lead entry was ~0.5s/3s old and ~90s past the hook.

**Fix:** when the hook is `waiting` and the pane reads finished, consult the transcript (`conversationActivePastHook`) — if the last lead (non-sidechain) entry advanced >2s past the hook and is <120s old, the agent resumed, so stay running. The read happens **before `s.mu`** (disk I/O) and is gated to that one frame. The running-hook path is intentionally left to its existing 10s-stable + tmux-activity guards.

## 2. Forking a forked session re-forked its ancestor

`fix(fork): forking a forked session re-forked its ancestor, not itself` (c7c1742)

Fork A → B, work on B, fork B → C was a fork of **A**, not B.

`claude --resume <parent> --fork-session` reports the **parent's** id at `SessionStart` and only switches to the fork's own id on the first prompt. `isSessionRotation` can't adopt that new id for a fork: fork transcripts carry no `logicalParentUuid`, and they copy the parent's history verbatim **with the parent's original timestamps** — so the 10s timestamp-handoff check measures the parent's whole conversation length, not a handoff gap, and misses for any parent older than ~10s. The miss is then permanently negative-cached, pinning the fork to the parent id; `forkSelected` reads `ClaudeSessionID` → re-forks the parent. (A trivial <10s parent squeaks under the window — why it looked intermittent.)

**Fix:** remember the fork parent id (`Session.forkParentID`, captured in `Start`). When a hook reports an id different from the parent while the session still owns the parent's id, adopt it **deterministically** — it can only be the fork diverging. No transcript heuristic, no 10s window, no negative cache; cleared on divergence; self-heals across restart.

## Tests
- `TestScenarioStaleWaitingConversationActiveStaysRunning` + `…IdleSettlesFinished`, `TestLastLeadTranscriptTimestamp` (status fix)
- `TestUpdateHookStatusAdoptsForkDivergence` (fork fix) — reproduces the exact failure (>10s parent, copied timestamps, no parent link)
- Each new test verified to fail with its fix disabled and pass with it; full suite green under `-race`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Forked-from-fork sessions now correctly switch to the fork’s own conversation ID once the fork diverges.
  * Sessions no longer show stale “waiting” or “finished” UI states while the agent is actively working, including after permission prompts, answered questions, and during long resumed turns.

### DIFF
--- a/internal/session/claude_name.go
+++ b/internal/session/claude_name.go
@@ -329,52 +329,15 @@ func transcriptContainsUUID(path, uuid string) bool {
 // firstTranscriptTimestamp returns the timestamp of the earliest timestamped JSONL
 // entry in the transcript at path, or the zero time if it's missing/unreadable or
 // has no timestamped entries.
-func firstTranscriptTimestamp(path string) time.Time { return scanTranscriptTimestamp(path, false) }
+func firstTranscriptTimestamp(path string) time.Time {
+	return scanTranscriptTimestamp(path, false, false)
+}
 
 // lastTranscriptTimestamp returns the timestamp of the latest timestamped JSONL
 // entry in the transcript at path, or the zero time. It scans the whole file;
 // transcripts are small and this runs only on the rare session-id-change path.
-func lastTranscriptTimestamp(path string) time.Time { return scanTranscriptTimestamp(path, true) }
-
-// scanTranscriptTimestamp walks the JSONL transcript at path and returns the
-// first (last=false) or last (last=true) entry timestamp it can parse.
-func scanTranscriptTimestamp(path string, last bool) time.Time {
-	if path == "" {
-		return time.Time{}
-	}
-	f, err := os.Open(path)
-	if err != nil {
-		return time.Time{}
-	}
-	defer f.Close()
-
-	scanner := bufio.NewScanner(f)
-	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024) // 1MB max line
-
-	var result time.Time
-	for scanner.Scan() {
-		line := scanner.Text()
-		if !strings.Contains(line, `"timestamp"`) {
-			continue
-		}
-		var entry struct {
-			Timestamp string `json:"timestamp"`
-		}
-		if err := json.Unmarshal([]byte(line), &entry); err != nil || entry.Timestamp == "" {
-			continue
-		}
-		// RFC3339Nano: Claude transcripts stamp millisecond precision (e.g.
-		// "...:04.226Z"). The layout also parses entries with no fractional part.
-		ts, err := time.Parse(time.RFC3339Nano, entry.Timestamp)
-		if err != nil {
-			continue
-		}
-		if !last {
-			return ts
-		}
-		result = ts
-	}
-	return result
+func lastTranscriptTimestamp(path string) time.Time {
+	return scanTranscriptTimestamp(path, true, false)
 }
 
 // lastLeadTranscriptTimestamp returns the timestamp of the last lead-conversation
@@ -384,6 +347,13 @@ func scanTranscriptTimestamp(path string, last bool) time.Time {
 // out-of-pane tiebreaker for the between-bursts frame where the pane is
 // indistinguishable from finished (see conversationActivePastHook).
 func lastLeadTranscriptTimestamp(path string) time.Time {
+	return scanTranscriptTimestamp(path, true, true)
+}
+
+// scanTranscriptTimestamp walks the JSONL transcript at path and returns the first
+// (last=false) or last (last=true) entry timestamp it can parse. When excludeSidechain
+// is set, sub-agent (isSidechain:true) entries are skipped.
+func scanTranscriptTimestamp(path string, last bool, excludeSidechain bool) time.Time {
 	if path == "" {
 		return time.Time{}
 	}
@@ -409,12 +379,17 @@ func lastLeadTranscriptTimestamp(path string) time.Time {
 		if err := json.Unmarshal([]byte(line), &entry); err != nil || entry.Timestamp == "" {
 			continue
 		}
-		if entry.IsSidechain {
+		if excludeSidechain && entry.IsSidechain {
 			continue
 		}
+		// RFC3339Nano: Claude transcripts stamp millisecond precision (e.g.
+		// "...:04.226Z"). The layout also parses entries with no fractional part.
 		ts, err := time.Parse(time.RFC3339Nano, entry.Timestamp)
 		if err != nil {
 			continue
+		}
+		if !last {
+			return ts
 		}
 		result = ts
 	}
--- a/internal/session/claude_name.go
+++ b/internal/session/claude_name.go
@@ -376,3 +376,47 @@ func scanTranscriptTimestamp(path string, last bool) time.Time {
 	}
 	return result
 }
+
+// lastLeadTranscriptTimestamp returns the timestamp of the last lead-conversation
+// (non-sidechain) entry in the transcript at path, or the zero time if it's
+// missing/unreadable or has no such entries. Sub-agent (sidechain) entries are
+// skipped so a still-running sub-agent can't mask a finished lead turn. Used as the
+// out-of-pane tiebreaker for the between-bursts frame where the pane is
+// indistinguishable from finished (see conversationActivePastHook).
+func lastLeadTranscriptTimestamp(path string) time.Time {
+	if path == "" {
+		return time.Time{}
+	}
+	f, err := os.Open(path)
+	if err != nil {
+		return time.Time{}
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024) // 1MB max line
+
+	var result time.Time
+	for scanner.Scan() {
+		line := scanner.Text()
+		if !strings.Contains(line, `"timestamp"`) {
+			continue
+		}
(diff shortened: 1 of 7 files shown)


internal/session/session.go
```diff
@@ -270,11 +307,26 @@ func (s *Session) UpdateHookStatus(hs *HookStatus, resolveRotation bool) bool {
 	if hs.SessionID != "" {
 		s.mu.RLock()
 		owner = s.ownerSessionID
+		forkParent := s.forkParentID
 		projectPath := s.ProjectPath
 		negCached := owner != "" && s.rotRejectOwner == owner && s.rotRejectForeign == hs.SessionID
 		s.mu.RUnlock()
 		if owner != "" && hs.SessionID != owner {
 			switch {
+			case forkParent != "" && owner == forkParent:
```
hayke102, line 349: **Fork-adoption branch bypasses the nested-child guards — and isn't cleared on restart.**

This `case` runs *before* `negCached` and `isSessionRotation`, so within the pre-divergence window (`owner == forkParent`) it adopts the first foreign `SessionID` unconditionally, with no check that it actually descends from the parent. The "no nested child can race this" comment is an unenforced timing assumption — a sub-`claude` spawned before the fork's first prompt (eval harness / startup hook inheriting `FLEET_INSTANCE_ID`) would be force-adopted, clobbering `ClaudeSessionID`/resume-id — the exact case `isSessionRotation` + the neg-cache exist to reject.

It's reachable today via restart: `clearHookState()` (session.go:488) resets `ownerSessionID` and the rotation neg-cache but **not** `forkParentID`, and `Restart()`/`RespawnClaude()` don't call `Start()`. So restarting an un-diverged fork leaves `forkParentID` set, re-claims `owner == parent` on the next SessionStart, and re-arms this branch — the next nested-child hook gets adopted. Suggest clearing `forkParentID` in `clearHookState()`, and/or verifying the adopted id against the parent transcript before adopting here.

internal/session/session.go
```diff
@@ -723,6 +796,17 @@ func (s *Session) applyHookWaiting(paneContent string, paneStatus Status, log *s
 	// - If user interrupts/escapes, no hook fires — pane override catches this
 	// - Content change detection below handles the gap if hooks are delayed
 	if paneStatus == StatusFinished {
+		if convActive {
```
hayke102, line 873: **The tiebreaker latches: once `hookOverriddenAt` is set, this branch is never reached again.**

The overridden-waiting branch above (`if !s.hookOverriddenAt.IsZero() && s.hookOverriddenAt.Equal(s.hookUpdatedAt)`, ~L767) settles a finished pane *without* consulting `convActive`. So if `convActive` is momentarily false on the first finished frame — the resumed lead entry hasn't flushed yet, or sits within `conversationActiveSkew` of the hook — the non-`convActive` path below sets `hookOverriddenAt = hookUpdatedAt` and flips to Finished. Every later tick of the *same* waiting hook (a granted permission fires no new hook to reset the flag) then takes the overridden branch and keeps settling Finished — so the session still shows "finished" for the rest of a silent between-bursts stretch, the flash this PR targets, just narrowed by one tick. Consider consulting `convActive` in the overridden-waiting `StatusFinished` case too (or not latching `hookOverriddenAt` when the transcript may still go active).

internal/session/session.go
```diff
@@ -571,6 +628,16 @@ func (s *Session) updateStatusFromHook(oldStatus Status, hookStatus string, hook
...
 	}
 
+	// Out-of-pane tiebreaker for the waiting path: a between-bursts frame (the agent
+	// resumed after a granted permission prompt, but the activity line is momentarily
+	// gone) is indistinguishable from a finished idle prompt, so consult the
+	// transcript. Read here, before the lock — conversationActivePastHook does disk
+	// I/O and must not run under s.mu. Gated to the only frame it matters on.
+	convActive := false
+	if hookStatus == "waiting" && paneStatus == StatusFinished {
+		convActive = s.conversationActivePastHook()
```
hayke102: **Full transcript scan on the ~500ms hot path, no caching, and the convActive path never settles.**

`fastPassSessions` includes `StatusWaiting`/`StatusRunning`, so for a waiting-hook + idle-pane session this calls `conversationActivePastHook` → `lastLeadTranscriptTimestamp`, which opens and line-scans the entire JSONL to EOF every cycle (~2×/sec). And because the `convActive`-true branch in `applyHookWaiting` deliberately doesn't set `hookOverriddenAt` ("re-evaluate each tick"), a multi-minute turn becomes hundreds of full re-reads of a growing multi-MB transcript. `ReadClaudeSessionName` throttles to ~30s; this has no equivalent guard. An `os.Stat` size/mtime short-circuit (skip the scan when the file is unchanged since the last read) would eliminate nearly all of it.

internal/session/claude_name.go
```diff
@@ -376,3 +376,47 @@ func scanTranscriptTimestamp(path string, last bool) time.Time {
...
 	return result
 }
+
+// lastLeadTranscriptTimestamp returns the timestamp of the last lead-conversation
+// (non-sidechain) entry in the transcript at path, or the zero time if it's
+// missing/unreadable or has no such entries. Sub-agent (sidechain) entries are
+// skipped so a still-running sub-agent can't mask a finished lead turn. Used as the
+// out-of-pane tiebreaker for the between-bursts frame where the pane is
+// indistinguishable from finished (see conversationActivePastHook).
+func lastLeadTranscriptTimestamp(path string) time.Time {
```
hayke102, line 349: Nit (reuse): this duplicates ~40 lines of `scanTranscriptTimestamp` — same open, `scanner.Buffer(64KB/1MB)`, `"timestamp"` prefilter, `RFC3339Nano` parse loop — differing only by the `IsSidechain` skip. `scanTranscriptTimestamp` already has a `last bool` seam; adding an `excludeSidechain bool` (or a small predicate) there would let one loop serve `firstTranscriptTimestamp`/`lastTranscriptTimestamp`/lead-only, keeping the scan contract in one place.

internal/session/scenario_test.go
```diff
@@ -483,6 +493,49 @@ func TestScenarioStaleOverriddenWaitingSettlesFinished(t *testing.T) {
...
+	// resumed, but the hook is still "waiting". Modern Claude keeps the ❯ box on
+	// screen while working, so a between-bursts frame (activity line momentarily gone)
+	// reads as a finished idle prompt and used to flip the session to finished. The
+	// transcript tiebreaker (ConvActive) proves the lead turn is still appending past
+	// the hook → stay running. HookAge backdates the hook so trustPane engages (the
+	// real bug had a >1m stale hook).
+	runScenario(t, Scenario{
+		Name: "stale waiting hook + active transcript: between-bursts idle frames stay running",
+		Events: []ScenarioEvent{
+			{At: 0, Hook: "waiting", SessionID: "owner-aaa", HookAge: 90 * time.Second, Pane: "@fixture:pane_finished_idle_prompt.txt", ConvActive: new(true)},
```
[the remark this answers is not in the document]
hayke102: Valid on Go 1.26 — the `new(expr)` spec change makes `new(true)` allocate a `*bool` initialized to `true`. go.mod is `go 1.26.0`; verified `go build` + these tests pass on go1.26.2. Intentional choice over a boolPtr helper, so leaving as-is.

internal/session/session.go
hayke102: Fixed 2a in 499bc1c — `clearHookState()` now clears `forkParentID`, so a restarted un-diverged fork no longer re-claims `owner==parent` and re-arms the adopt branch (regression test `TestClearHookStateClearsForkParent`). Left 2b (transcript descent check) out: the residual vector needs a nested `claude` firing a hook inside the sub-second pre-divergence window — narrow enough that it's not worth transcript I/O on the adoption path; documented it in a code comment instead.

internal/session/session.go
hayke102: Fixed in 499bc1c — the overridden-waiting branch now re-checks `convActive` before settling, so a momentarily-false tiebreaker on the latching frame no longer pins the turn to finished; it recovers to running once the lead entry flushes. Regression test `TestScenarioOverriddenWaitingTranscriptUnlatchesRunning`.

internal/session/session.go
hayke102: Fixed in 499bc1c — `conversationActivePastHook` now does an `os.Stat` size/mtime short-circuit and only runs the full JSONL scan when the transcript actually grew; the last lead timestamp is cached per session.

internal/session/claude_name.go
hayke102: Fixed in 499bc1c — folded `lastLeadTranscriptTimestamp` into `scanTranscriptTimestamp` via an `excludeSidechain bool` param; first/last/lead-only now share the one loop (~40 lines removed).

PR #134, opened 2026-06-16
feat: add OpenCode as a third agent

hayke102:
## What

Adds **OpenCode** ([opencode.ai](https://opencode.ai)) as a third agent alongside Claude and Codex — selectable in the `A` new-session picker or via `default_agent: "opencode"`.

## Status detection (no pane scraping)

OpenCode exposes a rich event bus, so status is plugin-driven — cleaner than Codex:

- When `opencode` is on PATH, fleet writes a generated TS plugin to `~/.config/opencode/plugin/fleet-status.ts` (`InjectOpenCodePlugin`, run at bootstrap, resolved fleet binary path baked in, idempotent).
- The plugin forwards OpenCode bus events to the agent-neutral `fleet hook-handler` via **`spawnSync`**:
  - `session.status{busy}` → **running**
  - `session.idle` → **finished**
  - `permission.asked` → **waiting**
- `spawnSync` (not async spawn) is deliberate: it blocks until the status is written, so the final status flushes before OpenCode exits and ordering is preserved — matching how Claude/Codex hooks block synchronously.
- **Sub-agent filtering:** sub-agent sessions carry a `parentID`; the plugin ignores their lifecycle events so they don't flip the root session's status.
- `UpdateStatus` routes OpenCode through the shared `applyHookStatus` (renamed from `applyCodexHookAtRest`); no pane heuristics run; sessions start idle until the first event.

## Agent + UI

- `internal/agent`: `opencode` binary; launch / `opencode --session <id>` / `opencode --session <id> --fork` (fork is a modifier, not a subcommand).
- `A`-key picker is now a 3-way cycle (Claude / Codex / OpenCode); Settings → default agent includes OpenCode.
- Sidebar glyph `△` (width-1, Geometric Shapes — distinct from `✻`/`◇`).
- `default_agent: "opencode"` accepted by config.

## Design decision: ship as-is

OpenCode defaults to **allow-all** permissions (verified empirically), so `permission.asked` only fires when the user opts into `permission: ask`. This PR matches OpenCode's default and does **not** seed `ask` — running/finished work always; Waiting + quick-approve light up only when the user has set `ask`.

## Why a plugin (not declarative hook JSON like Codex)

OpenCode's hook mechanism is a JS/TS plugin, not a declarative shell-command config, so fleet ships a generated `.ts` file rather than merging JSON. No dir-trust seeding is needed (OpenCode has no trust gate). No `SessionEnd` → `dead` comes from tmux pane-death (same as Codex).

## Testing

- Unit tests: agent launch/parse/glyph, injector output + idempotency + path-rebake.
- `go test ./...` green; `make build` green.
- E2E-verified against a real `opencode` 1.14.33 session driving this branch's `build/fleet`: `running → finished`, `permission.asked → waiting`, and sub-agent events filtered.

## Notes / follow-ups

- Auto-naming for OpenCode (reading its session title) is not wired yet — it falls back to the prompt heuristic for now.

## Release Notes

* **New Features**
  * OpenCode is now available as a third selectable agent alongside Claude and Codex
  * Agent selection now cycles through all available agents instead of toggling between two
  * Real-time status monitoring for OpenCode sessions
  * OpenCode can be set as the default agent
  * Fork-to-worktree operations limited to Claude-based sessions

* **Documentation**
  * Updated to document OpenCode as a first-class agent platform


internal/agent/agent.go
```diff
@@ -81,6 +100,17 @@ func (t Type) BuildLaunchCmd(o LaunchOpts) string {
 		}
 	}
 
+	if t == OpenCode {
+		switch {
+		case o.ForkID != "":
+			return fmt.Sprintf("opencode --session %s --fork", o.ForkID)
```
hayke102, line 106: **`Shift+F` lets OpenCode reach this broken fork path.** `forkToWorktreeSelected` (app.go:2483) only rejects `agent.Codex`, so an OpenCode session falls through into the Claude-only fork flow: it stages the parent's transcript from `~/.claude/projects` (which never exists for OpenCode), then launches this `opencode --session <id> --fork` against an id that was never staged where OpenCode looks — a mis-seeded fork. The rejection message there also says "fork this **Codex** session", which is wrong for OpenCode.

Suggest widening that guard from `s.Agent == agent.Codex` to `s.Agent != agent.Claude`.

internal/hooks/opencode_hooks.go
```diff
@@ -0,0 +1,130 @@
...
+const openCodePluginTemplate = `// fleet status plugin — auto-generated by fleet; do not edit.
+// Reports OpenCode session status to fleet via the fleet hook-handler.
+import { spawnSync } from "node:child_process"
+
+const FLEET_BIN = %s
+const subagents = new Set()
+
+function notify(event, sessionID) {
+  try {
+    spawnSync(FLEET_BIN, ["hook-handler", "--fleet-hook"], {
```
hayke102, line 65: **`notify()` spawns a subprocess for *all* opencode usage, not just fleet sessions.** The plugin is installed globally (`~/.config/opencode/plugin`), so this `spawnSync(fleet hook-handler ...)` fires on every `session.busy`/`session.idle`/`permission.asked` event for any standalone opencode session the user runs — only for the Go side to boot and discover `FLEET_INSTANCE_ID` is unset and return.

A one-liner at the top of `notify()` avoids the fork+exec entirely:
```js
if (!process.env.FLEET_INSTANCE_ID) return
```
The Go-side early-exit happens too late to save the spawn.

hayke102 (commented):
Review of OpenCode support — a few behavioral gaps worth a look before merge. (Skipped some pure cleanups: 3× duplicated atomic-write helper, `agentCycle` as a 4th source-of-truth for the agent list, and the stacked per-agent branches in `UpdateStatus`.)

internal/hooks/opencode_hooks.go
```diff
@@ -0,0 +1,130 @@
...
+
+export const FleetStatus = async (_ctx) => ({
+  event: async ({ event }) => {
+    const type = event && event.type
+    const props = (event && event.properties) || {}
+    // Sub-agent sessions carry a parentID; remember them so their lifecycle
+    // events never drive the root session's status in fleet.
+    if (type === "session.created" || type === "session.updated") {
+      const info = props.info || {}
+      if (info.parentID && info.id) subagents.add(info.id)
```
hayke102: **Sub-agent filtering can miss, flipping the root to a wrong status.** `subagents` is seeded only from `session.created`/`session.updated` events this plugin instance happens to observe. If a creation event is missed or late — plugin loaded mid-session (opencode resumed via `--session`), or `parentID` arrives in a later update than the creation event — the sub-agent id is never added. Its subsequent `session.idle` then passes the `subagents.has(sessionID)` guard at line 76 and is forwarded as the root, marking the root **Finished while it's still working** (the false-finish class this project has fought before).

internal/hooks/opencode_hooks.go
```diff
@@ -0,0 +1,130 @@
...
+      return
+    }
+    const sessionID = props.sessionID
+    if (!sessionID || subagents.has(sessionID)) return
+    if (type === "session.status") {
```
hayke102, line 86: **`session.error` is never forwarded.** This dispatch handles `session.status`/`session.idle`/`permission.asked` but not `session.error`. If OpenCode hits a turn-level error that doesn't kill the pane process, fleet gets no event and keeps the last status (e.g. "running" forever) — the only remaining error path is tmux pane death, which doesn't fire on a recoverable in-session error. Worth mapping `session.error` → an error/finished status.

internal/hooks/opencode_hooks.go
```diff
@@ -0,0 +1,130 @@
...
+      const st = props.status && props.status.type
+      if (st === "busy") notify("session.busy", sessionID)
+      else if (st === "idle") notify("session.idle", sessionID)
+    } else if (type === "session.idle") {
+      notify("session.idle", sessionID)
+    } else if (type === "permission.asked") {
+      notify("permission.asked", sessionID)
```
hayke102, line 95: **`permission.asked` → waiting has no return-to-running path.** There's no `permission.replied` handler, and waiting→running relies entirely on a fresh `session.status{busy}` re-firing after the reply. But the session was already busy when permission was requested, so OpenCode may not re-emit `busy` on approval — leaving fleet **stuck on Waiting**.

Note: the `session.go` comment and the project memory both state the plugin reports this waiting→running transition, but no code path actually implements it. Consider forwarding `permission.replied` → running.

internal/agent/agent.go
hayke102: Fixed — widened the guard from `== agent.Codex` to `!= agent.Claude` so OpenCode no longer falls into the Claude-only worktree-staging path, and made the message agent-neutral ("use 'f' to fork this session in place"). Plain `f` already routes OpenCode through `BuildLaunchCmd` → `opencode --session <id> --fork` correctly.

internal/hooks/opencode_hooks.go
hayke102: Fixed — `notify()` now early-returns when `FLEET_INSTANCE_ID` is unset, so a standalone (non-fleet) opencode session that loads this global plugin never forks fleet at all.

internal/hooks/opencode_hooks.go
hayke102: Fixed — replaced the session.created/updated exclude-set with a rootId latch: the root's turn goes busy before any sub-agent spawns, so the first sessionID the plugin sees is the root, and only its events are forwarded. There's no longer any dependency on observing a sub-agent's creation event, so the missed/late-creation false-finish can't occur.

internal/hooks/opencode_hooks.go
hayke102: Fixed — the plugin now forwards `session.error`, mapped to a transient `error` status in `applyHookStatus` (StatusError, Acknowledged reset). It self-corrects on the next running/finished hook, so a recoverable in-session error no longer pins the session on its last status.

internal/hooks/opencode_hooks.go
hayke102: Fixed — the plugin now forwards `permission.replied` → running, so Waiting clears after an approval even when OpenCode doesn't re-emit session.status{busy}. The next session.idle settles it to finished. The session.go comment is now actually backed by code.

Issue #139, opened 2026-06-21
Ctrl+K command palette shows an empty screen under some tmux/terminal key setups


denniszag:
## Summary

Pressing **Ctrl+K** to open the command palette doesn't work — instead of the palette overlay, the screen goes blank/empty. This happens when the terminal or tmux configuration intercepts/transforms `Ctrl+K` before it reaches fleet (e.g. a tmux key binding or a terminal that maps `Ctrl+K` to clear/kill-line), so fleet never receives the keystroke (or receives something else) and the palette never renders.

## Steps to reproduce

1. Run fleet inside a terminal/tmux where `Ctrl+K` is bound to something else (e.g. clear scrollback / `kill-line`).
2. Press `Ctrl+K`.

## Expected

The command palette overlay opens over the sidebar/preview.

## Actual

The palette does not open — the screen shows an empty/blank state instead.

## Environment

- fleet: v2.11.0
- macOS, tmux

## Notes

- Related to the existing Ctrl+K / Cmd+K mapping guidance in the docs (CLAUDE.md / README mention mapping `Cmd+K → Ctrl+K` in terminal prefs).
- Palette keybinding is defined in `internal/ui/keybindings.go` (`Ctrl+K`) and handled in `internal/ui/command_palette.go`.
- Possible directions: allow a configurable keybinding for the command palette, document the conflicting tmux/terminal bindings more prominently, or detect the conflict and surface a hint.

denniszag:
Possibly related to the popular [gpakosz/.tmux](https://github.com/gpakosz/.tmux) (oh-my-tmux) configuration — it rebinds a number of keys, so users running it may have `Ctrl+K` (or its surrounding bindings) intercepted before fleet receives the keystroke. Worth reproducing with that config installed to confirm whether it's the trigger.

denniszag:
## Root cause confirmed

Reproduced both symptoms — they're two separate, terminal-level issues:

[...]

PR #140, opened 2026-06-21
fix(ui): open Ctrl+K palette under CSI-u/extended-keys terminals

denniszag:
## Summary

Fixes the Ctrl+K command palette being unreachable in terminals (and tmux configs like [gpakosz/.tmux](https://github.com/gpakosz/.tmux)) that enable **extended-keys / modifyOtherKeys** reporting.

Under CSI-u key reporting, Ctrl+K arrives as `ESC[107;5u` instead of the legacy `0x0b` byte. Bubble Tea v1 can't decode that sequence and silently drops it, so the palette never opened. Ctrl+C / Ctrl+Q kept working because modifyOtherKeys mode 1 leaves well-known combos in legacy encoding — which is why *only* Ctrl+K appeared broken.

## Fix

On startup, fleet now requests **legacy key reporting** (disable xterm modifyOtherKeys `\x1b[>4;0m` + pop the Kitty/CSI-u keyboard protocol `\x1b[<u`) and restores the terminal's default on exit. Ctrl+K then arrives as `0x0b` and the palette opens reliably. The sequences are no-ops on terminals that don't use those modes, so this is safe unconditionally.

Implemented as a small `internal/termkeys` package, wired in around `p.Run()` in `cmd/fleet/main.go`.

## Note on Cmd+K

`Cmd+K` blanking the screen is **iTerm2's built-in `Cmd+K = Clear Buffer`** shortcut — a terminal action fleet can't intercept. Use Ctrl+K, or remap `Cmd+K → Send Hex 0x0b` in iTerm.

## Test plan

- [x] Reproduced: sending fleet a raw `ESC[107;5u` does nothing; legacy `0x0b` opens the palette.
- [x] Regression: normal Ctrl+K still opens the palette after the change (verified via tmux capture).
- [x] Unit tests for `termkeys.Disable` / `Restore` byte output + write-error path.
- [x] `make test` (554 pass), `make lint` (0 issues), `gofmt` clean.
- [ ] Manual verification in iTerm2 + gpakosz/.tmux on the reporter's setup.

Refs #139

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Command palette (Ctrl+K) in terminals and tmux where modified/extended key reporting previously prevented the action from triggering.
* **New Features**
  * Added terminal key-reporting mode controls to force a legacy-compatible key encoding while the TUI is running, and restore the previous behavior afterward.
* **Tests**
  * Expanded coverage for key-reporting mode disable/restore sequences and write-error handling.


internal/termkeys/termkeys.go
```diff
@@ -0,0 +1,51 @@
...
+func Disable(w io.Writer) error {
+	_, err := io.WriteString(w, disablePreamble)
+	return err
+}
+
+// Restore returns modifyOtherKeys reporting to the terminal's initial value.
+// Call on exit to leave the terminal as we found it.
+func Restore(w io.Writer) error {
+	_, err := io.WriteString(w, resetModifyOtherKeys)
+	return err
```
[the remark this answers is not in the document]
denniszag: Fixed in fa0bee1 — Disable now pushes Kitty flags=0 (CSI > 0 u) and Restore pops it (CSI < u), so the keyboard-protocol state is symmetric and restored on exit.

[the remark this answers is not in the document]
denniszag: Fixed in fa0bee1 — made the Kitty sequence symmetric: push flags=0 on Disable, pop on Restore (plus the modifyOtherKeys reset). Restore now balances the stack.

cmd/fleet/main.go
```diff
@@ -151,6 +152,15 @@ func runTUI() {
...
 	model.SetProgram(p)
 
+	// Force legacy key encoding so Ctrl+K (and other modified combos) reach the
+	// TUI as control bytes rather than CSI-u sequences Bubble Tea v1 can't parse.
+	// Terminals/tmux configs (e.g. gpakosz/.tmux) that enable modifyOtherKeys
+	// otherwise break the Ctrl+K command palette. Restore on exit.
+	if err := termkeys.Disable(os.Stdout); err != nil {
+		debuglog.Logger.Warn("failed to disable modifyOtherKeys", "err", err)
+	}
+	defer func() { _ = termkeys.Restore(os.Stdout) }()
```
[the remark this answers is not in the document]
denniszag: Fixed in fa0bee1 — Disable warning is now 'failed to set legacy key reporting' (it sets both modifyOtherKeys and Kitty mode), and the deferred Restore now logs a warning on failure instead of discarding the error.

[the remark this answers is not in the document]
denniszag: Fixed in fa0bee1 — the deferred Restore now logs a Warn on non-nil error instead of discarding it.

internal/termkeys/termkeys_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+type writeError struct{}
+
+func (*writeError) Error() string { return "write failed" }
+
+func TestDisablePropagatesWriteError(t *testing.T) {
+	if err := Disable(errWriter{}); err == nil {
+		t.Error("expected error from failing writer, got nil")
+	}
+}
```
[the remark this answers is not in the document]
denniszag: Fixed in fa0bee1 — added TestRestorePropagatesWriteError mirroring the Disable error-path test, and updated the byte assertions for the new symmetric sequences.

cmd/fleet/main.go
```diff
@@ -151,6 +152,19 @@ func runTUI() {
...
 
+	// Force legacy key encoding so Ctrl+K (and other modified combos) reach the
+	// TUI as control bytes rather than CSI-u sequences Bubble Tea v1 can't parse.
+	// Terminals/tmux configs (e.g. gpakosz/.tmux) that enable modifyOtherKeys
+	// otherwise break the Ctrl+K command palette. Restore on exit.
+	if err := termkeys.Disable(os.Stdout); err != nil {
+		debuglog.Logger.Warn("failed to set legacy key reporting", "err", err)
+	}
+	defer func() {
+		if err := termkeys.Restore(os.Stdout); err != nil {
```
hayke102: **`Restore` is skipped on the error-exit path.** This cleanup is only `defer`red, but the branch just below (`if _, err := p.Run(); err != nil { ... os.Exit(1) }`) exits via `os.Exit`, which does **not** run deferred functions. So if `p.Run()` returns an error (a Bubble Tea loop/cmd panic surfaces here as `ErrProgramKilled`), the user's terminal/tmux is left with `modifyOtherKeys` disabled and the Kitty keyboard-stack entry popped — extended-key reporting stays broken for *other* apps in that terminal until they manually reset it. The gpakosz/.tmux users this PR targets are exactly the ones who'd notice.

Call `Restore(os.Stdout)` explicitly before `os.Exit(1)` (and ideally on signal-kill too), in addition to the defer.

internal/ui/help.go
```diff
@@ -1,62 +1,191 @@
...
+// key label (currently "= = then digit", 14 cells); padding/truncating to it
+// keeps every binding on exactly one row so the height math stays exact.
+const helpKeyColWidth = 14
+
+// helpChromeHeight is the vertical space consumed by everything around the
+// scrollable body: dialog border (2) + padding (2) + title block (title +
+// blank = 2) + footer block (blank + footer = 2). The body therefore gets
+// h.height - helpChromeHeight rows, which keeps the rendered box no taller
+// than the terminal so lipgloss.Place never clips it.
+const helpChromeHeight = 8
```
hayke102: **This whole `help.go` rewrite duplicates already-merged #150** (`2bf4691`, "keep help overlay (?) reachable on short terminals"), which shipped the same scroll feature with a different design (newspaper columns + `MaxHeight(h.height)` clamp). This branch was cut from `facea81` (pre-#150), so the three-dot diff hides the collision — but a real merge produces hard conflicts in **both** `help.go` and `help_test.go` (add/add), redeclares `clampInt`, and reverts #150's `MaxHeight` guard + column layout. That's a user-visible regression from released 2.12.1. **Recommend dropping the help.go / help_test.go / changelog changes from this PR entirely** and keeping only the termkeys + main.go Ctrl+K fix.

If the rewrite is kept anyway, it has several latent bugs the merged version doesn't:
- `helpChromeHeight = 8` floors the box at 9 rows, so it still clips at terminal heights ≤ 8 (the exact short-terminal case the changelog promises to fix).
- No `MaxWidth(h.width)` clamp → overflows on narrow terminals (e.g. width 12 → 30-cell box).
- `View()` mutates `h.scroll` during render — resizing smaller while open permanently shrinks the saved offset.
- `innerWidth` ignores the footer, so the `X-Y of Z` scroll indicator is truncated to a stub on ~50-col terminals.
- `TestHelpOverlayNeverOverflows` sweeps height from 9 up and width from 40 up, so it never exercises the ranges where these overflow — the guard is green while the bugs are live.

internal/termkeys/termkeys.go
```diff
@@ -0,0 +1,58 @@
+// Package termkeys controls terminal key-reporting modes that affect how
+// modified keys (e.g. Ctrl+K) reach the TUI.
+//
+// Some terminals — and tmux configs such as gpakosz/.tmux, which enable
+// `extended-keys` for iTerm/xterm — report "other" modified keys using the
+// CSI-u / xterm modifyOtherKeys encoding instead of the legacy control byte.
+// A Ctrl+K then arrives as `ESC[107;5u` (or `ESC[27;5;107~`) rather than the
+// 0x0b byte. Bubble Tea v1 cannot decode those sequences, so it silently drops
+// the keystroke: the Ctrl+K command palette appears to do nothing. (Ctrl+C and
+// Ctrl+Q still work because modifyOtherKeys mode 1 keeps well-known combos in
```
hayke102: **This Ctrl+Q rationale is factually wrong.** The claim that "Ctrl+Q still works because modifyOtherKeys mode 1 keeps well-known combos in their legacy encoding" is contradicted by our own `internal/tmux/pty.go` (`ctrlQSequences`, ~lines 30-31), which has to hand-match `\x1b[113;5u` (CSI-u) and `\x1b[27;5;113~` (modifyOtherKeys) precisely *because* Ctrl+Q is **not** delivered as the legacy `0x11` byte under those modes. A future maintainer trusting this comment could remove that CSI-u Ctrl+Q matching and break the detach key in extended-keys terminals. Suggest correcting the parenthetical.

internal/termkeys/termkeys.go
```diff
@@ -0,0 +1,58 @@
...
+// the palette opens reliably. Alt+<digit> slot bindings benefit the same way.
+package termkeys
+
+import "io"
+
+// Escape sequences that select legacy vs. enhanced key reporting.
+const (
+	// disableModifyOtherKeys sets the xterm modifyOtherKeys resource (XTMODKEYS,
+	// Ps=4) to mode 0 (off): keys are reported with their legacy encoding.
+	disableModifyOtherKeys = "\x1b[>4;0m"
```
hayke102: **Reuse `x/ansi` instead of hand-rolling these escape consts.** The PR already imports `github.com/charmbracelet/x/ansi` (used as `ansi.Truncate` in help.go), and it exports all of these by name: `ansi.DisableModifyOtherKeys`, `ansi.ResetModifyOtherKeys`, `ansi.PushKittyKeyboard(0)` / `ansi.DisableKittyKeyboard`, `ansi.PopKittyKeyboard(1)`. Re-deriving the raw bytes by hand (with a test asserting the exact magic strings) makes any future protocol nuance or typo this package's burden and lets the consts drift from the canonical definitions.

changelog/unreleased/help-overflow-scroll.md
```diff
@@ -0,0 +1,4 @@
+---
+type: fixed
```
hayke102: **Duplicate changelog fragment.** Merged #150 already left `changelog/unreleased/fix-help-overflow.md` with near-identical wording ("no longer gets cut off on short terminals … scrolls (↑↓)"). After this merges, the unreleased dir holds two `type: fixed` fragments for the same fix, so the next release roll-up emits two redundant Fixed bullets. Moot if the help.go changes are dropped per the other comment.

hayke102 (commented):
Automated review (xhigh). 5 findings worth flagging — the two big ones: the `termkeys.Restore` cleanup is skipped on the error-exit path, and the `help.go` scroll rewrite duplicates/conflicts with already-merged #150.

cmd/fleet/main.go
hayke102: Fixed in 82cfb9d — the p.Run() error path now calls restoreKeys() before os.Exit(1), so key reporting is restored even though os.Exit skips deferred funcs.

cmd/fleet/main.go
```diff
@@ -151,6 +152,19 @@ func runTUI() {
...
+			debuglog.Logger.Warn("failed to restore key reporting", "err", err)
+		}
+	}()
```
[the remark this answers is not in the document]
hayke102: Fixed in 82cfb9d — Restore is now registered only when Disable succeeds (fail-safe) and is idempotent, and it's also called explicitly before os.Exit on the error path.

internal/termkeys/termkeys.go
hayke102: Fixed in 82cfb9d — corrected the doc: Ctrl+Q keeps working under these modes only because pty.go's ctrlQSequences decodes its CSI-u/modifyOtherKeys forms, not because modifyOtherKeys preserves legacy encoding.

internal/termkeys/termkeys.go
hayke102: Fixed in 82cfb9d — now reuse ansi.ResetModifyOtherKeys, ansi.DisableKittyKeyboard, and ansi.PopKittyKeyboard(1). Kept the literal \x1b[>4;0m for the modifyOtherKeys-off case: x/ansi's only equivalent (ansi.DisableModifyOtherKeys) is deprecated so staticcheck would flag it, and its suggested replacement resets to default rather than forcing off.

internal/ui/help.go
hayke102: Resolved by merging master (8a628c4) — help.go and help_test.go are reverted to the shipped #150; this branch no longer touches the help overlay.

internal/ui/help.go
```diff
@@ -1,62 +1,191 @@
...
+func (h *HelpOverlay) innerWidth(lines []string) int {
+	longest := 0
+	for _, l := range lines {
+		if w := lipgloss.Width(l); w > longest {
+			longest = w
+		}
+	}
+	maxInner := max(h.width-6, helpMinInnerWidth)
+	return clampInt(longest, helpMinInnerWidth, maxInner)
+}
```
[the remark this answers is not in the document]
hayke102: Resolved — the help.go rewrite was dropped in the master merge (8a628c4); help.go is now identical to the shipped #150, so the flagged width path no longer exists.

changelog/unreleased/help-overflow-scroll.md
hayke102: Resolved — the duplicate help-overflow-scroll.md fragment was removed in the master merge (8a628c4); only fix-help-overflow.md (from #150) remains.

changelog/unreleased/help-overflow-scroll.md
```diff
@@ -0,0 +1,4 @@
...
+---
+The help screen (?) no longer gets cut off on short terminals — it now fits the available height and scrolls (↑↓) when the keybindings don't all fit.
```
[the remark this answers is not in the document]
hayke102: Resolved — the help-overflow code and its changelog fragment were dropped in the master merge (8a628c4), so this PR is now scoped to just the Ctrl+K fix.

PR #144, opened 2026-06-21
fix(status): recover stuck-running status after a missed session rotation

hayke102:
## Problem

Diagnosed from a `D`-hotkey snapshot (`/debug-status`): a session was genuinely **waiting** on a permission prompt but the TUI was stuck showing **running**.

A `/clear`-style session rotation (`24e30322` → `a3f8eaa5`, triggered by interrupting a tool call) fires its `SessionStart` hook before the new transcript flushes its first *timestamped* entry — the file opens with timestamp-less `mode`/`permission-mode`/`file-history-snapshot` header entries. The proximity-based rotation check read zero timestamps, concluded "not a rotation", and **permanently neg-cached** the `(owner, foreign)` pair. From then on:

- fleet stayed pinned to the dead owner transcript, with its in-memory hook frozen ~55 min in the past (proven by `hookAge=55m7s` in the logs).
- Every hook from the new session — including the real `PermissionRequest` (waiting) — was dropped as "foreign".
- A latched stale-waiting hook + `applyHookWaiting`'s switch having no `StatusWaiting` case left the status pinned to `running` even with the prompt structurally on screen.

## Fix

Two compounding fixes (either resolves the case; together they're defense-in-depth):

1. **`sessionRotationVerdict` (was `isSessionRotation`) is now tri-state.** When a transcript timestamp can't be read yet it returns `rotationUnknown`, and the caller **defers** (rejects without neg-caching) so the next worker cycle re-adopts once the transcript flushes — instead of pinning the session to the dead owner id forever.

2. **`applyHookWaiting`'s latched-override switch gains a `StatusWaiting` case** so a real permission prompt on the pane recovers the status to `waiting`, rather than staying pinned to the last `running` it saw when a frozen hook never resets the latch.

## Tests

- `TestUpdateHookStatusDefersUndecidedRotation` — header-only transcript on first hook → deferred (not adopted, not neg-cached); after the timestamp flushes, a later hook adopts it.
- `TestScenarioStaleWaitingLatchRecoversToWaiting` — stale waiting hook latches finished→running, then a real prompt appears → must show `waiting`.

Both verified to fail without their fix and pass with it. Full `internal/session` suite green with `-race`; lint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed sessions sometimes getting stuck showing a `"running"` status after rotation when they’re actually waiting for a permission prompt.
  * Improved recovery from stale “waiting” signals so the session transitions back to waiting when the prompt reappears.
  * Added safer rotation decision handling so undecided cases are deferred and retried rather than permanently rejected.
* **Tests**
  * Added coverage for deferred/undecided rotation behavior, including retry and negative-caching limits.
  * Added scenario coverage for recovering from a stale waiting latch to the correct waiting state.
* **Chores**
  * Added an unreleased changelog entry documenting the fix.

### DIFF
--- a/internal/session/ownership_test.go
+++ b/internal/session/ownership_test.go
@@ -100,7 +100,7 @@ func TestUpdateHookStatusIgnoresNestedChild(t *testing.T) {
 // `claude --resume <parent> --fork-session` reports the PARENT's id at SessionStart
 // and only switches to its own id on the first prompt. The fork transcript copies
 // the parent's history verbatim (same OLD first timestamp) and carries no
-// logicalParentUuid, so BOTH isSessionRotation signals miss — leaving the fork
+// logicalParentUuid, so BOTH sessionRotationVerdict signals miss — leaving the fork
 // pinned to the parent id, so forking it re-forks the parent. With forkParentID
 // known, the fork's own id must be adopted deterministically.
 func TestUpdateHookStatusAdoptsForkDivergence(t *testing.T) {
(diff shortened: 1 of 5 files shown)


internal/session/session.go
```diff
@@ -365,23 +365,35 @@ func (s *Session) UpdateHookStatus(hs *HookStatus, resolveRotation bool) bool {
...
-				// rejection only when we actually evaluated it (worker path), so a
-				// persistent foreign id doesn't rescan transcripts every cycle; the
-				// UI path leaves the verdict to the worker.
-				if resolveRotation {
+				switch sessionRotationVerdict(projectPath, owner, hs.SessionID) {
+				case rotationYes:
+					debuglog.Logger.Info("adopting rotated claude session",
+						"id", s.ID, "old", owner, "new", hs.SessionID)
+					// fall through to the adoption block below
+				case rotationUnknown:
```
hayke102: `rotationUnknown` is correctly rejected without neg-caching so a later cycle can re-evaluate once the transcript flushes — but there's no terminal case / retry cap. For the documented `/clear` race this self-resolves in a cycle or two, and a normal nested child still reaches `rotationNo` (its transcript has timestamps). The gap is a pair that stays *permanently* undecidable: if the owner transcript is unreadable/missing then `ownerLast.IsZero()` in `sessionRotationVerdict` (claude_name.go:266) yields `rotationUnknown` every cycle, so it never neg-caches and re-opens + whole-file-scans up to three transcripts on every worker pass (500ms–2s) indefinitely — the exact rescan the neg-cache exists to prevent. Consider a bounded undecided-retry count keyed on (owner, foreign) that falls back to neg-cache after N cycles.

internal/session/session.go
```diff
@@ -365,23 +365,35 @@ func (s *Session) UpdateHookStatus(hs *HookStatus, resolveRotation bool) bool {
...
+				// block the render loop. Defer to the next worker cycle (~500ms), which
+				// re-evaluates with resolveRotation=true.
+				return false
 			default:
-				// Foreign (nested child), or deferred from the UI path. Cache the
-				// rejection only when we actually evaluated it (worker path), so a
-				// persistent foreign id doesn't rescan transcripts every cycle; the
-				// UI path leaves the verdict to the worker.
-				if resolveRotation {
+				switch sessionRotationVerdict(projectPath, owner, hs.SessionID) {
```
hayke102: Minor maintainability trap: `rotationYes` is the only branch here that must *not* return — it works only by being the last statement before both nested switches end, falling through to the adoption block (~L401). The `// fall through to the adoption block below` is a comment, not Go `fallthrough`; every other branch returns false. Adding a statement after the inner switch, or reordering cases, would silently route an adopted rotation to a no-op (hook stays frozen on the dead owner) with **no compiler error**. An explicit `if v := sessionRotationVerdict(...); v != rotationYes { … return false }` would make the single proceed-path load-bearing instead of leaning on the comment.

internal/session/session.go
hayke102: Fixed — added a bounded undecided-retry counter keyed on (owner, foreign): it defers while the pair is undecidable but falls back to the neg-cache after rotationUndecidedRetryCap (10) consecutive cycles, so a permanently unreadable/missing owner transcript can no longer rescan transcripts on every worker pass. Reset on pair change. Added TestUpdateHookStatusCapsUndecidedRotation.

internal/session/session.go
hayke102: Fixed — replaced the inner switch with an explicit guard: `if v := sessionRotationVerdict(...); v != rotationYes { …; return false }`. Now only rotationYes reaches the adoption block, so a stray statement or reordered case can't silently route an adopted rotation to a no-op.

internal/session/ownership_test.go
```diff
@@ -185,6 +185,52 @@ func TestClearHookStateClearsForkParent(t *testing.T) {
...
+	}
+	if s.ClaudeSessionID != "owner-aaa" {
+		t.Errorf("undecided rotation prematurely adopted: %q, want owner-aaa", s.ClaudeSessionID)
+	}
+
+	// The transcript now flushes its first real (timestamped) entry, handing off within
+	// the proximity window. The deferred id must now be adopted — proving the first,
+	// undecidable attempt did NOT permanently neg-cache it.
+	writeTranscript(t, proj, "rot-bbb", ownerEnd.Add(-2*time.Millisecond), ownerEnd.Add(10*time.Minute))
+	changed := s.UpdateHookStatus(&HookStatus{Status: "running", SessionID: "rot-bbb", UpdatedAt: time.Now()}, true)
```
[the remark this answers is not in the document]
hayke102: Leaving these as-is. These IDs are the claude_session_id (a Claude-generated UUID), not fleet's `<8hex>-<unix_timestamp>` session ID — that format applies to the fleet session ID field (s.ID), not this one. Nothing in the rotation path parses the id either (ClaudeTranscriptPath just appends `.jsonl`), so the symbolic role-names are kept for readability and consistency with the other tests in this file.

internal/session/ownership_test.go
```diff
@@ -185,6 +185,89 @@ func TestClearHookStateClearsForkParent(t *testing.T) {
...
+// sessionRotationVerdict's ownerLast is always zero) must not defer forever and rescan
+// transcripts on every worker pass. After rotationUndecidedRetryCap consecutive
+// undecided cycles it falls back to the neg-cache.
+func TestUpdateHookStatusCapsUndecidedRotation(t *testing.T) {
+	t.Setenv("HOME", t.TempDir())
+	proj := t.TempDir()
+
+	// Owner claims ownership, but its transcript is never written — so the proximity
+	// check can't read ownerLast and the (owner, foreign) pair stays undecidable.
+	s := &Session{ID: "x", ProjectPath: proj, Status: StatusRunning}
```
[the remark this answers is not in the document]
hayke102: Fixed — good catch on the corrected scope. Changed the new test's `Session.ID` from "x" to a contract-shaped `abcd1234-1700000000`. (Kept the symbolic claude_session_id fixtures as you noted — those are UUIDs, not the fleet ID format.)

PR #150, opened 2026-06-22
fix(ui): keep help overlay (?) reachable on short terminals

### LINKED ISSUE #141: Help screen (?) overflows and is cut off on small terminals
## Summary

The keybindings help overlay (`?`) renders all ~37 bindings as a single tall column and then centers the box with `lipgloss.Place`. On a small/short terminal the box is taller than the screen, so the top and bottom rows are clipped and there's no way to see them (no scrolling).

## Steps to reproduce

1. Make the terminal short (e.g. ~20 rows, or a split tmux pane).
2. Press `?` to open help.

## Expected

All keybindings are reachable — either the layout fits, or the sheet scrolls.

## Actual

The cheat sheet is cut off at top and bottom; the clipped bindings can't be viewed.

## Notes

- `internal/ui/help.go` `View()` builds one line per binding then `lipgloss.Place(h.width, h.height, …)` — Place clips when the box exceeds `h.height`.

hayke102:
## Summary

Fixes #141 — the `?` keybindings help overlay rendered all ~37 bindings as one tall single column and centered it with `lipgloss.Place`, which clips the top and bottom rows (including the title) whenever the box is taller than the terminal. On a short pane the clipped bindings were unreachable.

## What changed (`internal/ui/help.go`)

- **Responsive columns** — bindings now flow into newspaper-style columns sized to the window, so a wide-enough terminal fits everything without scrolling. A shared `layout()` method computes column count from both available width and height (used by `View` and `Update` so they can't disagree).
- **Scroll fallback** — when even the widest-possible column count is still too tall (e.g. a ~20-row pane), the sheet scrolls (`↑/↓`, `PgUp/PgDn`, `g/G`) with `⋮ +N above/below` cues. Any non-scroll key still closes it; when the sheet already fits, *any* key closes (unchanged behavior).
- **Never exceeds the screen** — a `MaxWidth/MaxHeight` clamp guarantees the box fits, so the title is always visible.
- **Fixed a wrapping bug** — `DialogStyle.Width(...)` made lipgloss count horizontal padding against the content area, wrapping the longest binding (`Bind/unbind slot …`) and making it unreachable. The box now auto-sizes to its content.

## Tradeoff

Because a few descriptions are long (~47 chars), two columns only engage around ~130+ cols wide; at typical laptop widths it stays single-column and relies on scroll. Both correct and reachable at any size.

## Testing

- `make build`, `go test -race ./internal/ui/`, and `go test ./...` all pass.
- New `internal/ui/help_test.go`: asserts the box never overflows the terminal across 6 sizes (incl. the 100×20 repro), and that every binding is reachable across the scroll range.
- Rendered at 100×20 and 160×22 to confirm visually — fits, title visible, long binding intact.

### DIFF
--- a/internal/ui/help.go
+++ b/internal/ui/help.go
@@ -21,10 +21,16 @@ func NewHelpOverlay() *HelpOverlay {
 	return &HelpOverlay{}
 }
 
-func (h *HelpOverlay) Show()             { h.visible = true; h.scroll = 0 }
-func (h *HelpOverlay) Hide()             { h.visible = false }
-func (h *HelpOverlay) IsVisible() bool   { return h.visible }
-func (h *HelpOverlay) SetSize(w, ht int) { h.width = w; h.height = ht }
+func (h *HelpOverlay) Show()           { h.visible = true; h.scroll = 0 }
+func (h *HelpOverlay) Hide()           { h.visible = false }
+func (h *HelpOverlay) IsVisible() bool { return h.visible }
+
+// SetSize records the terminal size and re-clamps the scroll offset, so it
+// owns the scroll invariant on resize and View() can stay read-only.
+func (h *HelpOverlay) SetSize(w, ht int) {
+	h.width, h.height = w, ht
+	h.scroll = clampInt(h.scroll, 0, h.layout().maxScroll)
+}
 
 // Update scrolls when the sheet overflows; any non-scroll key closes it.
 func (h *HelpOverlay) Update(msg tea.Msg) (*HelpOverlay, tea.Cmd) {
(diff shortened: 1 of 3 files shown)


internal/ui/help.go
```diff
@@ -12,51 +13,216 @@ type HelpOverlay struct {
...
+		hint = "↑↓ scroll · any other key to close"
+	}
+	lines = append(lines, "", DimStyle.Render(hint))
 
-	box := DialogStyle.Width(40).Render(b.String())
+	// Let the box auto-size to its widest line (the padded grid rows). Forcing
+	// an explicit Width would make lipgloss count the horizontal padding against
+	// the content area and wrap the longest binding.
+	box := DialogStyle.Render(strings.Join(lines, "\n"))
+	box = lipgloss.NewStyle().MaxWidth(h.width).MaxHeight(h.height).Render(box)
```
hayke102, line 208: On very short terminals (height ≤ 9) the rendered box is taller than the screen, so `MaxHeight(h.height)` clips the bottom rows and the `↑↓ scroll · any other key to close` hint disappears — verified at 120×8 (`hintShown=false`). It still doesn't overflow the terminal, but the user loses the only cue that the sheet scrolls, in exactly the short-terminal regime issue #141 targets. Not a blocker; worth reserving the hint line before the rows (or at least noting the floor).

internal/ui/help.go
```diff
@@ -12,51 +13,216 @@ type HelpOverlay struct {
...
+		b.WriteString(cell)
+	}
+	return b.String()
+}
+
 // View renders the keybinding cheat sheet.
 func (h *HelpOverlay) View() string {
-	bindings := HelpOverlayBindings()
+	lay := h.layout()
+	h.scroll = clampInt(h.scroll, 0, lay.maxScroll)
```
hayke102: `View()` mutates the model here (`h.scroll = …`), but Bubble Tea's `View` is meant to be read-only. This re-clamp is load-bearing only because `WindowSizeMsg` → `SetSize` (app.go:490) updates width/height without re-clamping `scroll`, so a resize that shrinks `maxScroll` leaves `scroll` stale until the next render. Cleaner depth: clamp in `SetSize` where the height actually changes, so `View` stays pure and there's a single owner of the invariant.

internal/ui/help.go
```diff
@@ -12,51 +13,216 @@ type HelpOverlay struct {
...
+	for _, b := range HelpOverlayBindings() {
+		if b.Key == "" { // drop section separators; columns provide the gaps
+			continue
+		}
+		entries = append(entries, b)
+	}
+	n := len(entries)
+
+	const (
+		keyCap  = 14 // widest key label is "= = then digit"
```
hayke102: Heads-up on coupling, not a bug: `keyCap=14` is pinned to today's longest label (`= = then digit`), and `chromeV=8`/`chromeH=6` below are hand-derived from `DialogStyle`'s border+padding and the exact title/blank/hint line count in `View()`. Nothing ties them together, so if `DialogStyle` is restyled or `View()` gains/loses a line, the scroll math desyncs silently (bottom rows clipped *and* unreachable, or columns over-pack and the longest binding wraps). Also note lipgloss `.Width(keyW)` is a *minimum*, not a truncation — a key wider than 14 would overflow `colW` rather than clip. Low priority.

internal/ui/help.go
hayke102: Fixed — moved the scroll re-clamp into SetSize() so it owns the invariant on resize and View() is now read-only.

internal/ui/help.go
hayke102: Fixed — dropped keyCap (keyW is now the true widest key, so colW can't be overflowed by a wider label), and the frame overhead is now read from DialogStyle.GetVerticalFrameSize()/GetHorizontalFrameSize() instead of hand-derived constants. Net values are unchanged today, so layout output is identical — just decoupled.

internal/ui/help.go
hayke102: On reflection this is below the usable floor — at height ≤ 8 there's no room for title + a row + the hint + the rounded frame, and the main sidebar/preview UI can't render at that size either. The box still never overflows into the rest of the UI; only the scroll cue drops. Documented the floor in a comment near the MaxHeight clamp.

PR #153, opened 2026-06-22
feat: live terminal drawer (tmux control-mode + x/vt), on Bubble Tea v2

hayke102:
## Summary

A **real, live terminal drawer** — plus the Bubble Tea v2 migration it's built on.

The drawer (`` ` ``) now streams the active shell's output through a true terminal emulator in **real time**, byte- and cursor-accurate, instead of polling `tmux capture-pane` for periodic snapshots. Streaming output, colors, and full-screen TUIs (vim, htop, lazygit) render faithfully.

Developed in sequence on this branch (4 commits):
1. **Terminal drawer** — always-typing, Ctrl-chord control, split into the right/preview column (`ff24648`)
2. **Bubble Tea v2 migration** — Charm stack → `charm.land/.../v2`; cell-diff renderer (`79751e5`)
3. **Live terminal** — tmux control-mode `%output` → `charmbracelet/x/vt` emulator (`c547e3b`)
4. **Perf fix** — git status-refresh fork contention, 5 → 3 subprocesses (`c75fcfb`)

## Architecture (live terminal)

- `internal/tmux/control_output.go` — `OutputReader`: `tmux -C` attached per active shell, octal-decodes `%output`, sizes the pane to the drawer body so wrap points match.
- `internal/vterm` — insulated `x/vt` wrapper (own mutex, never `Close()`; pins the pre-1.0 dep behind a thin interface; strips screen/tmux `ESC k … ST` set-title escapes that `x/vt` would otherwise leak as visible text).
- `internal/ui/drawer.go` — `syncShellStream` lifecycle (re-point on tab switch/restart), event-driven render wake, stable-height viewport, `Ctrl+G` full-attach coordination.
- Input stays on the existing control client; only **output** is the new stream — both control mode, no pipe-pane hybrid. This mirrors iTerm2's tmux integration.

## Type of Change

- [x] New feature
- [x] Refactor (Bubble Tea v2 migration)

## Checklist

- [x] Tests pass (`make test`, incl. `-race`)
- [x] Linter passes (`make lint` — 0 issues)
- [x] Commit messages follow conventional commits
- [x] Documentation updated (CLAUDE.md drawer bullet + package map)
- [x] Changelog fragments added (`terminal-drawer`, `bubbletea-v2`, `git-refresh-fork-lock`)

## Notes for reviewers

- The **v2 migration** touches ~27 files mechanically (imports → `/v2`, `View() → tea.View`, `KeyPressMsg`, `color.Color`, `bubbletea-overlay` → Lip Gloss compositor). It's bundled here because v2's cell-diff renderer is the foundation that makes the live drawer's frequent redraws cheap.
- `x/vt` is pre-1.0 (pseudo-versioned); deliberately isolated behind `internal/vterm` so the backend is a one-file swap.

https://claude.ai/code/session_01MKjvkYtVVUiFWG1C7fqTC5

## Summary by CodeRabbit

* **New Features**
  * Added a terminal drawer (backtick) with repo-scoped live shells, streaming terminal output with preserved colors, tab controls, and full-screen attach (Ctrl+T/Ctrl+W/Ctrl+PgUp/Ctrl+PgDn/Ctrl+G).
* **Improvements**
  * Smoother, lower-flicker TUI rendering using Bubble Tea v2 cell-diff updates.
  * Reduced UI stutters by optimizing git status refresh.
  * Configurable **drawer height (rows)** with sensible defaults and clamping.
* **Documentation**
  * Updated terminal drawer specs, shortcuts, and settings (including drawer height), plus unreleased changelog/tech stack notes.


internal/tmux/tmux.go
```diff
@@ -621,36 +634,53 @@ func captureSentinel() string {
...
 	activityCache := make(map[string]int64)
 	deadCache := make(map[string]bool)
+	cmdCache := make(map[string]string)
 	for _, line := range strings.Split(strings.TrimSpace(string(output)), "\n") {
 		if line == "" {
 			continue
 		}
-		parts := strings.SplitN(line, "\t", 3)
-		if len(parts) < 3 {
+		parts := strings.SplitN(line, "\t", 4)
```
hayke102, line 654: **Cache-drop regression on the last session.** The new 4th field `pane_current_command` is empty for a dead pane, so its line is `name\t<act>\t1\t`. `strings.TrimSpace(string(output))` (line 646) strips trailing whitespace from the *whole* output — including the final line's trailing tab — so when a dead pane is **last** in `list-panes -a`, this `SplitN` yields only 3 fields, `len(parts) < 4` drops it, and that session disappears from `deadCache`/`activityCache`/`cmdCache`. `IsPaneDead`/`Exists`/`PaneCurrentCommand` then misreport it: an exited shell shows Idle, and a dead Codex/OpenCode session is no longer detected dead. The old 3-field format was immune because `pane_dead` is never empty.

Fix: drop the whole-output `TrimSpace` — the `line == ""` guard already handles blank lines.

internal/ui/drawer.go
```diff
@@ -0,0 +1,662 @@
...
+}
+
+// startShellStream tears down any existing stream and attaches a fresh control
+// reader + emulator to targetSession at size w×ht. The reader's callback runs on
+// its own goroutine: it writes into the (mutex-guarded) emulator and schedules a
+// single coalesced render wake.
+func (h *Home) startShellStream(target string, w, ht int) {
+	h.teardownShellStream()
+	term := vterm.New(w, ht)
+	reader, err := tmux.NewOutputReader(target, w, ht, func(b []byte) {
```
hayke102: **Blocking tmux I/O on the Update goroutine.** `NewOutputReader` forks `tmux -C attach` + starts a PTY, and `teardownShellStream` → `OutputReader.Close()` does `Process.Kill()` + `cmd.Wait()` (control_output.go:159-170). Both run synchronously inside `Update()` via `syncShellStream`, called from the `drawerAnimTickMsg`/`drawerTypeTickMsg` handlers (app.go:552,562), `switchTab`, `openDrawerTyping`, and `closeDrawer`.

CLAUDE.md: *“All blocking I/O (tmux, git, gh) runs in background worker goroutine, never in Bubble Tea Update().”* A slow control-process attach/exit stalls input + rendering. Consider moving attach/teardown into a `tea.Cmd`.

internal/shell/shell.go
```diff
@@ -0,0 +1,209 @@
...
+	return s.exitCode
+}
+
+// RefreshStatus recomputes the shell's status from the tmux caches and returns
+// it. Cache-only except for one bounded PaneDeadInfo call when the pane is dead
+// (to capture the exit code). Called off the render thread (status worker).
+func (s *Shell) RefreshStatus() Status {
+	ts := s.Tmux()
+	name := s.TmuxName()
+	dead := ts.IsPaneDead()
```
hayke102: **A gone session derives as Idle, not Exited — and becomes unrevivable.** After a tmux server restart/reboot, `FromRow` reconnects to a session that no longer exists: `IsPaneDead()` returns false (absent from the cache) and `PaneCurrentCommand()` returns `""`, so `DeriveStatus(false, "")` hits `idleCommands[""]` → `StatusIdle`. The drawer shows a live “idle” tab, attaches a reader to a nonexistent session (blank body), and since Enter→restart is gated on `StatusExited` (drawer.go:428), the user can't restart it — only close it.

The FromRow comment “a dead session simply renders as exited” only holds for a dead *pane*, not an *absent* session. Consider checking `ts.Exists()` here and treating absent as Exited.

internal/ui/app.go
```diff
@@ -1908,6 +2128,7 @@ func (h *Home) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		if h.controlClient != nil {
 			h.controlClient.Close()
 		}
+		h.teardownShellStream()
```
hayke102: This teardown + `finalizeAllPendingDeletes()` was added to the Ctrl+C path, but the command-palette **“Quit”** command (`case "quit": return h, tea.Quit` at app.go:5695) returns a bare `tea.Quit` and runs none of it. Quitting via Ctrl+K → Quit leaves pending session deletes (still in the 5s undo window) unfinalized — their tmux sessions are orphaned after exit — and leaks the drawer reader. Route the palette quit through the same cleanup.

internal/tmux/control_output.go
```diff
@@ -0,0 +1,189 @@
...
+	r := &OutputReader{ptmx: ptmx, cmd: cmd, target: targetSession}
+	go r.readLoop(onData)
+	debuglog.Logger.Info("tmux output reader attached", "target", targetSession, "size", fmt.Sprintf("%dx%d", w, h))
+	return r, nil
+}
+
+func (r *OutputReader) readLoop(onData func([]byte)) {
+	sc := bufio.NewScanner(r.ptmx)
+	// %output lines can be a full-screen redraw; allow a generous max token.
+	sc.Buffer(make([]byte, 0, 64*1024), 8*1024*1024)
```
hayke102, line 111: **8 MiB token cap silently freezes the stream.** A single `%output` line larger than 8 MiB (a large single-line write — control mode escapes `0x0A` as `\012`, so a burst with no real newline is one scanner token) makes `Scan()` return false with `bufio.ErrTooLong`. `readLoop` then exits, `IsClosed()` is false so only a Debug log fires, and `syncShellStream` never re-attaches (target unchanged, `shellTerm != nil`). The drawer freezes on its last frame until the tab is switched/restarted. Consider re-attaching when the scan ends while not closed.

internal/ui/app.go
```diff
@@ -3966,6 +4226,13 @@ func (h *Home) statusWorkerCycle() {
 	// prematurely.
 	tmux.RefreshSessionCache()
 
+	// Refresh shell statuses from the just-updated cache (cache-only reads,
+	// plus one bounded PaneDeadInfo on an exited pane). Shells aren't in
+	// h.sessions, so they get their own pass.
+	for _, sh := range shells {
+		sh.RefreshStatus()
```
hayke102, line 4292: **Per-exited-shell subprocess every fast cycle.** This loop runs unconditionally on the ~500ms fast pass (even when the drawer is hidden), and `RefreshStatus` (shell.go:178) calls `PaneDeadInfo()` — a fresh `tmux list-panes` exec — for every `StatusExited` shell. N crashed dev-server tabs = N subprocess spawns ~2×/s forever, regressing this PR's sibling fork-contention reduction (5→3). Cache the exit code on the dead→exited transition, or fold `pane_dead_status`/`pane_dead_signal` into `RefreshSessionCache`'s single `list-panes -a`.

internal/ui/drawer.go
```diff
@@ -0,0 +1,662 @@
...
+		if s.ID != sh.ID {
+			out = append(out, s)
+		}
+	}
+	h.shells = out
+	h.workerMu.Unlock()
+	if err := h.storage.DeleteShell(sh.ID); err != nil {
+		debuglog.Logger.Error("storage: DeleteShell", "id", sh.ID, "err", err)
+	}
+	if n := len(h.shellsForActiveRepo()); h.drawerActiveTab >= n {
```
hayke102, line 509: **closeShell doesn't re-point the live stream.** After removing the active shell here, `h.shellStreamTarget` still names the killed session, so renderDrawer's match at drawer.go:534 fails → the body falls to blank, while the reader keeps streaming the dying session — until the next `drawerTypeTick` (~120ms) repoints. `switchTab` calls `h.syncShellStream()`; this path (and `killShellsForRepo`) should too.

internal/ui/drawer.go
```diff
@@ -0,0 +1,662 @@
...
+		innerWidth = 1
+	}
+	// Stable terminal viewport: a real terminal has a fixed size, so the body is
+	// the available rows (clamped), not content-fit. Recorded for syncShellStream,
+	// which sizes the reader + emulator to match so wrap points line up.
+	dispRows := maxOuterH - 2
+	if dispRows > drawerMaxBodyRows {
+		dispRows = drawerMaxBodyRows
+	}
+	if h.drawerHeight > 0 && dispRows > h.drawerHeight {
```
hayke102, line 668: **drawer_height can't exceed 14.** `dispRows` is already clamped to `drawerMaxBodyRows` (14) at line 515, so this `dispRows > h.drawerHeight` check never fires for `drawer_height ≥ 15` — setting `drawer_height: 24` silently yields 14 rows. If the 14-row ceiling is intentional, document/cap it in the config getter; otherwise apply the config value before the hard cap.

internal/tmux/control_output.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+}
+
+func (r *OutputReader) writeCommand(cmd string) error {
+	_, err := fmt.Fprintf(r.ptmx, "%s\n", cmd)
+	return err
+}
+
+// SendKeys sends named keys to the reader's target pane over the same connection.
+func (r *OutputReader) SendKeys(keys ...string) error {
```
hayke102: **Dead code / double control client.** `SendKeys`/`SendLiteralKeys`/`writeCommand`/`Target` have no callers — the drawer types via a *separate* `ControlClient` (`getControlClient` → `forwardKeyToPane`, drawer.go:435,440), so every live shell holds **two** control-mode tmux processes. The type doc says it attaches “so it can send keys to the same pane over one connection,” but that path is unused. Either route keystrokes through `shellReader.SendKeys` and drop `getControlClient` for the drawer, or delete these four methods.

internal/config/config.go
```diff
@@ -32,6 +32,7 @@ type Config struct {
 	StatusIndicator              string `json:"status_indicator,omitempty"` // "icon" (default) or "bar"
 	Telemetry                    *bool  `json:"telemetry,omitempty"`
 	DefaultAgent                 string `json:"default_agent,omitempty"` // "claude" or "codex"
+	DrawerHeight                 int    `json:"drawer_height,omitempty"` // terminal-drawer body rows (default 12)
```
hayke102, line 35: This new `drawer_height` option isn't reachable from the **Settings dialog** (no row / `cycleValue` case in the Behavior category) and the CLAUDE.md `- Config file: …` option list wasn't updated to include it — the add-config-option pattern wants both. As-is it's only settable by hand-editing `config.json`.

hayke102 (commented):
Automated `/code-review` pass on the terminal-drawer feature. Inline threads below, roughly highest-severity first. The tmux cache-drop one (#1) is the only one I'd consider a blocker — it can affect agent-session status too, not just shells.

internal/tmux/tmux.go
hayke102: Fixed — dropped the whole-output TrimSpace so the last dead pane keeps its trailing-tab 4th field (empty pane_current_command); the line=="" guard still handles blank lines.

internal/ui/drawer.go
```diff
@@ -0,0 +1,662 @@
...
+
+// attachShell does a full PTY takeover of the shell's tmux session (ctrl+q
+// returns), reusing the session attach path.
+func (h *Home) attachShell(sh *shell.Shell) tea.Cmd {
+	if sh == nil {
+		return nil
+	}
+	if !sh.Tmux().Exists() {
+		h.setError(fmt.Errorf("shell exited — press r to restart"))
+		return nil
```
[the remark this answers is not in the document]
hayke102: Fixed — the hint now says "press Enter to restart" to match the actual key.

changelog/unreleased/terminal-drawer.md
```diff
@@ -0,0 +1,4 @@
+---
+type: added
+---
+New terminal drawer (press `` ` ``): live, real-time shells — dev servers, log tails, scratch commands — scoped to the selected repo/worktree. Rendered through a true terminal emulator, so streaming output, colors, and full-screen tools (vim, htop, lazygit) display faithfully. Type straight in; `Ctrl+T` new, `Ctrl+W` close, `Ctrl+PgUp/PgDn` switch tabs, `Ctrl+G` full-screen attach.
```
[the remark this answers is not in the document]
hayke102: Fixed — trimmed the fragment to ~2 sentences and dropped the implementation detail.

internal/session/shell_storage_test.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if err := db.UpdateShellName("sh1", "logs"); err != nil {
+		t.Fatalf("UpdateShellName: %v", err)
+	}
+	if err := db.UpdateShellTmuxName("sh1", "fleetsh_logs_ffff0000"); err != nil {
+		t.Fatalf("UpdateShellTmuxName: %v", err)
+	}
+	shells, _ := db.LoadShells()
+	if len(shells) != 1 || shells[0].Name != "logs" || shells[0].TmuxName != "fleetsh_logs_ffff0000" {
```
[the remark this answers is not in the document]
hayke102: Fixed — both LoadShells() calls now capture and assert the error.

internal/shell/shell_test.go
```diff
@@ -0,0 +1,35 @@
...
+		}
+	}
+}
+
+func TestShellPrefixNotSessionPrefix(t *testing.T) {
+	// "fleetsh_" must NOT be a prefix of an agent-session name check ("fleet_"),
+	// or shells would leak into tmux.ListSessions().
+	if len(ShellPrefix) < 6 || ShellPrefix[:6] != "fleets" {
+		t.Fatalf("unexpected ShellPrefix %q", ShellPrefix)
+	}
```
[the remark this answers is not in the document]
hayke102: Fixed — now asserts ShellPrefix == "fleetsh_" and that it does not start with "fleet_".

internal/shell/shell.go
```diff
@@ -0,0 +1,209 @@
...
+	st := DeriveStatus(dead, tmux.PaneCurrentCommand(name))
```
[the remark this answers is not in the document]
hayke102: Fixed — RefreshStatus now snapshots tmux + tmuxName under a single lock.

internal/shell/shell.go
hayke102: Fixed — RefreshStatus treats an absent session (!ts.Exists()) as Exited, so Enter→restart works after a tmux server restart instead of stranding a dead "idle" tab.

internal/ui/app.go
hayke102: Fixed — extracted quitSequence() and routed the command-palette Quit through it, so it runs the same teardown + finalizeAllPendingDeletes as Ctrl+C.

internal/tmux/control_output.go
hayke102: Fixed — the reader now flags itself failed when the scan loop ends while not closed (e.g. ErrTooLong on an oversized %output line), and syncShellStream re-attaches on the next backstop tick instead of freezing on the last frame.

internal/ui/app.go
hayke102: Fixed — the exit code is now captured once on the dead→exited transition and cached (Restart clears it), so PaneDeadInfo no longer spawns a subprocess every fast cycle for each exited shell.

internal/ui/drawer.go
hayke102: Fixed — closeShell and killShellsForRepo now call syncShellStream to re-point (or tear down) the live stream immediately instead of waiting for the next tick.

internal/tmux/control_output.go
hayke102: Fixed — deleted the unused SendKeys/SendLiteralKeys/writeCommand/Target methods on OutputReader and updated the doc comment. The drawer keeps typing via the shared ControlClient, consistent with focus mode.

internal/ui/drawer.go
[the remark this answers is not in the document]
hayke102: Fixed — the attach (NewOutputReader, which forks tmux -C + a PTY) now runs off the Update goroutine via an async dispatch (startShellStreamAsync → shellStreamReadyMsg), and the reader Close (Kill+Wait) runs off-thread too. The in-flight result is discarded if the target or body size changed mid-attach.

hayke102: Fixed — moved attach + teardown off Update: the fork is dispatched async and the reader Close runs in a goroutine. attachShell keeps a synchronous detach (teardownShellStreamSync) before the full-screen takeover, since tmux sizes the shared window to the smallest client.

internal/ui/drawer.go
hayke102: Fixed — GetDrawerHeight now clamps to [4,14] (the 14-row hard body cap is documented in the getter), so a larger value no longer silently yields 14; the new Settings row uses bounded presets.

internal/config/config.go
hayke102: Fixed — added a "Drawer height (rows)" row to the Behavior settings (re-synced on close so it applies without a relaunch) and documented drawer_height in CLAUDE.md.

internal/ui/app.go
```diff
@@ -1507,45 +1674,89 @@ func (h *Home) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
...
+			}
+		}
+		return h, nil
+	}
+	if h.drawerHasFocus() {
+		if sh := h.activeShell(); sh != nil {
+			if cc := h.getControlClient(); cc != nil {
+				_ = cc.SendLiteralKeys(sh.TmuxName(), msg.Content)
+			}
+		}
```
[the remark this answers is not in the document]
hayke102: Not changing this now. Forwarding keystrokes to a tmux control client directly from Update is the established, shipped pattern — focus mode (handleFocusKey) does exactly this on master. A single send-keys is a few-byte PTY write that does not block in practice (the control client also drains its side via io.Copy to Discard). A serialized tmux-input worker would be a new subsystem and, to be consistent, would have to cover focus mode too — out of scope for this PR. The genuinely heavy op here, the stream attach/teardown fork, has been moved off Update.

internal/ui/drawer.go
```diff
@@ -0,0 +1,662 @@
...
+		}
+		return h, nil
+	}
+
+	cc := h.getControlClient()
+	if cc == nil {
+		h.setError(fmt.Errorf("type-mode unavailable — Ctrl+G to attach"))
+		return h, nil
+	}
+	forwardKeyToPane(cc, sh.TmuxName(), msg)
```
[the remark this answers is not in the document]
hayke102: Same as the paste/input thread: drawer key forwarding mirrors the established focus-mode path (handleFocusKey on master), and a single send-keys is non-blocking in practice. Deferring a unified, serialized tmux-input worker (which should also cover focus mode) to its own change. The attach/teardown fork — the real stall risk — is now off the Update goroutine.

internal/git/repo_info.go
```diff
@@ -22,12 +22,16 @@ type RepoInfo struct {
...
+// check, and the origin lookup — down from five, to ease fork/exec-lock
+// contention on the per-cycle, all-repos fan-out.
 func RefreshGitInfo(repoPath string) *RepoInfo {
+	branch, isWorktree := revParseLayout(repoPath)
 	return &RepoInfo{
-		Branch:         GetBranchName(repoPath),
+		Branch:         branch,
 		IsDirty:        HasUncommittedChanges(repoPath),
-		IsWorktreeRepo: IsWorktree(repoPath),
+		IsWorktreeRepo: isWorktree,
```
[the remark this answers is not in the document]
hayke102: Leaving as-is. revParseLayout derives isWorktree deterministically (compares --git-dir vs --git-common-dir); rev-parse does not return partial output under normal conditions. The only false-flip path is a hard git error returning <2 lines for one ~2s cycle (e.g. mid worktree add/remove), which self-heals next cycle — a cosmetic regroup, not a correctness bug. The cited guideline governs the UI delete-scope path (repoIsWorktree + git.IsWorktree fallback), not RefreshGitInfo, which has no access to the cached RepoInfo without a signature change. Not worth that for a speculative, self-healing blip.

CLAUDE.md
```diff
@@ -82,7 +86,8 @@ chrome-extension/                # Chrome MV3 extension (service worker, manifes
...
...capes that x/vt would otherwise leak as visible text. **Always-typing, 2 states** (`drawerMode`: hidden/typing — see `internal/ui/drawer.go`): `` ` `` opens straight into TYPING (auto-creates a shell if the repo has none) — keystrokes forward to the shell pane via the focus-mode control client (`forwardKeyToPane`, which maps any `Ctrl+<letter>` → tmux `C-<letter>` so the shell's own line-editing keeps working; `Esc` passes through to the shell). **No menu mode**; chrome is a small set of Ctrl chords intercepted before forwarding: `Ctrl+T` new shell, `Ctrl+W` close (twice to confirm a running one, armed via `drawerCloseArmed`), `Ctrl+PgUp/PgDn` switch tab, `Ctrl+G` full attach (Ctrl+Q returns), `` ` `` close drawer. An **exited** shell restarts on `Enter` (no live process to type to). Cost: the shell loses `Ctrl+T` (transpose) and `Ctrl+W` (delete-word) to the drawer. Status (○ idle / ● running / ✕ exited+code) derives from tmux `pane_current_command` + pane-dead, no hooks. Shell tmux sessions use the `fleetsh_` prefix; removing a worktree kills its shells first (`killShellsForRepo`) so the dir frees for `git worktree remove`. Max body height via `drawer_height` config (default 12).
```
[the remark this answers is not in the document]
hayke102: Fixed — clarified line 82: fleet_ is the agent-session prefix, drawer shells use a distinct fleetsh_ prefix (intentionally not a prefix of fleet_, so shells never leak into tmux.ListSessions). They were always two separate prefixes, not a contradiction.

internal/ui/app.go
```diff
@@ -2481,9 +2732,52 @@ func (h *Home) worktreeDeleteWarnings(repoPath string) []string {
...
+			kept = append(kept, sh)
+		}
+	}
+	h.shells = kept
+	h.workerMu.Unlock()
+	for _, sh := range doomed {
+		if err := h.storage.DeleteShell(sh.ID); err != nil {
+			debuglog.Logger.Error("storage: DeleteShell", "id", sh.ID, "err", err)
+		}
+		_ = sh.Kill()
```
hayke102: **Blocking tmux I/O on the Update goroutine.** `killShellsForRepo` runs `sh.Kill()` (a synchronous `tmux kill-session` subprocess) once per doomed shell, and it's called from `deferDeleteRepo` — i.e. directly on the Bubble Tea Update thread. Deleting a worktree/origin group that has drawer shells freezes the whole TUI while it spawns a kill-session per shell; with several shells or a slow tmux server the loop stalls visibly.

This is exactly the Update()-thread I/O the project rule pushes into background workers — note the `git worktree remove` right below is already async. Suggest moving the kills (and `DeleteShell`) into a `tea.Cmd` closure and keeping only the in-memory slice update + `syncShellStream` on the Update thread.

internal/ui/drawer.go
```diff
@@ -0,0 +1,796 @@
...
+	}
+}
+
+// attachShell does a full PTY takeover of the shell's tmux session (ctrl+q
+// returns), reusing the session attach path.
+func (h *Home) attachShell(sh *shell.Shell) tea.Cmd {
+	if sh == nil {
+		return nil
+	}
+	if !sh.Tmux().Exists() {
```
hayke102: **Ctrl+G attaches to a dead pane instead of restarting.** `Session.Start` sets `remain-on-exit=on`, so an exited shell keeps a `pane_dead=1` pane that's still enumerated in the session cache — meaning `sh.Tmux().Exists()` returns `true` for an exited shell. So this guard passes, and `tea.Exec` does a full-screen attach into a dead `[dead]` pane; the user has to Ctrl+Q back out.

The Enter→restart path handles this correctly, but Ctrl+G routes to `attachShell` before that branch. Suggest gating on the shell's exited *status* here (e.g. `sh.Status() == shell.StatusExited`) rather than pane existence, and routing to the restart path when exited.

internal/ui/drawer.go
```diff
@@ -0,0 +1,796 @@
...
+	first := h.drawerMode == drawerHidden
+	if first {
+		h.drawerRepo = h.resolveCurrentRepo()
+		h.drawerActiveTab = 0
+		h.actionLog.Add("open terminal drawer", h.drawerRepo, true)
+	}
+	h.drawerMode = drawerTyping
+	h.drawerCloseArmed = false
+	h.drawerTarget = 1
+	cmds := []tea.Cmd{h.drawerAnimTick(), h.drawerTypeTick()}
```
hayke102: **`drawerTypeTick` loops leak on fast open/close toggling.** This unconditionally starts a new self-rescheduling tick loop every time the drawer opens. The loop only terminates if it happens to fire in a frame where `drawerMode != drawerTyping` (the guard in `app.go`'s `drawerTypeTickMsg` case). Toggle open→close→open faster than the 120ms `drawerSyncInterval` and the first loop is still pending when the reopen starts a second one; when the first then fires while the drawer is open again, it reschedules itself instead of dying. Every fast toggle adds a permanent loop, each calling `syncShellStream` + a render wake every 120ms — background CPU climbs and never subsides until restart.

Suggest a dedup guard (e.g. only start the loop when `first`, or an `h.drawerTickRunning` flag cleared when the guard stops it).

internal/ui/drawer.go
```diff
@@ -0,0 +1,796 @@
...
+			if pad < 0 {
+				pad = 0
+			}
+			body[cursorY] = line + strings.Repeat(" ", pad) + cur.Render(" ")
+		} else {
+			cell := ansi.Strip(ansi.Cut(line, cursorX, cursorX+1))
+			if cell == "" {
+				cell = " "
+			}
+			body[cursorY] = ansi.Truncate(line, cursorX, "") + cur.Render(cell) + ansi.TruncateLeft(line, cursorX+1, "")
```
hayke102: **Block-cursor overlay garbles a wide glyph under the cursor.** `ansi.Cut(line, cursorX, cursorX+1)` / `ansi.Truncate(line, cursorX)` / `ansi.TruncateLeft(line, cursorX+1)` all treat cursor position as a width-1 cell offset. When the shell cursor rests on a double-width glyph (CJK, emoji, wide box-drawing — common in `less`/`vim`/non-ASCII prompts inside a drawer shell), the cut splits the glyph: half renders under the cursor and the remainder resumes mid-glyph, producing a garbled/misaligned cursor row. Consider advancing by the rune's display width (2 for wide cells) when slicing.

internal/session/storage.go
```diff
@@ -313,6 +348,70 @@ func (s *StateDB) DeleteSession(id string) error {
...
+}
+
+// UpdateShellTmuxName updates a shell's tmux session name (after a restart).
+func (s *StateDB) UpdateShellTmuxName(id, tmuxName string) error {
+	_, err := s.db.Exec("UPDATE shells SET tmux_name = ? WHERE id = ?", tmuxName, id)
+	return err
+}
+
+// UpdateShellName updates a shell's display name (rename).
+func (s *StateDB) UpdateShellName(id, name string) error {
```
hayke102: **Dead code.** `UpdateShellName` (and its dedicated test at `shell_storage_test.go:60`) has no caller — there's no shell-rename feature in the UI (no R-key path routes to a shell). Speculative persistence API a maintainer has to keep in sync. Suggest dropping it until a rename feature actually needs it.

internal/ui/drawer.go
```diff
@@ -0,0 +1,796 @@
...
+	return "", false
+}
+
+// --- rendering ---
+
+// renderDrawer draws the drawer as a bordered panel at its current slide
+// height. maxOuterH caps the panel's total rows (border included) so it leaves
+// room for whatever shares its space — the preview, in the dual split.
+func (h *Home) renderDrawer(width, maxOuterH int) string {
+	shells := h.shellsForActiveRepo()
```
hayke102, line 655: **Minor: redundant per-frame shell scans.** `renderDrawer` and its helpers call `shellsForActiveRepo` several times per render pass (`renderDrawer`, the body switch, `drawerTitle`, `drawerModeLabel`→`activeShell`), each re-filtering the full `h.shells` slice. Since a render is scheduled per output byte, this repeats O(N) filtering work every frame. Computing the active-repo slice once at the top of `renderDrawer` and threading it into the helpers would avoid it. Low priority.

hayke102 (commented):
Automated review of the terminal-drawer branch (high effort, independently verified). 6 findings below as resolvable threads — 3 correctness regressions, 1 rendering bug, 2 cleanups. Not blocking; flagging the top 3 as worth addressing before merge.

internal/ui/app.go
hayke102: Fixed — killShellsForRepo now returns a tea.Cmd; the in-memory removal + syncShellStream stay on Update, but the per-shell kill + DeleteShell run off-thread. deferDeleteRepo tea.Sequence()s the kill command ahead of destroyWorktree so the dir is still freed before git worktree remove (nil kill cmd is dropped by Sequence).

internal/ui/drawer.go
hayke102: Fixed — attachShell now gates on sh.Status() == StatusExited (not pane existence, which stays true under remain-on-exit) and routes to restartShell, same as Enter. Also covers the absent-session case.

internal/ui/drawer.go
hayke102: Fixed — added an h.drawerTickRunning guard: openDrawerTyping starts the type-tick loop only when one is not already live, and the drawerTypeTickMsg handler clears the flag when it stops (drawerMode != typing). At most one loop, so fast toggling no longer accumulates them.

internal/ui/drawer.go
hayke102: Fixed — the cursor overlay now detects a double-width grapheme (a 1-cell cut returns "" while the 2-cell cut has width 2) and slices by the full cell width, so the cursor block covers the whole glyph and the right-hand remainder stays aligned.

internal/session/storage.go
hayke102: Fixed — dropped UpdateShellName (no caller) and its assertion from the storage test; kept UpdateShellTmuxName + DeleteShell coverage.

internal/ui/drawer.go
hayke102: Fixed — renderDrawer computes the active-repo shell slice once and threads it into drawerTitle/drawerModeLabel (via a new activeShellIn helper), so a render pass no longer re-filters h.shells several times.

PR #162, opened 2026-07-02
fix(ui): stop Settings › Behavior from wrapping labels and cutting off rows

hayke102:
## Problem

In **Settings → Behavior**, the last rows (Telemetry, Default agent) were cut off and couldn't be reached, and long labels wrapped across multiple lines, wasting vertical space.

before

## Root cause

Not a missing-scroll issue — a **wrap → truncation** bug:

- `renderDetail` rendered every label in a fixed `Width(13)` column. Behavior's long labels ("Confirm before restart", "Drawer height (rows)", "Origin forget removes worktrees") don't fit 13 columns, so lipgloss wrapped each to 2–3 lines.
- The detail block is clamped to `blockH = maxBlockHeight()`, which counts **rows** (14), not rendered **lines** (~16 after wrapping). `ensureExactHeight` then truncated the overflow, dropping the last two rows off the bottom.

So both symptoms — inefficient screen use and "can't see below" — came from the one narrow label column.

## Fix

- **`labelColWidth(rows)`** — size the label column to the widest label *per category* so labels never wrap. Behavior's 11 rows now render as 11 lines, comfortably within the block; nothing is truncated.
- **`detailContentWidth()`** — size the detail column to the widest row (label + arrows + value) across categories instead of a hardcoded `36`, so the content fits and horizontal space is used. Still flexes down on narrow terminals.

No scrolling viewport added — once wrapping is fixed everything fits at normal terminal sizes.

## Testing

- `make build` passes.
- Verified line-count math: Behavior renders 11 single-line rows within `blockH=14`; Appearance labels (≤13 chars) are unaffected.

### DIFF
--- a/internal/ui/settings.go
+++ b/internal/ui/settings.go
@@ -39,6 +39,7 @@ type settingRow struct {
 	label     string
 	subheader bool
 	value     func(c *config.Config) string    // display value (interactive rows only)
+	valueW    func() int                       // widest *possible* display width of value; stable across cfg (nil for subheaders)
 	cycle     func(d *SettingsDialog, dir int) // mutate cfg + live-apply (interactive rows only)
 }
 
@@ -60,6 +61,7 @@ type SettingsDialog struct {
 	cfg            *config.Config
 	origTheme      string
 	categories     []settingsCategory
+	detailWByCat   []int // natural detail-column width per category (cached; depends only on static labels/presets)
 	categoryCursor int
 	rowCursor      int
 	focus          settingsFocus
@@ -67,7 +69,15 @@ type SettingsDialog struct {
 
 // NewSettingsDialog creates a settings dialog.
 func NewSettingsDialog(cfg *config.Config) *SettingsDialog {
-	return &SettingsDialog{cfg: cfg, categories: buildSettingsCategories()}
+	d := &SettingsDialog{cfg: cfg, categories: buildSettingsCategories()}
+	// The natural detail width per category depends only on static label text and
+	// the fixed preset sets (via each row's widest *possible* value), so compute
+	// it once here rather than on every View() frame.
+	d.detailWByCat = make([]int, len(d.categories))
+	for i, c := range d.categories {
+		d.detailWByCat[i] = categoryContentWidth(c)
+	}
+	return d
 }
 
 func (d *SettingsDialog) Show() {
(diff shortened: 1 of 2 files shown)


internal/ui/settings.go
```diff
@@ -309,7 +352,7 @@ func (d *SettingsDialog) renderDetail() string {
 		}
 
 		selected := d.focus == focusDetail && i == d.rowCursor
-		labelStyle := lipgloss.NewStyle().Width(13).Align(lipgloss.Left)
+		labelStyle := lipgloss.NewStyle().Width(labelW).Align(lipgloss.Left)
```
hayke102, line 408: **Values get clipped on narrow/medium terminals — a horizontal-truncation regression.**

`labelW` here is the widest label in the *current* category (31 cols for Behavior's "Origin forget removes worktrees"), but it isn't reduced when `detailW` clamps down to fit a narrow pane (`View`, L249-251). So on a ~55–62 col terminal `detailW` clamps to ~39 while the label still eats 31 + `"  ◂ "` (4) = 35 cols, leaving ~4 for the value — and `padBlock`'s `ensureExactWidth` truncates the right side: `Default agent` renders `◂ Clau` instead of `Claude`, Telemetry's `On`/`Off` gets cut.

Before this commit (`Width(13)`, `detailW=36`) those short-label rows showed their values in full at the same width — so the vertical-truncation fix trades in a horizontal one for short-label rows on narrow panes. Consider clamping `labelW` once `detailW` is clamped so a minimum value width is preserved (e.g. `labelW = min(labelW, detailW - rowChrome - minVal)`).

internal/ui/settings.go
```diff
@@ -199,11 +241,11 @@ func (d *SettingsDialog) View() string {
...
-	// The detail column flexes with the terminal so the modal fits narrow panes
-	// (the old dialog clamped its width the same way); the floor keeps values
-	// readable. detailW + railW + ruleW never exceeds the content area, so the
-	// bordered box stays within d.width.
-	detailW := 36
+	// Size the detail column to the widest row across all categories so labels
+	// render on one line (no wrap → no truncation of the rows below). It still
+	// flexes down to fit narrow panes; the floor keeps values readable. detailW +
+	// railW + ruleW never exceeds the content area, so the box stays within d.width.
+	detailW := d.detailContentWidth()
```
hayke102: **`detailW` grew from a fixed 36 to ~45, so the Appearance live preview now disappears at medium widths.**

The widest row is Behavior's 31-col label + 6 chrome + widest value ≈ 45. The preview gate at L268 (`avail = width - 6 - 14 - 3 - detailW - 3 >= 24`) therefore moves from roughly `width ≥ 86` to `width ≥ ~95`. On an ~86–94 col terminal, opening Settings → Appearance no longer shows the mock-sidebar live preview it rendered before this change.

internal/ui/settings.go
```diff
@@ -100,6 +100,48 @@ func (d *SettingsDialog) maxBlockHeight() int {
...
+	const rowChrome = 6
+	need := 0
+	for _, c := range d.categories {
+		labelW := labelColWidth(c.rows)
+		maxVal := 0
+		for _, r := range c.rows {
+			if r.subheader || r.value == nil {
+				continue
+			}
+			if vw := lipgloss.Width(r.value(d.cfg)); vw > maxVal {
```
hayke102: **`detailContentWidth()` measures *current* config values, so the modal resizes and re-centers when you cycle a value.**

Because `detailW` now depends on live state, pressing → to change Default agent Claude (6) → OpenCode (8), or cycling Editor to/from a long `$EDITOR` path, makes the whole box grow/shrink and re-center on that keypress instead of staying put. Sizing to the widest *possible* value (or a stable constant) keeps the constant-width invariant the old `detailW=36` had.

Minor, same spot: this walks every category and invokes every row's `value(d.cfg)` closure on every `View()` frame — including Appearance's rows while you're on Behavior — purely to measure width. Widths only change on config change, so caching the computed width (invalidated on cycle) would avoid the per-frame `fmt.Sprintf`/getter work.

internal/ui/settings.go
hayke102: Fixed — renderDetail now clamps labelW when detailW is clamped for a narrow pane (min 4 cols reserved for the value) and truncates the over-long label with … instead of letting it wrap (wrapping would re-inflate the block height and truncate the rows below). Values stay readable now.

internal/ui/settings.go
hayke102: Fixed — width is now computed from each row's widest *possible* value (valueW, walking the preset sets) rather than the live value, and cached per-category in detailWByCat at construction. So the modal no longer resizes/re-centers when cycling a value, and the per-frame value(d.cfg) closure calls are gone.

internal/ui/settings.go
hayke102: Fixed — detailW is now sized per current category (Appearance ~35, Behavior ~45) instead of the global max, so Appearance no longer inherits Behavior's 31-col label. The preview gate returns to roughly its old threshold and the mock-sidebar shows again on ~86-94 col terminals. Kept all labels verbatim (chose per-category sizing over shortening the labels).

PR #163, opened 2026-07-02
feat(telemetry): three-way telemetry (full/minimal/off) + daily-active heartbeat

hayke102:
## What & why

Replace the on/off telemetry consent with a **three-way mode** so I can count daily active users without collecting anyone's name — respecting people who don't want their identity in telemetry while still getting DAU.

| Mode | What's sent |
|------|-------------|
| **full** | Usage events tagged with git name/email (unchanged from today) |
| **minimal** | Anonymous only: `app_started` (version + mode) + a per-day `app_active` heartbeat, keyed on the device hash. No git name/email, no people profile, every other event no-ops. |
| **off** | Nothing. Env `FLEET_TELEMETRY_DISABLED` / `DO_NOT_TRACK` force this regardless of mode. |

## Onboarding & migration

- **First-run consent:** accept → `full`, decline → `minimal`. The dialog now honestly discloses that declining still sends an **anonymous daily-active ping** (no name, no email, no identity), and points to Settings for turning telemetry **off** entirely.
- **`off` is Settings-only** — a 3-way cycler in the Behavior category (was an on/off toggle).
- **Legacy migration:** the old `telemetry` bool maps true → `full`, false → `minimal` (previously-declined users start sending anonymous DAU).

## Daily-active heartbeat

`app_started` only fires at launch, but fleet is often left open for days — so a once-a-day user of a long-running instance would be undercounted. New `analytics.Heartbeat()` fires on the UI tick, **rate-limited to once per calendar day in-memory**, so a midnight rollover re-emits `app_active`. Runs in both full and minimal. Drops the now-obsolete `telemetry_declined` beacon.

## Scope note

Minimal is intentionally the **leanest** thing that counts DAU (just `app_started` + heartbeat), not a full anonymized usage stream. Easy to widen later if we want anonymous usage trends from minimal users too.

## Verification

- `make build` ✓
- `make test` (full suite) ✓ — no failures; `-race` clean on `analytics` + `config`
- New/updated tests: `GetTelemetryMode` migration cases, `TelemetryConfigured`, env opt-out
- Lint clean on changed code

## Summary by CodeRabbit

* **New Features**
  * Added three telemetry modes: **Full**, **Minimal**, and **Off**.
  * First-run consent now maps to **Full** or **Minimal**, and settings can switch telemetry fully off.
  * Added a daily active heartbeat event to record app activity more consistently.

* **Bug Fixes**
  * Existing telemetry preferences now migrate cleanly from the older on/off setting.
  * Settings and startup behavior now honor the selected telemetry mode immediately.


internal/config/config.go
```diff
@@ -413,12 +429,31 @@ func (c *Config) GetSidebarDensity() string {
...
+func (c *Config) GetTelemetryMode() string {
+	switch c.TelemetryMode {
+	case TelemetryFull, TelemetryMinimal, TelemetryOff:
+		return c.TelemetryMode
+	}
+	if c.Telemetry != nil {
+		if *c.Telemetry {
+			return TelemetryFull
+		}
+		return TelemetryMinimal
```
hayke102, line 463: Worth an explicit sign-off: legacy `Telemetry=false` (a user who clicked "No thanks" on an older build) migrates to `minimal`, which now transmits. Before this change those users sent nothing on subsequent launches (`Init(false)` → disabled client); after upgrade `TelemetryConfigured()` is true, so startup runs `Init("minimal")` and emits `app_started` + a per-day `app_active` (device-hash keyed) on every launch, with no re-prompt.

The comment here and the docs say this is intentional ("a previously-declined user keeps anonymous daily-active tracking"), so this is a confirm-intent flag, not a defect — but a previously opted-*out* cohort silently beginning transmission is the kind of decision I'd want deliberate. If preserving their choice was the intent, `false` should migrate to `off` instead.

internal/analytics/analytics.go
```diff
@@ -193,22 +225,60 @@ func (c *Client) worker() {
...
+	day := time.Now().Format("2006-01-02")
+	c.mu.Lock()
+	if c.lastActiveDay == day {
+		c.mu.Unlock()
+		return
+	}
+	c.lastActiveDay = day
+	c.mu.Unlock()
+
+	mode := ModeFull
```
hayke102: Minor: the mode is now represented three ways. `Init(mode string)` derives `minimal := mode == ModeMinimal` and discards the string; here `Heartbeat` reconstructs it (`ModeFull` / `ModeMinimal`), and `TrackAppStarted` separately hardcodes the same literals into its payload. Storing `c.mode` once at `Init` would drop the reconstruction and keep the emitted `"mode"` property consistent if a third+ mode ever needs to appear in events. Not blocking.

internal/config/config.go
```diff
@@ -413,12 +429,31 @@ func (c *Config) GetSidebarDensity() string {
...
 	}
-	return *c.Telemetry
+	return TelemetryFull
+}
+
+// TelemetryConfigured reports whether the user already has an explicit
+// telemetry preference on disk (the new mode field or the legacy bool). Used
+// to skip the first-launch consent prompt for upgrading users.
+func (c *Config) TelemetryConfigured() bool {
+	return c.TelemetryMode != "" || c.Telemetry != nil
```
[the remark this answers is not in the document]
hayke102: Fixed in ad6d474 — added isValidTelemetryMode; an unrecognized telemetry_mode no longer counts as configured (consent prompt still fires) and GetTelemetryMode falls back to minimal, never full.

internal/config/config_test.go
```diff
@@ -142,29 +142,44 @@ func TestIsCopyClaudeSettingsEnabled(t *testing.T) {
...
+	cases := []struct {
+		name string
+		cfg  *Config
+		want string
+	}{
+		{"unset defaults to full", &Config{}, TelemetryFull},
+		{"explicit full", &Config{TelemetryMode: TelemetryFull}, TelemetryFull},
+		{"explicit minimal", &Config{TelemetryMode: TelemetryMinimal}, TelemetryMinimal},
+		{"explicit off", &Config{TelemetryMode: TelemetryOff}, TelemetryOff},
+		{"unrecognized mode falls through to default", &Config{TelemetryMode: "bogus"}, TelemetryFull},
```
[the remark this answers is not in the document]
hayke102: Fixed in ad6d474 — the invalid-mode case now expects minimal (conservative), and TestTelemetryConfigured asserts a bogus mode reports not-configured while a bogus mode + legacy bool still counts.

internal/analytics/analytics.go
hayke102: Fixed in ad6d474 — the resolved mode is now stored once on Client (dropped the separate minimal bool); Heartbeat and TrackAppStarted read c.mode for the emitted "mode" property instead of reconstructing/hardcoding it.

internal/config/config.go
hayke102: Intentional — previously-declined users migrate to an anonymous, no-identity daily-active ping (device-hash keyed, no name/email), documented in GetTelemetryMode + CLAUDE.md and disclosed in the consent copy. Preserving the exact opt-out cohort would mean false -> off, but the deliberate product call is anonymous DAU, so keeping false -> minimal.

PR #164, opened 2026-07-02
feat(ui): shutdown indicator on quit

hayke102:
## Problem

Pressing **Ctrl+C** to quit fleet looked like a freeze: the UI sat on its last frame for ~2s, then the process vanished — no signal that anything was happening.

Root cause: `quitSequence()` ran the **entire** teardown (worker cancel, hook watcher, tmux control-client close, drawer stream teardown, pending-delete finalize, telemetry drain) **synchronously inside the Bubble Tea `Update()` loop**, and only then returned `tea.Quit`. Frames render *between* `Update` calls, so nothing ever painted during teardown. The dominant blocker was `analytics.Shutdown()` draining queued telemetry over the network (up to **2s**).

## Change

Split the quit into two phases so a frame renders — and a spinner animates — while the blocking teardown runs off the `Update` loop:

- **`beginQuit`** flips into a **"Shutting down…"** overlay: it captures the current frame, cancels the worker, and hands the teardown to a command goroutine. `View()` dims the captured frame and floats the box, reusing the existing command-palette overlay pipeline (`dimBackdrop` + `overlayAt`, `DialogStyle`).
- **`performShutdown`** holds the former `quitSequence` body, now running off-loop, ending with `tea.Quit`.
- An 80ms self-rescheduling spinner tick (mirroring the boot splash's `splashTick`) animates the box.
- While `quitting`, `Update` swallows every message except the spinner tick, so the teardown goroutine never races the render side (render touches only `frozenFrame`/`shutdownFrame`; teardown touches a disjoint set).
- Also shortened the analytics shutdown drain **2s → 500ms** so exit is prompt.

Both quit entry points (Ctrl+C and the command-palette Quit) route through `beginQuit`.

## Testing

- `make build` ✅
- `make test` (`-race`) ✅ — confirms the async-teardown-vs-render split is race-free
- Manual: Ctrl+C (and Ctrl+K → Quit) now dims the UI and shows an animated "Shutting down…" box, then exits noticeably faster.

https://claude.ai/code/session_01BTeU6hdAbqjBh9f5m5RY8A

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -528,11 +528,19 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	}
 	// While shutting down, the blocking teardown runs off-loop (see beginQuit);
 	// keep the spinner animating and swallow everything else so no key or worker
-	// message mutates state the teardown goroutine is touching concurrently.
+	// message mutates state the teardown goroutine is touching concurrently. A
+	// second Ctrl+C is the escape hatch: if teardown wedges (a stuck worktree
+	// remove, a wedged telemetry drain), force an immediate exit rather than
+	// spinning forever with no way out.
 	if h.quitting {
-		if _, ok := msg.(shutdownFrameMsg); ok {
+		switch m := msg.(type) {
+		case shutdownFrameMsg:
 			h.shutdownFrame++
 			return h, h.shutdownTick()
+		case tea.KeyPressMsg:
+			if m.String() == "ctrl+c" {
+				return h, tea.Quit
+			}
 		}
 		return h, nil
 	}
@@ -2199,8 +2207,19 @@ func (h *Home) beginQuit(source string) tea.Cmd {
 	}
 	debuglog.Logger.Info("quit requested", "source", source)
 	h.quitting = true
-	h.frozenFrame = h.renderBody() // capture the exact last frame to dim behind the overlay
-	h.cancel()                     // stop the worker now so it stops feeding Update
+	// Capture the frame to dim behind the overlay, mirroring View()'s guard
+	// order so a quit during boot freezes what the user is actually looking at
+	// (splash) rather than a half-built body — and never runs renderBody with a
+	// negative inner width in the pre-first-WindowSizeMsg (width == 0) window.
+	switch {
+	case h.width == 0:
+		h.frozenFrame = "" // nothing meaningful to freeze yet
+	case !h.booted:
+		h.frozenFrame = RenderSplash(h.width, h.height, h.bootProgress(), h.splashFrame)
+	default:
+		h.frozenFrame = h.renderBody()
+	}
+	h.cancel() // stop the worker now so it stops feeding Update
 	return tea.Batch(h.shutdownTick(), h.performShutdown())
 }
 
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -157,6 +157,9 @@ type (
 	// splashFrameMsg advances the boot-splash spinner. Self-rescheduled
 	// while !booted; not emitted after bootstrapDoneMsg.
 	splashFrameMsg time.Time
+	// shutdownFrameMsg advances the shutdown-overlay spinner. Self-rescheduled
+	// while `quitting`, until the teardown command emits tea.Quit.
+	shutdownFrameMsg time.Time
 	// discoveryMsg carries the launchpad's scan of ~/.claude/projects back to
 	// Update. Fired once, only on an empty fleet.
 	discoveryMsg struct {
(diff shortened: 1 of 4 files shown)


internal/ui/app.go
```diff
@@ -2144,72 +2174,98 @@ func (h *Home) handleKey(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
...
+// overlay (freezing the current frame behind it) and hands the blocking teardown
+// off to performShutdown, so a frame renders and the spinner animates instead of
+// the UI appearing frozen. Shared by Ctrl+C and the command-palette Quit command.
+func (h *Home) beginQuit(source string) tea.Cmd {
+	if h.quitting {
+		return nil // already tearing down; ignore repeat presses
+	}
+	debuglog.Logger.Info("quit requested", "source", source)
+	h.quitting = true
+	h.frozenFrame = h.renderBody() // capture the exact last frame to dim behind the overlay
```
hayke102: `beginQuit` captures `frozenFrame` via `renderBody()` unconditionally, but `View()` guards both `h.width == 0` and `!h.booted` before it ever calls `renderBody`.

- Quitting **during bootstrap** (`booted == false` — can be seconds with many repos) captures the empty/half-built body instead of the splash the user is actually looking at, so the overlay dims the wrong frame.
- In the narrower **pre-first-`WindowSizeMsg`** window (`width == 0`), `renderBody` runs with `innerW = -2` — exactly the state the `width == 0` early-return in `View()` exists to avoid.

Consider guarding the capture the same way `View()` does (fall back to splash/empty when `!booted` or `width == 0`).

internal/ui/app.go
```diff
@@ -515,6 +526,16 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		tok := perfwatch.MarkUpdateStart(fmt.Sprintf("%T", msg))
 		defer perfwatch.MarkUpdateEnd(tok)
 	}
+	// While shutting down, the blocking teardown runs off-loop (see beginQuit);
+	// keep the spinner animating and swallow everything else so no key or worker
+	// message mutates state the teardown goroutine is touching concurrently.
+	if h.quitting {
```
hayke102, line 535: Once `quitting` is set, the only path to `tea.Quit` is the end of `performShutdown`, and this guard swallows every key — including a repeat Ctrl+C.

If teardown blocks (e.g. `finalizeAllPendingDeletes` → `destroyWorktree` stuck on `git worktree remove` "Directory not empty", or the telemetry drain wedging near its 500ms cap on a bad network), `performShutdown` never reaches `tea.Quit()`, the spinner keeps animating (looks alive), and there's no way to force-quit.

Worth a hard deadline that forces `tea.Quit()` regardless, or letting a second Ctrl+C hard-exit.

internal/ui/app.go
```diff
@@ -2144,72 +2174,98 @@ func (h *Home) handleKey(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
...
-	if h.controlClient != nil {
-		h.controlClient.Close()
-	}
-	h.teardownShellStream()
-	// Finalize all pending deletes before quitting.
-	h.finalizeAllPendingDeletes()
+// shutdownTick schedules the next shutdown-overlay spinner advance (~80ms,
+// matching splashTick). Self-rescheduled by Update while `quitting`, until the
+// teardown command emits tea.Quit and the program exits.
+func (h *Home) shutdownTick() tea.Cmd {
```
hayke102, line 2194: nit: `shutdownTick` + `shutdownFrame` + `shutdownFrameMsg` duplicate the existing `splashTick`/`splashFrame`/`splashFrameMsg` spinner stack (same 80ms cadence, same `splashSpinnerFrames`), and `renderShutdownBox` re-writes the spinner-glyph render already in `RenderSplash`.

Since the boot and quit phases are mutually exclusive, the splash spinner state/tick could be reused (or a small `renderSpinnerGlyph(frame, color)` helper extracted) to drop a field, a msg type, and a tick func.

internal/ui/app.go
hayke102: Fixed — the capture in beginQuit now mirrors View()'s guard order: empty string when width == 0, RenderSplash(...) when !booted, renderBody() otherwise. So a quit during boot freezes the splash the user is actually seeing, and renderBody never runs with a negative inner width.

internal/ui/app.go
hayke102: Fixed — a second Ctrl+C during shutdown now hard-exits. The quitting guard intercepts a repeat ctrl+c KeyPressMsg and returns tea.Quit immediately, abandoning the in-flight teardown goroutine, so a wedged worktree-remove or telemetry drain can no longer trap the user on the spinner.

internal/ui/app.go
hayke102: Fixed — extracted renderSpinnerGlyph(frame, color) in splash.go, now used by both RenderSplash and renderShutdownBox, so the spinner-glyph render lives in one place. Kept the separate shutdownFrame/shutdownFrameMsg/shutdownTick stack decoupled from boot (mirrors the existing per-context spinner-tick precedent rather than overloading splashFrame across two phases).

PR #165, opened 2026-07-05
fix: show sessions as running during /compact

hayke102:
## What

Two related session-status changes, plus debug-status skill docs.

### 1. `/compact` shows **running** (the fix)
Sessions previously read **idle/finished** for the entire multi-minute compaction that `/compact` (and auto-compaction) triggers — a phase that fires **no** status hook, and whose pane reads `finished` to `detectStatus` (the queued-message/prompt `❯`s). This was reproduced from a `D` snapshot: hook, pane, transcript recency, *and* Claude's own status flag were all blind to compaction at once.

Fix: subscribe to Claude Code's **`PreCompact`** hook and map it → `running`. It brackets cleanly:
- `PreCompact` → running (covers manual `/compact` **and** auto-compaction)
- compaction completes → `SessionStart` (source=compact) → finished
- any queued prompt then flips back to running via `UserPromptSubmit`

Running survives the whole phase because the animated `(2m 27s)` timer + progress bar repaint every second, so the "stable >10s → finished" downgrade never fires. Existing installs pick up the new hook on next launch (a missing subscribed event re-triggers injection).

### 2. Snapshots capture Claude's session status file (diagnostics)
The `D` snapshot now freezes `~/.claude/sessions/<pid>.json` (Claude Code's own `busy`/`idle`/`shell` flag), matched by `sessionId` with the **live** process preferred over a stale dead one — as `claude_session_status.json` plus a `claude_status` block in `snapshot.json` (`status`, `pid_alive`, `status_age`, …).

### 3. Docs
`debug-status` `SKILL.md` + `reference.md` document the status file as a fifth signal (Running/Idle corroborator; no `waiting` value; blind to `/compact`; trust the value while `pid_alive`, not `status_age`).

## Tests
- `TestMapEventToStatus` — full mapping table incl. `PreCompact` → running (new `cmd/fleet` test)
- `buildHooksMapWithMarker` updated so `PreCompact` is a subscribed event (install-wiring)
- Golden fixture `pane_finished_compacting.txt` — real mid-`/compact` pane classifies `finished` (documents why the hook is needed)
- `snapshot_status_test.go` — status-file picker (live-beats-stale-dead, newest-among-dead tiebreak, no-match) + block builder; verified end-to-end against live `~/.claude/sessions` during development

`go test -race ./...` green · gofmt/vet/golangci-lint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sessions now remain **running** throughout Claude conversation compaction, including the compaction-close transition.
  * Improved status detection by freezing and including Claude Code’s per-process busy/idle/shell status, with clearer handling of live vs stale processes.

* **Documentation**
  * Updated debug-status and snapshot guidance to document the new `claude_status` block (backed by the frozen Claude session status file) and how to interpret `pid_alive` and missing/ambiguous matches.

### DIFF
--- a/cmd/fleet/hook_handler.go
+++ b/cmd/fleet/hook_handler.go
@@ -33,6 +33,12 @@ func mapEventToStatus(event string) string {
 		return "running"
 	case "Stop":
 		return "finished"
+	case "PreCompact":
+		// /compact (and auto-compaction) is a multi-minute busy phase that fires no
+		// other hook. Without this, the session reads finished/idle for the whole
+		// compaction (SessionStart→finished on completion closes the bracket, and
+		// any queued prompt then flips it back to running via UserPromptSubmit).
+		return "running"
 	case "PermissionRequest":
 		return "waiting"
 	case "Notification":
--- a/cmd/fleet/hook_handler.go
+++ b/cmd/fleet/hook_handler.go
@@ -36,8 +36,9 @@ func mapEventToStatus(event string) string {
 	case "PreCompact":
 		// /compact (and auto-compaction) is a multi-minute busy phase that fires no
 		// other hook. Without this, the session reads finished/idle for the whole
-		// compaction (SessionStart→finished on completion closes the bracket, and
-		// any queued prompt then flips it back to running via UserPromptSubmit).
+		// compaction. The closing SessionStart(source="compact") is skipped in
+		// handleHookHandler (not force-finished), so pane + stability detection —
+		// or a queued prompt's UserPromptSubmit — settle the end of compaction.
 		return "running"
 	case "PermissionRequest":
 		return "waiting"
@@ -64,6 +65,16 @@ func mapEventToStatus(event string) string {
 	default:
 		return ""
 	}
+}
+
+// isCompactSessionStart reports the SessionStart that Claude Code fires when a
+// compaction completes. Its status must NOT be forced to "finished": on
+// auto-compaction the turn is still running, so finishing here would flash a
+// spurious "finished" mid-turn. Skipping it lets the prior "running" (from the
+// PreCompact hook) stand; pane + stability detection settle a manual /compact
+// that ends at an idle prompt.
+func isCompactSessionStart(event, source string) bool {
+	return event == "SessionStart" && source == "compact"
 }
 
 // handleHookHandler processes a Claude Code hook event.
@@ -98,6 +109,14 @@ func handleHookHandler() {
 			"claudeSession", payload.SessionID,
 			"source", payload.Source,
 		)
+		return
+	}
+
+	// A compaction's closing SessionStart must not force "finished" (see
+	// isCompactSessionStart): keep the prior status file so PreCompact's "running"
+	// stands and detection settles the end.
+	if isCompactSessionStart(payload.HookEventName, payload.Source) {
+		log.Debug("hook-handler: skipping compact SessionStart", "instance", instanceID)
 		return
 	}
 
--- a/internal/ui/snapshot.go
+++ b/internal/ui/snapshot.go
@@ -11,11 +11,11 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/brizzai/fleet/internal/debuglog"
 	"github.com/brizzai/fleet/internal/hooks"
+	"github.com/brizzai/fleet/internal/proc"
 	"github.com/brizzai/fleet/internal/session"
 )
 
@@ -324,6 +324,7 @@ func buildClaudeLogBlock(st *claudeLogStat, hookUpdatedAt, now time.Time) map[st
 // the untouched file bytes so the snapshot can freeze a verbatim copy.
 type claudeSessionStatus struct {
 	raw             []byte
+	alive           bool   // liveness verdict, computed once during the match
 	PID             int    `json:"pid"`
 	SessionID       string `json:"sessionId"`
 	Status          string `json:"status"`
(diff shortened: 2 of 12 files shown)


cmd/fleet/hook_handler.go
```diff
@@ -33,6 +33,12 @@ func mapEventToStatus(event string) string {
 		return "running"
 	case "Stop":
 		return "finished"
+	case "PreCompact":
+		// /compact (and auto-compaction) is a multi-minute busy phase that fires no
+		// other hook. Without this, the session reads finished/idle for the whole
+		// compaction (SessionStart→finished on completion closes the bracket, and
+		// any queued prompt then flips it back to running via UserPromptSubmit).
+		return "running"
```
hayke102, line 42: **Dismissed idle session gets un-dismissed by `/compact`.** `PreCompact→running` flows into `applyHookRunning`, which unconditionally sets `s.Acknowledged = false` (`internal/session/session.go:849`). So if the user had *dismissed* an idle session and then runs `/compact` on it, the acknowledgment is wiped — and when compaction ends, `SessionStart→finished` maps to `StatusFinished` (not `StatusIdle`), so the session re-surfaces with a finished dot and re-enters the `Space`-jump attention cycle after a command the user expected to be a no-op.

Before this change the session stayed Idle. Worth gating the ack reset (or restoring Idle on the closing `SessionStart` when the session was previously acknowledged) so maintenance commands don't resurrect dismissed sessions.

cmd/fleet/hook_handler.go
```diff
@@ -33,6 +33,12 @@ func mapEventToStatus(event string) string {
 		return "running"
 	case "Stop":
 		return "finished"
+	case "PreCompact":
+		// /compact (and auto-compaction) is a multi-minute busy phase that fires no
+		// other hook. Without this, the session reads finished/idle for the whole
+		// compaction (SessionStart→finished on completion closes the bracket, and
```
hayke102: **Auto-compaction can flash "finished" mid-turn.** On auto-compaction during a long turn: `PreCompact→running` (this fix), then when compaction completes Claude fires `SessionStart(source="compact")` which `mapEventToStatus` returns `"finished"` for (line 47). The session flips to Finished mid-work — green dot, shows in the Space cycle — until a later worker tick recovers it to running via the transcript tiebreaker.

So the fix trades "stuck finished *during* compaction" for a briefer "finished flash *right after*" on every auto-compaction. Env/timing-dependent (PLAUSIBLE, not confirmed). Might be worth mapping `SessionStart` with `source=compact` to running (or ignoring it) rather than finished — otherwise the closing bracket keeps re-firing.

internal/ui/snapshot.go
```diff
@@ -303,6 +318,97 @@ func buildClaudeLogBlock(st *claudeLogStat, hookUpdatedAt, now time.Time) map[st
...
+// value while alive, not status_age — the flag is event-driven, not a heartbeat,
+// and it stays "idle" through /compact (so it can't explain a compaction stall).
+func buildClaudeStatusBlock(css *claudeSessionStatus, now time.Time) map[string]any {
+	if css == nil {
+		return nil
+	}
+	m := map[string]any{
+		"file":      "claude_session_status.json",
+		"pid":       css.PID,
+		"pid_alive": pidAlive(css.PID),
```
hayke102: **PID reuse can report a dead session's stale status as live.** If a session ends and its `claude_session_status.json` lingers with a last status (e.g. `busy`), the OS can recycle its PID for an unrelated process. `pidAlive` then returns true and this emits `pid_alive: true, status: busy`. The debug-status skill's own rule is "trust the value while `pid_alive` is true, regardless of age" — so the snapshot actively misleads the debugger into thinking a finished session is still running.

Narrow trigger (PLAUSIBLE), and it's an inherited class rather than newly introduced, but since the snapshot is a debugging aid the false-positive is worth guarding — e.g. cross-check the process identity, or note in the block that liveness is PID-only.

internal/ui/snapshot.go
```diff
@@ -303,6 +318,97 @@ func buildClaudeLogBlock(st *claudeLogStat, hookUpdatedAt, now time.Time) map[st
...
+		m["status_updated_at"] = t.UTC().Format(time.RFC3339Nano)
+		m["status_age"] = fmtSnapshotAge(t, now)
+	}
+	return m
+}
+
+// pidAlive reports whether a process with pid currently exists. Mirrors
+// internal/proc's liveness probe: signal 0 checks existence without delivering a
+// signal. A dead pid means a status file is only a last-known value.
+func pidAlive(pid int) bool {
```
hayke102: **Two small cleanups on `pidAlive` (both confirmed):**

1. It re-implements `internal/proc.alive` verbatim — `proc.go:181` already has `func alive(pid int) bool { return syscall.Kill(pid, 0) == nil }`, and this comment even says it "Mirrors internal/proc's liveness probe." Two copies will silently diverge if liveness semantics ever change (e.g. treating `EPERM` as alive). Consider exporting `proc.Alive` and calling it.
2. `findClaudeSessionStatusIn` already computes `alive := pidAlive(css.PID)` (line 367) to pick the winner, then discards it; `buildClaudeStatusBlock` re-probes the same PID at line 389. Redundant syscall per snapshot, plus a small window where the two verdicts disagree if the process exits between the calls. Thread the known verdict through instead.

hayke102 (commented):
Code review of the compacting-status + snapshot changes (xhigh, multi-agent + verify pass). Two correctness concerns from the new `PreCompact→running` mapping, one PID-reuse concern in the snapshot diagnostics, plus two small cleanups. Left as resolvable threads.

cmd/fleet/hook_handler.go
hayke102: Good catch. Accepting this as a narrow tradeoff: it only fires when you attach to a *dismissed* session and run `/compact`, and re-dismissing is one `Space` away. The clean fix (plumbing the hook event into `applyHookRunning` so PreCompact-sourced running skips the ack reset) adds disproportionate complexity for this flow, so leaving as-is for now.

cmd/fleet/hook_handler.go
hayke102: Fixed in e15e489 — `handleHookHandler` now skips `SessionStart(source="compact")` so the prior running (from PreCompact) stands; no finished flash during auto-compaction. A manual /compact at an idle prompt settles via pane + stability detection.

internal/ui/snapshot.go
hayke102: Fixed in e15e489 — annotated that `pid_alive` is PID-only (a recycled PID can report a dead session's last status as live) in the block-builder comment and the debug-status reference. Kept it a note rather than an identity cross-check, proportionate for a debug aid.

internal/ui/snapshot.go
hayke102: Fixed in e15e489 — exported `proc.Alive` (guarded for pid<=0) and reused it instead of the duplicate `pidAlive`; threaded the liveness verdict through `claudeSessionStatus` so `buildClaudeStatusBlock` reads it instead of re-probing. One liveness syscall per matched file now.

PR #170, opened 2026-07-06
feat(ui): notice when macOS TCC blocks tmux from ~/Documents

hayke102:
## Problem

A user's friend hit `ls: .: Operation not permitted` / `claude: EPERM` for a project in `~/Documents/work/dog-scratch`, while plain (non-fleet) terminals worked fine in the same folder.

**Root cause (confirmed by research + a code trace, not a guess):** macOS TCC gates `~/Documents`, `~/Desktop`, `~/Downloads`. tmux daemonizes its server (double-fork + `setsid`), so the server becomes its own TCC *responsible process* and loses the terminal's Documents grant. Every pane under it — agent sessions **and** drawer shells — inherits that grantless attribution. It's a known, live issue for tmux-wrapping tools on macOS Tahoe (see [cmux#2866](https://github.com/manaflow-ai/cmux/issues/2866)); **not a fleet defect.**

**What we can't do (verified, so we don't pretend to):**
- No code fix — a trace proved agent sessions and drawer shells share one default tmux server via a byte-identical `tmux new-session -d -c <dir>` call. fleet can't change tmux's daemonization or hand out TCC grants.
- No native "would you like to allow access?" popup — that needs a signed `.app` bundle as the responsible process. fleet is a CLI, the access happens in the daemonized tmux server, and there is no API to request a folder prompt on demand.

So the ceiling is **detect early → explain clearly → shortest path to the fix.**

## What this does

When a session/shell is created under a protected folder, fleet **probes actual access** and, if tmux is blocked, shows a persistent bottom-right notice:

> ⚠ macOS blocks tmux from `~/Documents` — sessions there fail with "Operation not permitted". Move the project outside it, or grant Full Disk Access to your terminal + tmux and run `tmux kill-server`. Press Ctrl+K → "Open Full Disk Access Settings".

That palette command deep-links straight to the macOS Full Disk Access pane — the most graceful path the platform allows for a CLI. The notice clears itself once the probe passes (after the user grants FDA + `tmux kill-server`, or moves the folder).

## How

| Piece | Where |
|---|---|
| `DirAccessBlocked` — a `run-shell ls` **inside the tmux server** so it reproduces the pane's TCC attribution (not fleet's own process); best-effort, `FLEET_NO_TCC_WARNING` opt-out | `internal/tmux/tmux.go` |
| Lazy probe, cached per protected root per launch, wired at session create + fork + shell create | `internal/ui/app.go`, `internal/ui/drawer.go` |
| `tipRecurring` notice (priority 120) + palette command gated to when blocked | `internal/ui/tips.go`, `internal/ui/app.go` |
| Unit tests for path classification + flag logic | `internal/ui/tcc_test.go` |

## Correctness properties
- **Zero false positives** — the probe tests the real shell, so a user who granted access (verified locally: `~/Documents` probes *not blocked*) sees nothing.
- **World-agnostic** — probing sidesteps the "does Claude work but the shell doesn't?" ambiguity; we test the thing that fails.
- **Covers agent sessions + forks + shells**, since they all share one daemon.
- Dropped the plan's separate inline drawer hint: the drawer isn't a modal, so the bottom-right notice already shows with the drawer open — one surface, not two.

## Verification
- `make build`, `make test` (incl. new `tcc_test.go`), `make lint` (0 issues) all green.
- Probe round-trip validated on real tmux 3.5a: readable dir → not blocked, unreadable → blocked, `run-shell` is synchronous and writes the exit code as assumed.
- Full manual end-to-end steps (blocked folder → notice → FDA key → grant + kill-server → notice clears → move folder → no notice) in the plan file.

## Summary by CodeRabbit

* **New Features**
  * Added a macOS security notice when tmux can’t access projects in protected folders like Documents, Desktop, or Downloads.
  * Added an in-app shortcut to open Full Disk Access settings when this issue is detected.
  * Added a recurring tip that explains the blocked-folder issue and suggests how to fix it.

* **Bug Fixes**
  * Replaced the vague “Operation not permitted” experience with clearer guidance for affected sessions.


internal/ui/app.go
```diff
@@ -2409,7 +2425,98 @@ func (h *Home) handleSessionCreate(msg sessionCreateMsg) (tea.Model, tea.Cmd) {
 		}
 	}
 	msg.agent = ag
-	return h, h.startSessionCmd(msg)
+	return h, tea.Batch(h.probeTCCCmd(msg.path), h.startSessionCmd(msg))
```
hayke102: **Cold-start race can silently disable this feature.** `probeTCCCmd` is batched alongside `startSessionCmd`, so on a fresh launch with no pre-existing tmux server the two race — and the probe usually loses. The probe's `tmux run-shell` finds no server yet (`new-session` is still spawning the daemon), so `DirAccessBlocked` fails open to `false` (tmux.go:133), and `probeTCCCmd` has already latched `h.tccProbed[root]=true` optimistically (line 2470) — so it never re-probes for the rest of the launch.

Net effect: the first session in `~/Documents/…` — the exact case this feature targets — dies with "Operation not permitted" and neither the tip nor the *Open Full Disk Access* command ever appears. The optimistic latch turns a transient false-negative into a permanent one.

Consider only latching `tccProbed[root]` once a probe has actually run against a live server, and treating "no server / tmux error" as *unknown* (re-probe later) rather than *not blocked*.

internal/ui/app.go
```diff
@@ -2409,7 +2425,98 @@ func (h *Home) handleSessionCreate(msg sessionCreateMsg) (tea.Model, tea.Cmd) {
...
+	if err != nil || path == "" {
+		return "", false
+	}
+	abs, err := filepath.Abs(path)
+	if err != nil {
+		return "", false
+	}
+	for _, name := range []string{"Documents", "Desktop", "Downloads"} {
+		root := filepath.Join(home, name)
+		if abs == root || strings.HasPrefix(abs, root+string(os.PathSeparator)) {
```
hayke102: **`protectedRoot` misses two classes of genuinely-protected paths**, so the probe/tip never fire for them:

1. **Case-sensitivity** (this line): `strings.HasPrefix` is case-sensitive, but the default macOS APFS volume is case-*insensitive*. A session at `~/documents/proj` (lowercase, valid on disk) won't match `~/Documents/` → returns `("", false)`.
2. **Symlinks** (line 2447): only `filepath.Abs` is used, no `EvalSymlinks`. A repo reached via `~/code/app → ~/Documents/app` resolves to a path that never contains "Documents", so it isn't detected even though the pane's cwd hits the TCC wall.

Both end in the same cryptic "Operation not permitted" this feature is meant to explain. `EvalSymlinks` + a case-folded compare would cover both.

internal/tmux/tmux.go
```diff
@@ -100,6 +100,52 @@ func envIsTruthy(name string) bool {
...
+	defer cancel()
+	if err := exec.CommandContext(ctx, "tmux", "run-shell", script).Run(); err != nil {
+		return false // no server yet, or tmux error — fail open.
+	}
+	out, err := os.ReadFile(tmpPath)
+	if err != nil {
+		return false
+	}
+	code := strings.TrimSpace(string(out))
+	return code != "" && code != "0"
```
hayke102: **Two edge cases in the exit-code interpretation:**

- **False negative (empty temp file):** if `ls` was denied but the server couldn't write the temp file (sandbox/permission hiccup, or the `run-shell` child torn down before flush), `code == ""` → returns `false` (not blocked). Combined with `probeTCCCmd`'s one-shot latch, a genuinely blocked folder then shows no tip at all.
- **False positive (non-TCC failure):** *any* non-zero exit is treated as a TCC block — a deleted/renamed root or a transient error surfaces the "macOS blocks tmux, grant Full Disk Access" tip when that isn't the cause, contradicting the "never warn on a false alarm" intent in the doc comment.

Distinguishing the specific EPERM / "Operation not permitted" condition from a generic non-zero exit would tighten both ends.

internal/ui/app.go
```diff
@@ -2409,7 +2425,98 @@ func (h *Home) handleSessionCreate(msg sessionCreateMsg) (tea.Model, tea.Cmd) {
...
+	}
+	if len(roots) == 0 {
+		return ""
+	}
+	sort.Strings(roots)
+	return roots[0]
+}
+
+// tildeHome shortens a home-relative absolute path to ~/… for display.
+func tildeHome(path string) string {
```
hayke102: Minor: `tildeHome` duplicates the existing `shortenPath` in this same `ui` package (preview.go:212), which does the same home→`~/…` conversion with the same separator-boundary check. It's also subtly weaker — `HasPrefix(path, home+sep)` alone means a path exactly equal to `$HOME` renders as the literal home path instead of `~` (shortenPath has an explicit `if p == home { return "~" }`). `shortenPath(h.firstTCCBlockedRoot())` would drop the duplicate.

hayke102 (commented):
Automated high-effort review of the TCC-blocked-folder notice. Four threads below — the first is the important one (feature can silently no-op on cold start); the rest widen coverage or tidy up.

internal/ui/app.go
```diff
@@ -2409,7 +2425,98 @@ func (h *Home) handleSessionCreate(msg sessionCreateMsg) (tea.Model, tea.Cmd) {
...
+func (h *Home) probeTCCCmd(path string) tea.Cmd {
+	root, ok := protectedRoot(path)
+	if !ok || h.tccProbed[root] {
+		return nil
+	}
+	h.tccProbed[root] = true // optimistic: don't double-probe while in flight
+	return func() tea.Msg {
+		return tccProbeResultMsg{root: root, blocked: tmux.DirAccessBlocked(root)}
+	}
+}
```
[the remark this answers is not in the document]
hayke102: Fixed in 6ccb536 — the probe now runs from the session/shell result handlers (after Start() has spun up the tmux server), and DirAccessBlocked is tri-state (blocked, determined): an inconclusive result unlatches tccProbed so a later create re-probes instead of caching a false negative.

internal/ui/app.go
hayke102: Fixed in 6ccb536 — probe moved post-Start (server guaranteed live) + tccProbed is only latched on a definitive result; 'no server / tmux error' is now treated as unknown and re-probed.

internal/ui/app.go
hayke102: Fixed in 6ccb536 — protectedRoot now resolves symlinks (best-effort filepath.EvalSymlinks, falling back to abs) and compares case-insensitively (EqualFold + a ToLower-prefixed separator boundary), so ~/documents and repos symlinked into a protected folder are caught. Added a lowercase test case.

internal/tmux/tmux.go
hayke102: Fixed in 6ccb536 — the probe now captures ls stderr and treats the folder as blocked only when it contains "Operation not permitted". Empty stderr → accessible; a non-TCC error (deleted root, transient) → not blocked; run/read failure → inconclusive (determined=false → re-probe). Verified: a nonexistent dir no longer false-positives.

internal/ui/app.go
hayke102: Fixed in 6ccb536 — dropped tildeHome for shortenPath (preview.go), which also handles the p == $HOME case. Removed the now-dead helper and its test.

PR #171, opened 2026-07-06
feat(ui): drawer shells labeled by last command + shown in collapsed bottom border

hayke102:
Three related drawer/shell improvements.

## 1. Shell tab name = latest command it ran (truncated)
- `tmux.RefreshSessionCache` now also caches each pane's `pane_pid` (new `tmux.PanePID`).
- New `proc.ForegroundCommands` — one `ps` call maps every shell's pane-shell PID to its foreground child's full command line (pipeline → leftmost/first command wins).
- `shell.Shell` gains `lastCmd` + `SetLastCommand`/`DisplayName`. Resolved on the **2s heavy** worker cadence and **held while idle**, so a stopped dev server keeps its label with an idle (○) dot. `Name` stays the stable fallback (`shell`, `shell 2`).
- Labels truncate to 22 runes with `…`; exited tabs show e.g. `npm run dev…(1)`.

## 2. Collapsed shells inline in the bottom border
- New `collapsedShellChips` renders the selected repo's **1–3 shells** (status dot + name, `+N` overflow); empty while the drawer is open (the drawer shows them then).
- Chips ride the **bottom-most panel where the drawer opens**: preview in dual/stacked, sidebar in single. Added a `footerLeft` inset to the bordered-panel renderer (`RenderBorderedPanelInsets`), mirroring the top border's title+right logic; when the border is too tight it drops the chips and keeps the cwd.

```
╰─ ● npm run dev  ○ tail -f app.…  +2 ── ~/code/fleet ─╯
```

## 3. Selected shell tab gets a background
- The active drawer tab now renders as the sidebar's filled accent pill (`Foreground(ColorBg).Background(ColorAccent)`, bold, padded) instead of plain bold text.

## Tests
- `parseForeground` unit tests (idle shell → no child, pipeline picks lowest pid, grandchildren excluded).
- Verified panel border widths line up exactly with the new insets across widths 24–60, including the tight-drop fallback.
- `make build` + full `make test` (`-race`) green.

https://claude.ai/code/session_01YGr6N867gHLZQaohsnLdge

## Summary by CodeRabbit

* **New Features**
  * Drawer tabs now show the most recent command run in each shell.
  * The active drawer tab is highlighted more clearly.
  * When the drawer is closed, the panel border now shows a compact summary of active shell commands, with overflow handling for extra shells.

### DIFF
--- a/internal/proc/proc.go
+++ b/internal/proc/proc.go
@@ -154,10 +154,10 @@ func normalizeCmd(s string) string {
 }
 
 // ForegroundCommands maps each given pane-shell pid to the command line of the
-// process it is currently running — the shell's own child process (e.g. a
-// `pane_pid` shell running "npm run dev" has an npm child). Shells sitting at a
-// prompt have no child and are absent from the result. One `ps` call regardless
-// of how many pids are passed. macOS-only.
+// process it is currently running — the leader of its tty's foreground process
+// group (e.g. a `pane_pid` shell running "npm run dev" reports "npm run dev").
+// Background jobs are excluded and a shell sitting at its prompt is absent from
+// the result. One `ps` call regardless of how many pids are passed. macOS-only.
 func ForegroundCommands(shellPIDs []int) map[int]string {
 	if len(shellPIDs) == 0 {
 		return nil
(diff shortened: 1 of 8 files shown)


internal/proc/proc.go
```diff
@@ -153,6 +153,84 @@ func normalizeCmd(s string) string {
...
+		}
+		pid, err := strconv.Atoi(pidStr)
+		if err != nil {
+			continue
+		}
+		cmd = strings.TrimSpace(cmd)
+		if cmd == "" {
+			continue
+		}
+		if cur, seen := best[ppid]; !seen || pid < cur.pid {
```
hayke102: **Foreground-child selection can mislabel a shell with a lingering background job.** The `pid < cur.pid` tiebreaker keeps the shell's *lowest-pid* (oldest) child, not the process currently in the foreground. If a drawer shell runs `some-server &` (say pid 500) and then `npm run dev` (pid 600) — both direct children of the pane shell — `want[ppid]` matches both and this keeps pid 500, so the tab and the collapsed bottom-border chip read `some-server` while the shell is actually running `npm run dev`, and stay wrong for as long as the background job outlives it. That contradicts the doc comment ("the process it is currently running").

Niche (needs a backgrounded job alongside a foreground command in the same shell) and only cosmetic, so not a blocker — but worth a note, or preferring the highest pid / tmux's own foreground-process info if you want it to track the current command.

internal/ui/styles.go
```diff
@@ -235,20 +235,28 @@ func ApplyPalette(p Palette) {
...
 // right-aligned status pill into the TOP border (after the title).
 func RenderBorderedPanelTopRight(content, title, titleRight string, width, height int, accent bool) string {
-	return renderBorderedPanel(content, title, titleRight, "", width, height, accent)
+	return renderBorderedPanel(content, title, titleRight, "", "", width, height, accent)
 }
 
 // RenderBorderedPanelFooter is like RenderBorderedPanel but also embeds a
 // right-aligned status pill into the bottom border. Empty footer renders an
 // unbroken bottom border.
 func RenderBorderedPanelFooter(content, title, footerRight string, width, height int, accent bool) string {
```
hayke102: **`RenderBorderedPanelFooter` is now orphaned by this PR.** Both of its callers (the stacked + dual preview panels in `app.go`) were switched to `RenderBorderedPanelInsets`, and `grep -rn RenderBorderedPanelFooter internal/` shows no remaining non-test callers — only this definition and comments. Safe to delete the wrapper (and fold its mention in the `RenderBorderedPanel` doc comment).

internal/ui/app.go
```diff
@@ -4375,6 +4386,24 @@ func (h *Home) statusWorkerCycle() {
...
+	if heavy && len(shells) > 0 {
+		pids := make([]int, 0, len(shells))
+		for _, sh := range shells {
+			if pid := tmux.PanePID(sh.TmuxName()); pid > 0 {
+				pids = append(pids, pid)
+			}
+		}
+		if fg := proc.ForegroundCommands(pids); len(fg) > 0 {
+			for _, sh := range shells {
+				sh.SetLastCommand(fg[tmux.PanePID(sh.TmuxName())])
```
hayke102: Minor: `tmux.PanePID(sh.TmuxName())` is called twice per shell each heavy cycle — once to build `pids` (line 4396) and again to key into `fg` here — each taking the `sessionCache` RLock, so 2N locked lookups for N shells. Building a `map[int]*shell.Shell` (pid→shell) in the first loop and iterating that for `SetLastCommand` drops the redundant re-lookup.

internal/ui/styles.go
hayke102: Fixed in 89756e1 — deleted the orphaned `RenderBorderedPanelFooter` and repointed the `RenderBorderedPanel` doc comment at `RenderBorderedPanelInsets`.

internal/ui/app.go
hayke102: Fixed in 89756e1 — the first loop now builds a `map[int]*shell.Shell` (pid→shell), so `tmux.PanePID` is taken once per shell per cycle instead of twice.

internal/proc/proc.go
hayke102: Fixed in 89756e1 — took your "tmux's own foreground-process info" suggestion to its logical end: `ForegroundCommands` now reads the tty's foreground process group (`tpgid`, same single `ps` call) and picks that group's leader. A backgrounded job in a different group no longer wins, a pipeline resolves to its first stage (the leader), and an idle shell yields nothing so its previous command persists. Verified with a live tmux repro (`sleep 300 &` + `sleep 400` → reports `sleep 400`).

PR #172, opened 2026-07-06
feat(ui): auto-suspend idle sessions under memory pressure

hayke102:
## Why

A full fleet of idle `claude --resume` processes (~400 MB each) can exhaust RAM/swap and get the **shared tmux server OOM-killed** — taking every session down at once (observed live: 51 sessions reporting `tmux list-panes: exit status 1` within 0.25s, swap at 10.5/12 GB). This adds graceful degradation: hibernate the most-idle sessions under memory pressure, freeing both the agent process **and** the tmux server's pane buffers, and resume lazily on attach.

## What

- **New `StatusSuspended` state** (dim `◌`, not error-red). Short-circuits the top of `UpdateStatus` so its intentionally-gone tmux isn't flagged as an error + crash dump; `UpdateHookStatus` ignores the killed agent's `SessionEnd` death-rattle.
- **`Session.Suspend()`** = `clearHookState` + kill the whole tmux session + set `StatusSuspended`. `ClaudeSessionID` is persisted, so the conversation survives. **Resume reuses `Restart()`** (`--resume <id>`) — no new resume logic.
- **Worker sweep** (`maybeSuspendIdleSessions`, throttled ~20s) probes `perfwatch.MemoryPressure()` — `kern.memorystatus_vm_pressure_level` + free swap, a far better OOM signal than "Pages free" — and suspends **only `StatusIdle`** sessions, ≤5 per pass (gentle shedding, re-checks each sweep).
- **Configurable aggressiveness** `session_suspend_mode` (Settings › Behavior › "Idle-session suspend"), default **Light**:

  | Mode | Suspends idle sessions when |
  |---|---|
  | Off | never |
  | **Light** | critical pressure + idle >2h |
  | Balanced | warning + idle >45m, or idle >4h anytime |
  | Aggressive | idle >20m (>10m under pressure) |

- **UX**: Enter on a `◌` row resumes + attaches; "suspended · enter to resume" on the focused row; a one-time tip the first time it fires; toast + palette commands ("Suspend This Session" / "Suspend Idle Sessions Now"); excluded from Reload All and status-summary pills.

## Logging
Info-level lines only when it **acts** (per-session `session suspend` + one sweep summary with mode/pressure/swap). The **why-it-didn't-act** trail is `Debug` (behind `FLEET_DEBUG`) — no ~20s Info churn.

## Tests
- `TestUpdateStatus_SuspendedShortCircuits` / `TestUpdateHookStatus_IgnoredWhileSuspended` — the correctness-critical guard that a suspended (tmux-gone) session isn't clobbered to error.
- `TestSuspendIdleThreshold` — table test pinning the mode/pressure matrix.
- `MemoryPressure()` verified against real `sysctl` output. Full `go test -race ./...` green; build + vet + gofmt clean.

## Note for review
Default **Light** means on upgrade, only *critical* memory pressure + multi-hour-idle sessions are auto-suspended — a quiet safety net. Only `Idle` sessions are ever auto-suspended (never running/waiting/starting/finished).

https://claude.ai/code/session_01FdAwb6Jg8NsGaDfNovXKLA

## Summary by CodeRabbit

* **New Features**
  * Added idle-session suspend to auto-hibernate eligible idle sessions under memory pressure, with lazy resume via Enter.
  * Added a configurable “Idle-session suspend” setting (off/light/balanced/aggressive) and documented the behavior in Settings.
  * Suspended sessions now show a dimmed status indicator plus sidebar prompt, and a one-time tip explains resuming.
* **Bug Fixes**
  * Suspended sessions no longer trigger liveness/error handling, reload-all restarts, or stale hook status updates.
  * Improved hook status cache purging on file remove/rename and added tests for suspend/resume and threshold behavior.


internal/ui/app.go
```diff
@@ -3167,6 +3210,209 @@ func (h *Home) restartSession(s *session.Session) tea.Cmd {
...
+
+	// Gather idle candidates first (cheap, no I/O) so we can skip the sysctl probe
+	// entirely when there's nothing we could shed.
+	type cand struct {
+		s    *session.Session
+		idle time.Duration
+	}
+	var cands []cand
+	for _, s := range sessions {
+		if s.GetStatus() == session.StatusIdle {
```
hayke102: **Auto-suspend can kill the session you're actively attached to.** The sweep only filters on `StatusIdle` and never excludes the currently-attached session. `IdleFor()` is measured from `LastAccessedAt` (set at attach time, not refreshed while attached), so if you attach and read/think without typing, this loop eventually calls `Suspend()` → `tmux.Kill()` on the session you're still in and ejects you into a dead pane.

- aggressive mode: fires after idle > 20min at normal pressure
- default **light** mode: still fires under memory-critical pressure after ~2h attached-idle

The worker runs on its own ticker, independent of the attach, and `maybeSuspendIdleSessions` never consults `h.isAttaching` or the attached session id. Suggest skipping the session that's currently attached (and ideally treating an active attach as activity that refreshes `LastAccessedAt`).

internal/ui/app.go
```diff
@@ -805,6 +819,29 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			}
 		}
 		h.rebuildFlatItems()
+		// A resume-from-suspend requested an attach once the tmux was live again.
+		if h.attachAfterResumeID == msg.id {
+			h.attachAfterResumeID = ""
+			if msg.err == nil {
+				return h, h.attachSelected()
```
hayke102: **Resume-from-suspend attaches the cursor's row, not the session that was resumed.** `resumeSelected` stashes `attachAfterResumeID = s.ID` then kicks off an async `Restart()` (recreate tmux + `--resume`, hundreds of ms). If the user presses `j`/`k` during that window, this handler matches `attachAfterResumeID == msg.id` but then calls `h.attachSelected()`, which attaches `h.flatItems[h.cursor].Session` — now a *different* session. The user is dropped into a session they didn't pick while the resumed one attaches to nothing.

Attach by the resumed id here (look the session up by `msg.id` and attach that), rather than whatever the cursor points at now.

internal/ui/app.go
```diff
@@ -2018,6 +2055,12 @@ func (h *Home) handleKey(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
 			h.toggleRepoGroup()
 			return h, nil
 		}
+		// A suspended session has no live tmux — resume it (recreates tmux +
+		// `--resume`), then attach once it's up. Overrides split mode: there is no
+		// live preview to focus.
+		if s := h.selectedSession(); s != nil && s.GetStatus() == session.StatusSuspended {
```
hayke102, line 2078: **The slot-hotkey double-tap silently no-ops on a suspended session.** This Enter path correctly routes `StatusSuspended` → `resumeSelected`, but the slot double-tap-to-attach path (`jumpToSlot` at app.go:5649) calls `attachSelected()` directly, which returns `nil` for a suspended session because its tmux is dead (`!s.IsAlive()` guard in `attachSelected`). So a slot-bound session that gets auto-suspended can't be woken via its hotkey and gives no feedback — nothing opens, no hint to press Enter. Route the suspended case through `resumeSelected` in `jumpToSlot` too.

internal/ui/app.go
```diff
@@ -3167,6 +3210,209 @@ func (h *Home) restartSession(s *session.Session) tea.Cmd {
...
+	case session.StatusSuspended:
+		h.setInfo("Session already suspended — enter to resume")
+		return nil
+	default:
+		h.setInfo("Only idle or finished sessions can be suspended")
+		return nil
+	}
+	id, title := s.ID, s.Title
+	return func() tea.Msg {
+		if err := s.Suspend(); err != nil {
```
hayke102, line 3489: **Manual suspend has a TOCTOU race with the status worker that can strand the session in `StatusError`.** This runs in a `tea.Cmd` goroutine. `Suspend()` kills tmux, then sets `StatusSuspended`. Meanwhile the worker may already be inside `UpdateStatus()`: it snapshots `oldStatus` (still `StatusIdle`, so the `StatusSuspended` short-circuit is missed), then hits `if !s.IsAlive()` → tmux now dead → `SetStatus(StatusError)` + a spurious crash dump. If that Error write lands after `Suspend()`'s `SetStatus(StatusSuspended)`, the row shows the red ✕, `UpdateStatus` no longer short-circuits it, and Enter won't wake it (resume is gated on `StatusSuspended`).

Note `Suspend()`'s `clearHookState()` only blocks the hook=dead path — this flip is liveness-based (`session.go:674`), so it isn't covered. The auto-sweep is immune because it calls `Suspend()` on the worker goroutine itself; these manual paths (`suspendSelected` here, `suspendIdleNow` at app.go:3370) don't establish `StatusSuspended` on the Update thread first. Consider setting the optimistic `StatusSuspended` on the Update thread before the goroutine, the way `resumeSelected` does.

hayke102 (commented):
Automated high-effort review of the idle-session-suspend feature — 5 findings on the suspend/resume lifecycle (3 confirmed correctness, 2 plausible). Threads are resolvable.

internal/session/session.go
```diff
@@ -579,6 +611,26 @@ func (s *Session) RespawnClaude() error {
 	return nil
 }
 
+// Suspend hibernates the session to free memory: it tears down the agent process
+// and the whole tmux session (freeing the tmux server's pane buffers too), and
+// marks the session StatusSuspended. The conversation survives — ClaudeSessionID
```
hayke102, line 616: **"The conversation survives" only holds when `ClaudeSessionID` is non-empty.** A session that reached `StatusIdle` before any `SessionStart` hook wrote `ClaudeSessionID` (id never captured) can still be auto-suspended by the sweep, which advertises "nothing is lost." On Enter, `Restart()` → `buildAgentCmd()` sees an empty `ResumeID` and launches a brand-new `claude` with no `--resume`, silently discarding the prior conversation. Worth gating suspend (or at least auto-suspend) on `ClaudeSessionID != ""` so we never park a session we can't actually resume.

internal/session/session.go
```diff
@@ -579,6 +611,26 @@ func (s *Session) RespawnClaude() error {
...
+	// Drop hook state first so the killed agent's SessionEnd ("dead") death-rattle
+	// can't be picked up and flip the row to error during/after teardown.
+	s.clearHookState()
+	if s.tmuxSession.Exists() {
+		if err := s.tmuxSession.Kill(); err != nil {
+			debuglog.Logger.Error("session suspend: kill failed", "id", s.ID, "title", s.Title, "err", err)
+			return err
+		}
+	}
+	s.SetStatus(StatusSuspended)
```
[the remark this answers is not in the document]
hayke102: Fixed — Suspend() now sets StatusSuspended before clearHookState/Kill (restoring the prior status on kill failure), so the status pipeline short-circuits during teardown.

internal/session/suspend_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// A suspended session must be left untouched by the status pipeline. Its tmux is
+// intentionally gone (pane dead), which would otherwise trip the liveness gates at
+// the top of UpdateStatus and flip it to error + write a spurious crash dump.
+func TestUpdateStatus_SuspendedShortCircuits(t *testing.T) {
+	debuglog.Init()
+	mock := &mockPane{dead: true} // dead pane → IsAlive()=false, IsPaneDead()=true
+	s := &Session{
+		ID:           "suspend-test",
```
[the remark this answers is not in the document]
hayke102: Fixed — test fixtures now use the <8hex>-<unix_timestamp> shape (HookStatus.SessionID left symbolic).

internal/ui/app.go
```diff
@@ -805,6 +819,29 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			}
 		}
 		h.rebuildFlatItems()
+		// A resume-from-suspend requested an attach once the tmux was live again.
+		if h.attachAfterResumeID == msg.id {
+			h.attachAfterResumeID = ""
+			if msg.err == nil {
+				return h, h.attachSelected()
+			}
+		}
```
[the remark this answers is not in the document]
hayke102: Fixed — a failed resume now resets the row to StatusSuspended (persisted) so Enter can retry, instead of leaving it stranded.

internal/ui/app.go
```diff
@@ -3167,6 +3210,209 @@ func (h *Home) restartSession(s *session.Session) tea.Cmd {
...
+
+	sort.Slice(cands, func(i, j int) bool { return cands[i].idle > cands[j].idle }) // most-idle first
+
+	n := 0
+	for _, c := range cands {
+		if n >= suspendSweepBatch || c.idle < minIdle {
+			break // sorted desc: once one is too fresh, so is everything after it
+		}
+		if err := c.s.Suspend(); err != nil { // Suspend() logs the per-session event
+			debuglog.Logger.Error("suspend idle session failed", "id", c.s.ID, "title", c.s.Title, "err", err)
```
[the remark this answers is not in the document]
hayke102: Fixed — the auto-sweep re-validates GetStatus()==Idle (and not the attached session) right before Suspend(); the manual paths now commit StatusSuspended on the Update thread first, so eligibility is checked there.

internal/ui/app.go
hayke102: Fixed — added h.attachedSessionID (set in attachSession, cleared on detach, guarded by workerMu); the sweep skips it at gather time and re-checks right before the kill, so it can't hibernate the pane you're attached to.

internal/ui/app.go
hayke102: Fixed — resume now attaches via attachSessionByID(msg.id) instead of the cursor-based attachSelected(), so moving the cursor during the async Restart no longer attaches the wrong session.

internal/ui/app.go
hayke102: Fixed — jumpToSlot's double-tap now routes a suspended session through resumeSelected (same as Enter) instead of the no-op attachSelected().

internal/ui/app.go
hayke102: Fixed — suspendSelected and suspendIdleNow now set StatusSuspended (and persist) on the Update thread before dispatching the teardown goroutine, so the worker sees Suspended and short-circuits instead of racing into StatusError.

internal/session/session.go
hayke102: Fixed — auto-suspend and manual suspend now skip/refuse sessions with an empty ClaudeSessionID (added GetClaudeSessionID), so we never park a session whose conversation we can't actually resume.

PR #177, opened 2026-07-07
feat(ui): release notes viewer in the command palette

hayke102:
## What

Adds a **Release Notes** command to the `Ctrl+K` palette that opens a scrollable, in-app changelog of every release — the running version is marked `INSTALLED`, newer releases are grouped up top as "update to get these", and older ones follow.

## Why

Nothing surfaced the changelog inside the running app — you couldn't see what's new, what shipped, or what's coming without leaving the TUI.

## How

- **`internal/releasenotes/`** — fetches `GET /repos/brizzai/fleet/releases` (public repo, no auth needed; a token is used only if present to raise rate limits), caches to `~/.config/fleet/releases.json` with a 1h freshness window, and falls back to stale cache when offline. Release bodies are the curated `CHANGELOG.md` sections (the release workflow feeds GoReleaser `--release-notes`), so the notes read as polished as the changelog. Parses out the install boilerplate before the `---`, then `### Section` + bullets. Includes version normalize/compare and a relative-time `Ago` helper.
- **`internal/ui/releasenotes.go`** — a scrollable dialog cloning `help.go`'s scroll idiom (`↑↓ j/k PgUp/PgDn g/G`, `⋮ +N above/below`). Uses ~80% of the terminal width, right-aligns the date into a column with a relative `(3 days ago)`, frames the header with a rule + release count, dim dividers between releases, an `INSTALLED` badge, color-coded section labels, and inline markdown rendering for `` `code` `` and `**bold**` (proper CommonMark code-span handling so `` `` ` `` `` literal-backtick spans don't bleed).
- **`internal/ui/app.go`** — the 8 standard dialog integration points + palette entry + a `dispatchCommand` case + an async `loadReleaseNotes()` command (network runs off the Update goroutine).
- **`changelog/README.md` + `CLAUDE.md`** — since fragments now render in-app, codify a "fun to read" voice guide (bold headline, second person, name the pain, numbers over adjectives, `code`-format concretes, no hype).

## Testing

- `go build ./...` clean; `go test -race` green for both packages.
- Unit tests: release parsing/normalize/compare/`ago`, dialog grouping, INSTALLED marker, dev-build handling, no-line-overflow, loading/error states, and inline-markdown (single/double-backtick/bold/unclosed).
- Verified with live renders against the real GitHub API.

To try it: `Ctrl+K` → type "release" → Enter.

https://claude.ai/code/session_01CZBN2UXd2x1U4etNuKyzRs

## Summary by CodeRabbit

* **New Features**
  * Added an in-app Release Notes viewer accessible from the command palette.
  * Release notes now open in a scrollable modal that highlights the current version and newer available releases.
  * Notes support clear formatting for headings, code snippets, and emphasized text.

* **Documentation**
  * Updated contributor guidance for writing changelog entries with a more readable, reader-focused style.
  * Expanded examples to show the preferred release-note format and tone.

* **Bug Fixes**
  * Improved version handling so installed, newer, and prerelease releases are shown more accurately.

### DIFF
--- a/internal/releasenotes/releasenotes.go
+++ b/internal/releasenotes/releasenotes.go
@@ -1,0 +1,286 @@
+// Package releasenotes fetches fleet's GitHub Releases so the TUI can show an
+// in-app changelog. Content is cached to ~/.config/fleet so reopening is instant
+// and works offline. Each release body is the curated CHANGELOG section (the
+// release workflow feeds GoReleaser --release-notes built from CHANGELOG.md), so
+// what we render reads exactly like the changelog.
+package releasenotes
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const (
+	repo         = "brizzai/fleet"
+	apiURL       = "https://api.github.com/repos/" + repo + "/releases?per_page=100"
+	cacheFile    = "releases.json"
+	cacheMaxAge  = time.Hour // reuse the cache within this window; older triggers a refresh
+	fetchTimeout = 3 * time.Second
+)
+
+// Section is one "### Added"/"### Fixed" group parsed from a release body.
+type Section struct {
+	Title   string   `json:"title"`
+	Bullets []string `json:"bullets"`
+}
+
+// Release is a normalized GitHub release ready for rendering.
+type Release struct {
+	Version    string    `json:"version"` // normalized, e.g. "2.15.0"
+	Name       string    `json:"name"`    // release name / tag as shown by GitHub
+	Date       string    `json:"date"`    // "2006-01-02", from published_at
+	Sections   []Section `json:"sections"`
+	Prerelease bool      `json:"prerelease"`
+	URL        string    `json:"url"` // html_url
+}
+
+// configDir returns ~/.config/fleet/, matching internal/update and the rest of fleet.
+func configDir() string {
+	home, _ := os.UserHomeDir()
+	return filepath.Join(home, ".config", "fleet")
+}
+
+// Load returns the release list, preferring a fresh on-disk cache and otherwise
+// fetching from GitHub. On a network error it falls back to a stale cache; it
+// only returns an error when there's no cache to fall back to.
+func Load() ([]Release, error) {
+	cached, cachedAt, haveCache := readCache()
+	if haveCache && time.Since(cachedAt) < cacheMaxAge {
+		return cached, nil
+	}
+
+	fresh, err := fetch()
+	if err != nil {
+		if haveCache {
+			return cached, nil // offline / rate-limited: stale is better than nothing
+		}
+		return nil, err
+	}
+	writeCache(fresh)
+	return fresh, nil
+}
+
+// ghRelease is the subset of the GitHub Releases API response we use.
+type ghRelease struct {
+	TagName     string `json:"tag_name"`
+	Name        string `json:"name"`
+	PublishedAt string `json:"published_at"`
+	Body        string `json:"body"`
+	Prerelease  bool   `json:"prerelease"`
+	Draft       bool   `json:"draft"`
+	HTMLURL     string `json:"html_url"`
+}
+
+// fetch pulls all releases from GitHub and normalizes them, newest-first.
+func fetch() ([]Release, error) {
+	client := &http.Client{Timeout: fetchTimeout}
+	req, err := http.NewRequest("GET", apiURL, nil)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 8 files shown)


hayke102 (commented):
A few notes from reviewing the release-notes viewer — none blocking, all optional. Core logic (parseBody pointer pattern, markdown parser, async load) looks solid.

internal/releasenotes/releasenotes.go
```diff
@@ -0,0 +1,321 @@
...
+	Version    string    `json:"version"` // normalized, e.g. "2.15.0"
+	Name       string    `json:"name"`    // release name / tag as shown by GitHub
+	Date       string    `json:"date"`    // "2006-01-02", from published_at
+	Sections   []Section `json:"sections"`
+	Prerelease bool      `json:"prerelease"`
+	URL        string    `json:"url"` // html_url
+}
+
+// configDir returns ~/.config/fleet/, matching internal/update and the rest of fleet.
+func configDir() string {
```
hayke102, line 46: **Reuse:** `configDir()` here is byte-for-byte identical to `internal/update/update.go:37`, and `ghToken()` (line 231 below) is identical to `internal/update/update.go:64`. The `fetch()` HTTP boilerplate (client timeout, `Accept: application/vnd.github+json`, `Bearer`+token, `StatusCode != 200 → "GitHub API returned %d"`, `json.NewDecoder().Decode`) also mirrors `update.go:79`.

Not a bug, but a change to the config-dir path or token resolution now has to be made in two places or they silently diverge. Worth exporting one copy (e.g. a shared `config.Dir()` / small `ghauth` helper) and calling it from both packages.

internal/ui/releasenotes.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+
+	newerEmitted := false
+	sep := false // emit a divider before the next release?
+	for _, r := range d.releases {
+		cmp := 0
+		if !devBuild {
+			cmp = releasenotes.CompareVersions(r.Version, d.installed)
+		}
+		installed := !devBuild && cmp == 0
```
hayke102: **Double INSTALLED badge:** `installed := cmp == 0` tags *every* release whose normalized version equals the running one, but `NormalizeVersion` strips the pre-release suffix — so `v2.15.0-rc.1` and `v2.15.0` both collapse to `2.15.0`.

If both a prerelease (which this dialog explicitly renders a `pre-release` tag for) and its GA exist and you're running `2.15.0`, two different rows both get the accent `INSTALLED` badge. Cosmetic, but it tells the user two releases are the installed one. Could gate on an exact tag match, or skip the badge when `r.Prerelease`.

internal/ui/releasenotes.go
```diff
@@ -0,0 +1,410 @@
...
+
+		if !devBuild && cmp > 0 && !newerEmitted {
+			raw(lipgloss.NewStyle().Foreground(ColorAccent).Bold(true).Render("◆ NEWER — update to get these"))
```
hayke102: **Scroll desync on narrow terminals:** this banner is appended via `raw()` (untruncated), and the version-header rows via `padLR` (line 264) can also exceed `inner`. `scrollGeometry`/`contentLines` count each as one line, but lipgloss wraps them to two when they overflow the box's content width — so `content[scroll:scroll+visible]` in `View` drifts off by a row.

The `"◆ NEWER — update to get these"` string is ~29 display cols, so at terminal width ~40 (`inner ≈ 26`) it wraps and scrolling to `end` clips the last bullet/footer. Only bites on unusually narrow terminals, but the fix is a one-liner: route these rows through `fit()` (`ansi.Truncate` to `inner`) like the other content lines.

internal/ui/releasenotes.go
hayke102: Fixed — the installed marker now requires `!r.Prerelease`, so a prerelease and its GA (which normalize to the same version) no longer both get the badge. The GA keeps it; the prerelease renders as a normal row with its pre-release tag. Added a regression test.

internal/ui/releasenotes.go
hayke102: Fixed — the "◆ NEWER" header and the version/date row now go through `fit()` (truncate to `inner`), so they can't wrap on a narrow terminal and desync the scroll window. Extended the no-overflow test to widths 28/38/60/90/200 with the NEWER group present.

internal/releasenotes/releasenotes.go
hayke102: Good call. The `~/.config/fleet` path is inlined in ~18 places across the codebase (config, update, snapshot, hooks, chrome, session, diagnostics, …) — inlining it is the existing convention, so extracting just these two sites would be inconsistent. A shared `config.Dir()` + small `ghauth` helper is a worthwhile repo-wide cleanup, but out of scope for this feature PR; leaving the intentional decoupling (releasenotes not importing update) and tracking the dedup as a separate follow-up.

PR #178, opened 2026-07-07
fix(status): show sessions running background agents as Running

hayke102:
## What

A session that delegates to **background agents** (`Agent` tool + `run_in_background`) fires its lead `Stop` hook and parks until they re-invoke it — so it read **idle/finished** even while minutes of work (hundreds of thousands of tokens) were actively in flight.

This detects the background-agents dock from the pane and reports **Running**:

```
✻ Waiting for 3 background agents to finish
  ◯ Explore  Map analytics/platform pages   2m 49s · ↓ 78.1k tokens
  ◯ Explore  Map AI assistant, labels …     2m 35s · ↓ 104.1k tokens
  ◯ Explore  Classify settings/admin …      2m 22s · ↓ 65.6k tokens
```

`detectBackgroundAgentsRunning` matches the in-flight dock rows (`◯ <type> … · ↓/↑ <n> tokens`). The rows render **below** Claude's input box, so a match in the bottom lines is the live dock, never quoted scrollback. `applyHookFinished`'s existing `paneStatus == running` override handles the transition — no new state-machine plumbing. Only the in-flight `◯` glyph matches, so it **self-clears to finished** when the agents complete.

## Reverses a prior decision (deliberately)

This flips the earlier "finished-idle while a background agent runs" behavior:
- Golden `finished_idle_background_agent.txt`: `StatusFinished` → `StatusRunning`.
- New golden `pane_running_background_agents_multi.txt` (seeded from a real snapshot) + new `TestScenarioBackgroundAgentsShowRunning`.
- The two window-activity **bridge** scenarios (`…NotHeldByBackgroundAgent`, `…BridgesRecentHookThenReleases`) are re-pointed at a **dockless** idle fixture — the bridge/anti-pin mechanism they test is orthogonal to the dock, so it stays covered.

The old anti-"pin forever" concern was about a stale **`waiting`**; an honest **`running`** while agents work doesn't have that failure mode.

**Trade-off (accepted):** a long background job shows `Running` for its full duration (won't surface in the Space-to-jump attention queue until it completes) — chosen over hiding live work under `idle`.

## Why pane-scrape and not a "more first-party" signal

`docs/background-agent-detection.md` (decision record) has the full analysis. Short version: there's **no stable, documented** first-party signal for in-session background agents. Background agents get no status file of their own, no separate transcript, and no live `isSidechain` entries — verified on Claude 2.1.202. The Claude session status file *does* stay `busy`, but Anthropic's docs say its format "can break on any release", and it would break **silently, pipeline-wide**. The pane scrape's fragility is instead **tiny and loud** — one function + one fixture, a visibly-wrong status. The durable fix is upstream (see #7881 / #19170 references in the doc).

## Test plan
- `make build` ✅
- `go test -race ./internal/session/` ✅ (flipped golden, new golden, new scenario, re-pointed bridge scenarios all pass)

https://claude.ai/code/session_01FdAwb6Jg8NsGaDfNovXKLA


internal/session/session.go
```diff
@@ -1520,8 +1520,52 @@ func extractRecentLines(lines []string, n int) []string {
...
+// row lacks the "(…)" duration and ")" suffix isWhimsicalActivity requires, which is
+// why that check misses it — the "· ↓/↑ … tokens" trailer plus the glyph identifies it.
+func detectBackgroundAgentsRunning(recentLines []string) bool {
+	n := min(8, len(recentLines))
+	for _, line := range recentLines[:n] {
+		t := strings.TrimSpace(line)
+		if !strings.HasPrefix(t, backgroundAgentGlyph) {
+			continue
+		}
+		if strings.Contains(t, "tokens") &&
```
hayke102, line 1561: **Loose anchor can pin an idle session to Running from quoted text.** The match here is only `TrimSpace` starts-with `◯` + contains `tokens` + contains `· ↓`/`· ↑`. That's materially weaker than the sibling checks: busy patterns are gated to the bottom 5 lines and `isWhimsicalActivity` requires a `(<dur>)` paren anchor — both deliberately narrowed to dodge the meta-problem where the pane *discusses* these patterns.

The header comment claims quoted content "can only reach the upper 50-line window," but that's not right: an idle pane's bottom chrome is only ~5 non-empty lines (status bar, `❯`, status bar, model line, PR line), so `recentLines` indices 5–7 land on the **last 3 conversation lines** — inside this bottom-8 window.

**Failure:** a fleet dev asks Claude about this very feature (the row format is now documented verbatim in CLAUDE.md + `docs/background-agent-detection.md`). Claude's final turn echoes a row like `◯ Explore  Map analytics/platform pages  2m 49s · ↓ 78.1k tokens`. hook=finished, but `detectRunning` fires first → `paneStatus=Running` → `applyHookFinished` latches Status=Running. Content is now stable, so it's **stuck on Running** until the user types or the line scrolls off.

Suggest tightening the anchor — reuse the whimsical `(<dur>)` check or require the dock's two-space column layout — so quoted prose can't trip it.

internal/session/session.go
```diff
@@ -1520,8 +1520,52 @@ func extractRecentLines(lines []string, n int) []string {
...
+	}
+	return false
+}
+
 // detectRunning checks for busy indicators, spinner chars, and whimsical activity patterns.
 func detectRunning(recentLines []string, _ string, log *slog.Logger) Status {
+	// Background-agents dock: the lead delegated to run_in_background agents and
+	// parked (it fired Stop), but those agents are still in-flight — the session is
+	// running. Checked first: it's specific and bottom-anchored (scrollback-safe).
+	if detectBackgroundAgentsRunning(recentLines) {
```
hayke102: **Dock check runs before `detectWaiting`, so a background agent that needs permission reads Running instead of Waiting.** If a permission prompt renders in the lead pane while one or more `◯` rows are still shown, this returns `StatusRunning` before the waiting structural checks ever run. With hook=finished, that preempts `applyHookFinished`'s `paneStatus==StatusWaiting` override — the session shows Running, so Space-to-next-attention skips it and Quick-Approve (Y) is gated off, and it stalls until the user manually attaches.

_PLAUSIBLE_ — depends on whether a `run_in_background` sub-agent's permission prompt actually renders in the lead pane alongside live dock rows. Worth confirming; if it can, the waiting check should win over the dock in that frame.

hayke102 (commented):
Two false-positive/precedence edges in the new dock matcher — mechanism, override path, and ANSI handling all check out otherwise.

internal/session/session.go
hayke102: Fixed in 28e706c. You're right — verified idle chrome is exactly recentLines[0..4], so bottom-8 reached conversation lines at [5..7], and the row is now documented verbatim so it's a live trigger. Since ANSI is stripped, position is the only discriminator: narrowed the scan to the bottom 3 lines (where the live dock always sits; quoted rows land at index ≥5) and corrected the wrong "upper 50-line window" comment. Added golden pane_finished_quoted_dock_row (quoted row in idle pane → finished).

internal/session/session.go
hayke102: Fixed in 28e706c. Took the safe precedence regardless of the PLAUSIBLE condition: moved the dock check out of detectRunning into detectStatus after detectWaiting, so a permission prompt wins over the dock (running > waiting > dock > finished). Added golden pane_waiting_permission_with_dock (menu + live dock rows → waiting).

PR #180, opened 2026-07-08
feat(ui): animated "What's New" badge with curated release highlights

hayke102:
## What

A **"What's New"** experience for fleet:

- **Animated top-right badge** — a rainbow-shimmer `✦ What's New · press W` lights up when there's an **unseen highlighted release within the last 7 days**, and clears once viewed. It's a static Lab-blended ROYGBIV rainbow with a narrow glint that sweeps then rests (modeled on Claude Code's "ultrathink" shimmer). The ~60ms tick self-reschedules **only while the badge is visible**, so it costs no CPU when hidden.
- **The reel** — `W` (or `Ctrl+K → "What's New"`) opens the existing Release Notes dialog in a scoped mode showing **only each release's Highlights**, filtered to the last 7 days / since last seen, with a caught-up empty state. Opening it marks everything seen → badge clears.
- **Curation** — a changelog fragment opts in with `highlight: true`. `prepare-release.yml` copies those into a `### Highlights` block leading the release body; the runtime parser reads it as any other section (no parser change). The full changelog and Release Notes view are unchanged — the full view hides the curated duplicate.

## Notes

- **Highlights only appear from the next release onward** — older GitHub releases carry no `### Highlights` section, so the badge stays quiet until a release ships with a `highlight: true` fragment. This PR's own fragment is flagged, so it'll be the first to show.
- On a genuinely empty fleet where the first-run launchpad is showing, the badge is intentionally suppressed (`modalOpen()` includes the launchpad).
- Docs updated: `changelog/README.md`, `CLAUDE.md`, and the `update-changelog` skill (which was also stale about the fragment workflow generally — corrected here).

## Test plan

- `make build` ✅ · `make test` ✅ · `make lint` = 0 issues ✅
- New tests: badge render (stable width, animates, colored, `press W` hint), reel filtering (highlights-only, non-highlights excluded, caught-up state), full-view-hides-Highlights, `isUnseenHighlight`, `WithinLastDays`, `### Highlights` parsing.
- Release-pipeline awk reproduced locally: `highlight: true` → `### Highlights` block leads, item also present in its type section.
- Verified end-to-end against a seeded release cache in a throwaway `$HOME`: badge shows, `W` opens the reel, opening it persists "seen" and clears the badge across restarts.

## Summary by CodeRabbit

* **New Features**
  * Added a new “What’s New” view for recently highlighted release notes.
  * A top-right badge now appears when new highlights are available, with keyboard and command-palette access via **W**.
  * Release notes can now show a curated highlights reel separate from the full changelog.

* **Bug Fixes**
  * Improved release-note visibility so only recent, unseen highlights are surfaced.
  * Preserved a stable badge layout while the animation changes frame-to-frame.


internal/ui/app.go
```diff
@@ -6448,6 +6497,58 @@ func (h *Home) loadReleaseNotes() tea.Cmd {
...
+	seen := releasenotes.NormalizeVersion(h.cfg.GetReleaseNotesSeenVersion())
+	h.hasUnseenWhatsNew = false
+	for _, r := range h.cachedReleases {
+		if isUnseenHighlight(r, seen) {
+			h.hasUnseenWhatsNew = true
+			return
+		}
+	}
+}
+
```
hayke102, line 6556: **Badge over-stays its 7-day window on long-lived instances.** `recomputeWhatsNew()` (the only writer of `hasUnseenWhatsNew`) runs solely on `releaseNotesLoadedMsg`, which fires from `Init` and explicit opens — there's no periodic reload. `isUnseenHighlight` applies the `whatsNewWindowDays` (7) cap via `WithinLastDays(r.Date, …)`, but only at load time.

fleet is built to run for days/weeks (it even hibernates idle sessions for exactly that). So if a highlighted release is within 7 days at launch the badge lights, and once that release crosses the 7-day boundary while the app keeps running, `hasUnseenWhatsNew` is never re-evaluated and stays `true`. The `whatsNewTickMsg` loop (`app.go:1288`, gated on `!h.hasUnseenWhatsNew || h.modalOpen()`) then keeps firing every 60ms, re-rendering the whole TUI at ~16fps indefinitely — the badge documented as "capped to the last 7 days / quietly retires" never actually ages out at runtime; only a restart or pressing `W` clears it.

If you want the window to genuinely bound the badge (and let the animation stop), recompute on the ~2s tick — `handleTick` already calls `ensureWhatsNewShimmer`, so a `recomputeWhatsNew()` alongside it would let `hasUnseenWhatsNew` flip back to false and kill the loop once the release ages past 7 days.

internal/ui/app.go
```diff
@@ -213,6 +222,14 @@ type Home struct {
 	onboardingDialog      *OnboardingDialog
 	releaseNotes          *ReleaseNotesDialog
 
+	// What's New badge: an animated top-right indicator shown while an unseen
+	// highlighted release exists. cachedReleases is loaded once at startup so
+	// the badge can be computed without opening the dialog.
+	cachedReleases     []releasenotes.Release
+	hasUnseenWhatsNew  bool
+	whatsNewFrame      int  // shimmer animation frame
+	whatsNewShimmering bool // whether the ~80ms shimmer loop is currently armed
```
hayke102: **Nit (stale doc):** this says `~80ms`, but `whatsNewTickInterval` is `60 * time.Millisecond`. `CLAUDE.md`'s What's New entry repeats the "~80ms" figure and also calls it a "per-rune HSV rainbow", though `renderWhatsNewBadge` uses `lipgloss.Blend1D` (Lab-space blend of fixed RGB stops), not HSV. `whatsnew.go`'s own comments correctly say "~every 60ms" — worth aligning these two spots.

internal/ui/app.go
hayke102: Fixed — `handleTick` now calls `recomputeWhatsNew()` on each ~2s tick, so a highlighted release that crosses the 7-day boundary while fleet keeps running flips `hasUnseenWhatsNew` back to false; the next `whatsNewTickMsg` then stops the 60ms shimmer loop. Good catch — the badge/animation genuinely ages out at runtime now, not just on restart.

internal/ui/app.go
hayke102: Fixed — `~80ms`→`~60ms` in the app.go struct comment and CLAUDE.md, and corrected CLAUDE.md's 'per-rune HSV rainbow' to the Lab-blended `lipgloss.Blend1D` ROYGBIV description.

.claude/skills/update-changelog/SKILL.md
```diff
@@ -11,37 +11,46 @@ argument-hint: "[description or 'skip']"
...
+file in `changelog/unreleased/`; at release time (`/ship`) they're merged into
+`CHANGELOG.md` and deleted automatically. Full guide + voice: `changelog/README.md`.
 
 ### If argument is "skip", "no-changelog", or "none"
-Add the `no-changelog` label to the current PR and stop:
+Comment `/no-changelog` on the PR and stop:
 ```
-gh pr edit --add-label "no-changelog"
+gh pr comment --body "/no-changelog"
 ```
```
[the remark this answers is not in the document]
hayke102: Fixed — tagged the fence as ```bash (matches the repo's other shell examples). Note markdownlint isn't in CI here, so this was cosmetic-only, but happy to keep the doc lint-clean.

PR #184, opened 2026-07-09
feat(ui): re-viewable What's New, Shift+W label, Tab to full notes

hayke102:
## What & why

Three usability follow-ups on the What's New reel (shipped in v2.16.0/2.16.1).

### 1. What's New is re-viewable
**Problem:** opening the reel stamped "seen", and the reel filtered its *content* by that same seen version — so the first open showed highlights, but every re-open showed the empty "you're all caught up" state. Users saw it, exited, tried again, and lost it.

**Fix:** the reel now filters by the **7-day window only** (`isRecentHighlight`), independent of the seen version. The badge keeps owning the "unseen" signal via `isUnseenHighlight` (refactored to reuse `isRecentHighlight`). Opening still stamps seen (badge clears), but the content stays viewable. The dead `seenVersion` field is removed from the dialog and `ShowWhatsNew`.

### 2. `Shift+W` label
**Problem:** the badge said "press W"; users hit lowercase `w`, which opens the new-worktree dialog.

**Fix:** the badge, the help entry, and the `Ctrl+K` palette shortcut all spell it `Shift+W`. Key matching is unchanged (`handleKey` still switches on `msg.String()`).

### 3. `Tab` toggles reel ↔ full release notes
**Problem:** the reel and the full changelog were separate modes set only from `app.go`; no in-dialog way to move between them.

**Fix:** `Tab` (either direction) flips `whatsNew` and resets scroll; the footer hint is mode-aware (`tab all release notes` / `tab what's new`) and truncated so it can't wrap and break scroll math.

## Files
- `internal/ui/releasenotes.go` — window-only reel filter, `isRecentHighlight` helper, `ShowWhatsNew` signature, `Tab` case, mode-aware footer.
- `internal/ui/whatsnew.go` — badge hint → `Shift+W`.
- `internal/ui/keybindings.go`, `internal/ui/app.go` — help + palette shortcut label; updated call site.
- `internal/ui/whatsnew_test.go` — updated badge/reel tests, new `TestReleaseNotesTabToggle`.

## Tests
`go test -race ./internal/ui/ ./internal/releasenotes/ ./internal/config/` all pass. Unit tests assert the rendered `View()` for each flow: badge shows `Shift+W`; reel shows highlights regardless of seen; out-of-window → empty state; `Tab` flips title + content.

https://claude.ai/code/session_01JcZVtLAZLUpKgYRAEqNthR

## Summary by CodeRabbit

* **New Features**
  * “What’s New” now reopens to recent highlights instead of an “all caught up” view.
  * You can press **Tab** to switch between the highlights reel and the full release notes.
* **Bug Fixes**
  * The keyboard hint now displays **Shift+W** to better match the actual shortcut.
  * The “What’s New” action and related UI labels now use **Shift+W** consistently.

### DIFF
--- a/changelog/unreleased/whats-new-reviewable.md
+++ b/changelog/unreleased/whats-new-reviewable.md
@@ -1,4 +1,4 @@
 ---
 type: improved
 ---
-**What's New is re-viewable now.** Reopening it shows the recent highlights again instead of "you're all caught up", the badge spells the shortcut `Shift+W` so it's not mistaken for lowercase `w` (new worktree), and `Tab` flips between the highlights reel and the full release notes.
+**Re-viewable now.** Reopening What's New shows the recent highlights again instead of "you're all caught up". The badge spells the key `Shift+W` (so it's not read as lowercase `w`), and `Tab` flips between the highlights reel and the full release notes.
--- a/changelog/unreleased/whats-new-reviewable.md
+++ b/changelog/unreleased/whats-new-reviewable.md
@@ -1,0 +1,4 @@
+---
+type: improved
+---
+**What's New is re-viewable now.** Reopening it shows the recent highlights again instead of "you're all caught up", the badge spells the shortcut `Shift+W` so it's not mistaken for lowercase `w` (new worktree), and `Tab` flips between the highlights reel and the full release notes.
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -6545,8 +6545,9 @@ func (h *Home) newestKnownVersion() string {
 }
 
 // recomputeWhatsNew refreshes hasUnseenWhatsNew from the cached releases and the
-// stored seen version. Shares isUnseenHighlight with the dialog so the badge and
-// the reel always agree on what counts.
+// stored seen version, using isUnseenHighlight — the seen-aware, badge-only
+// predicate. The reel filters by isRecentHighlight (window only) instead, so the
+// badge clears once seen while the reel stays re-viewable.
 func (h *Home) recomputeWhatsNew() {
 	seen := releasenotes.NormalizeVersion(h.cfg.GetReleaseNotesSeenVersion())
 	h.hasUnseenWhatsNew = false
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -6218,7 +6218,7 @@ func (h *Home) buildPaletteItems() []PaletteItem {
 		{Kind: PaletteKindCommand, ID: "settings", Name: "Settings", Shortcut: "S"},
 		{Kind: PaletteKindCommand, ID: "bug_report", Name: "Bug Report", Shortcut: "!"},
 		{Kind: PaletteKindCommand, ID: "help", Name: "Help", Shortcut: "?"},
-		{Kind: PaletteKindCommand, ID: "whats_new", Name: "What's New", Shortcut: "W"},
+		{Kind: PaletteKindCommand, ID: "whats_new", Name: "What's New", Shortcut: "Shift+W"},
 		{Kind: PaletteKindCommand, ID: "release_notes", Name: "Release Notes"},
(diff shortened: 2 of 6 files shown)


internal/ui/app.go
```diff
@@ -6523,7 +6523,7 @@ func (h *Home) loadReleaseNotes() tea.Cmd {
 func (h *Home) openReleaseNotes(whatsNew bool) (tea.Model, tea.Cmd) {
 	h.actionLog.Add("open release notes", "", true)
 	if whatsNew {
-		h.releaseNotes.ShowWhatsNew(h.version, h.cfg.GetReleaseNotesSeenVersion())
+		h.releaseNotes.ShowWhatsNew(h.version)
```
hayke102, line 6526: Stale comment introduced by this change (the real location is `recomputeWhatsNew` at app.go:6548, just below — outside the diff hunk, hence anchored here). Its doc comment still reads:

> Shares `isUnseenHighlight` with the dialog so the badge and the reel always agree on what counts.

That invariant is exactly what this PR removes: the reel now filters via `isRecentHighlight` (window only) while `recomputeWhatsNew`/the badge keep `isUnseenHighlight` (seen-version-aware). The whole point of re-viewability is that the badge clears while the reel still shows — so they intentionally *disagree* now. Suggest rewording to say the predicate is shared only with the **badge**, not the reel.

changelog/unreleased/whats-new-reviewable.md
```diff
@@ -0,0 +1,4 @@
+---
+type: improved
+---
+**What's New is re-viewable now.** Reopening it shows the recent highlights again instead of "you're all caught up", the badge spells the shortcut `Shift+W` so it's not mistaken for lowercase `w` (new worktree), and `Tab` flips between the highlights reel and the full release notes.
```
hayke102: Minor / non-blocking, per the changelog voice rules in `CLAUDE.md`: *"open each bullet with a **bold 2–4 word headline**"* and *"one idea per bullet"*. The headline **"What's New is re-viewable now."** is 5 words, and the bullet packs three changes (re-viewable reel + `Shift+W` relabel + `Tab` toggle) into one sentence. It renders in-app (Ctrl+K → Release Notes), so a tighter lead scans better — e.g. **"Re-viewable now."** as the headline, trimming or splitting the `Shift+W`/`Tab` details.

hayke102 (commented):
Reviewed the diff — the code change itself is correct and well-tested (the `content[scroll:scroll+visible]` slice stays in bounds, the `Tab` toggle resets scroll to a valid 0, `case "W"` is still the real key dispatch so the `Shift+W` relabel is display-only, and the `IsFirstRun() && seen==""` gate is the right fix for auto-updated users). Two small notes inline.

One more that can't be an inline thread since the file isn't in this diff:

**`CLAUDE.md:95`** still documents the pre-PR behavior. It reads *"...for releases newer than the seen version and within `whatsNewWindowDays` (7) — shared predicate `isUnseenHighlight` keeps the badge and the reel in agreement"*, and the item header still shows *"(`W` / `Ctrl+K` → …)"*. After this PR the reel filters by window only (`isRecentHighlight`), the badge alone honors the seen version, and the binding is `Shift+W`. Worth updating here since it's this feature's own doc.

internal/ui/app.go
hayke102: Fixed — reworded the recomputeWhatsNew doc comment: isUnseenHighlight is now described as the seen-aware, badge-only predicate, and the reel's window-only isRecentHighlight is called out as the intentional divergence.

changelog/unreleased/whats-new-reviewable.md
hayke102: Fixed — tightened the fragment: 2-word bold headline ("Re-viewable now."), main idea first, with the Shift+W/Tab details trailing in a second sentence.

PR #194, opened 2026-07-12
feat(ui): `.` context menu for the selected sidebar row

hayke102:
Press `.` on any sidebar row to open a dropdown listing **exactly the actions that apply to it**. Sessions, worktrees, repos, and origins each get their own list.

```
╭──────────────────────────────────────╮   ╭───────────────────────────────╮
│ session: context-menu                │   │ worktree: fleet-wt            │
│ ▸ Attach                           ⏎ │   │ ▸ Collapse                  ⏎ │
│   Focus Preview                    ⇥ │   │   New Session               a │
│   Quick Approve — not waiting      Y │   │   New Session (Pick Agent)  A │
│   Restart                          r │   │   New Worktree Session      w │
│   Rename                           R │   │   Switch Branch             b │
│   Mark as Unread — hasn't run yet  m │   │   Open PR — no PR           p │
│   Open in Editor                   e │   │   Remove Worktree           d │
│   Open PR — no PR                  p │   ╰───────────────────────────────╯
│   Fork Session                     f │
│   Fork to Worktree — Claude only   F │   ╭───────────────────────────╮
│   Suspend Session                    │   │ origin: brizzai/fleet     │
│   New Worktree Session             w │   │ ▸ Collapse              ⏎ │
│   Delete Session                   d │   │   New Worktree Session  w │
╰──────────────────────────────────────╯   │   Forget Origin Group   d │
                                           ╰───────────────────────────╯
```

## Why

fleet's actions are row-sensitive: `d` deletes a session, removes a worktree, forgets a repo, or forgets a whole origin group depending on where the cursor is. Until now the only way to know what applied was the footer bar (a hand-curated 5–6 key subset) or `?` (every key in the app, regardless of what's selected). Nothing showed *the actions valid for this row, right now*.

It also gives a home to actions that had no key at all — Suspend This Session was palette-only.

## Design: a view over commands that already exist

`dispatchCommand(id)` was already a central action-ID → handler switch driving the command palette, and `cursorBarContext()` already classified the row under the cursor. So the menu **adds no handler logic**: each entry carries a `dispatchCommand` id and inherits handlers, action-log entries, and analytics for free.

Three ids were added:

- **`delete_at_cursor`** → `deleteAtCursor()`, the scope-follows-cursor delete **extracted from the `d` key** so both share one implementation and can't drift.
- **`toggle_group`** → expand/collapse a header.
- **`resume`** — a suspended session's tmux is gone, so `attach` would simply fail. It has to be restarted with `--resume`. (`dispatchCommand("attach")` calls `attachSelected()`, which does not resume — routing a suspended row through it would have been a latent bug.)

## Behavior

- **Dimmed, not hidden.** Actions that exist but can't run right now stay visible, greyed, with the reason (`Fork to Worktree — Claude only`, `Mark as Unread — hasn't run yet`), and are skipped by `j`/`k`. The guards **mirror the real handlers exactly**, so a lit row can never dead-click.
- **The menu teaches the keybinding.** Every row shows its real key, and pressing that key inside the menu fires it — so the menu trains you onto the faster path rather than competing with it.
- **The title names the kind.** `worktree: fleet-wt` vs `repo: fleet` — derived from the same branch that picks the delete label, so the title can't call a row a repo while offering to remove a worktree.
- **Anchored, not modal.** Hangs below its row, flips above near the footer, and deliberately skips `dimBackdrop` — it's a small box beside the row it acts on, not a takeover.

## Notable review points

- **The footer bar changed.** It doesn't truncate, so a 7th key would bleed past a narrow terminal. Each context stays at 6 keys by dropping the one the menu now cheaply replaces (`r Restart` in attach mode, `p PR` in split mode). Easy to revert if you'd rather keep those.
- **`TestContextMenuIDsAreDispatchable`** parses `dispatchCommand`'s case literals via `go/ast` and fails if a menu id isn't handled. `dispatchCommand` **no-ops silently on unknown ids**, so a rename would otherwise turn a menu entry into a dead click with no error anywhere. Verified the guard actually bites by breaking an id.
- **Quick Approve's enabled path isn't unit-tested** — the guard mirrors `quickApproveSelected`, which requires `IsAlive()` (a live tmux pane). Covered the negative cases and said so in a comment rather than weakening the guard to make it testable.

## Verification

`make build`, `make test` (race), and `make lint` all clean. Beyond unit tests, `TestContextMenuEndToEndOpensAndDispatches` drives the real path: `.` → `handleKey` → `routeToModal` → `contextMenuMsg` → `dispatchCommand` → the actual confirm dialog.

https://claude.ai/code/session_01AvyHmniRXX8K3VLk92fgJX

## Summary by CodeRabbit

* **New Features**
  * Added a cursor-anchored sidebar context menu (open with `.`) with entity-specific actions for sessions, worktrees, repositories, and origins.
  * Unavailable actions stay visible but dimmed with an explanation; keyboard shortcuts are shown in the menu.
  * Includes navigation, scrolling, and auto-positioning (and silently does nothing when there’s no available action).
* **Documentation**
  * Updated footer key hints to advertise the `.` menu action by context.
* **Tests**
  * Added coverage for menu contents, enabled/disabled states, dispatch behavior, and layout/keyboard interactions (open/close, navigation, activation, positioning).


internal/ui/app.go
```diff
@@ -2813,6 +2837,24 @@ func (h *Home) handleSessionCreateResult(msg sessionCreateResultMsg) (tea.Model,
 	return h, tea.Batch(h.probeTCCCmd(s.ProjectPath), h.fetchPreviewForSelected())
 }
 
+// deleteAtCursor runs the delete whose scope follows the cursor: an origin
+// header forgets the whole group, any other header acts on the container
+// (§ confirmDeleteHeader), and a session row deletes just that session. Shared by
+// the `d` key and the context menu's delete entry so the two can't drift.
+func (h *Home) deleteAtCursor() tea.Cmd {
```
hayke102, line 2857: **The menu holds no row identity — an async rebuild can move the cursor out from under it.**

Every entry re-resolves `h.cursor` at dispatch time (`deleteAtCursor`, `selectedSession`, `resolveCurrentRepo`), and the CLAUDE.md note argues the cursor can't move because `routeToModal` swallows every key while the menu is open. That's true for keys, but not for messages: `handleSessionCreateResult` **explicitly** moves the cursor (`h.cursor = i`, the auto-select loop just above this func), and `rebuildFlatItems()` also runs on `tickMsg` and `hookChangedMsg`.

Repro: kick off a worktree session (creation is non-blocking by design), press `.` on session X while it's still spinning, then let `sessionCreateResultMsg` land. The menu is still open and still titled `session: X`, but the cursor is now on the new session — press `d` and you get the delete confirm for the wrong session.

Fix would be to capture the target at `Show()` (session ID, or repo path + row kind) and have the menu's dispatch re-resolve the cursor from it — or at minimum close the menu whenever `rebuildFlatItems` moves the cursor.

internal/ui/app.go
```diff
@@ -6198,6 +6260,161 @@ func (h *Home) setInfo(msg string) {
...
+	enterKey, tabKey := "⏎", "⇥"
+	if h.cfg.GetEnterMode() == "split" {
+		enterKey, tabKey = "⇥", "⏎"
+	}
+
+	// A suspended session has no tmux to attach to — it has to be restarted with
+	// --resume first, so the primary action changes shape.
+	open := ContextMenuItem{ID: "attach", Label: "Attach", Shortcut: enterKey, Enabled: true}
+	if status == session.StatusSuspended {
+		open = ContextMenuItem{ID: "resume", Label: "Resume Session", Shortcut: enterKey, Enabled: true}
```
hayke102: **In `split` enter-mode this labels Resume with `⇥`, but Tab doesn't resume.**

The Enter↔Tab swap above is right for `attach`, but `handleKey`'s suspended-session branch lives under `case "enter"` only — it runs *before* the split check, so Enter resumes in **both** modes. `case "tab"` in split mode goes straight to `attachSelected()`, which can't work on a suspended session (its tmux is gone).

So in split mode the menu teaches a key that fails. `resume` should keep `⏎` regardless of enter mode.

internal/ui/app.go
```diff
@@ -6198,6 +6260,161 @@ func (h *Home) setInfo(msg string) {
...
+		unread.Enabled = true
+	}
+
+	items := []ContextMenuItem{
+		open,
+		{ID: "focus", Label: "Focus Preview", Shortcut: tabKey, Key: "tab", Enabled: true},
+		{
+			ID: "approve", Label: "Quick Approve", Shortcut: "Y", Key: "Y",
+			Enabled: s.IsAlive() && status == session.StatusWaiting,
+			Note:    "not waiting",
```
hayke102: **"not waiting" is the wrong reason when the session *is* waiting but its pane is dead.**

The guard is a conjunction (`s.IsAlive() && status == StatusWaiting`) but the note is a constant. A waiting session whose pane died renders `Quick Approve — not waiting` right next to a `◐ waiting` dot in the sidebar — the note contradicts the visible status.

(This is exactly the state the new `"approve disabled when the pane is dead"` test constructs, it just doesn't assert on the note.) Same shape as `mark_unread`, which already switches on the reason — worth doing the same here.

internal/ui/app.go
```diff
@@ -6198,6 +6260,161 @@ func (h *Home) setInfo(msg string) {
...
+		},
+		{
+			ID: "fork", Label: "Fork Session", Shortcut: "f", Key: "f",
+			Enabled: resumable,
+			Note:    "no session id yet",
+		},
+		{
+			ID: "fork_worktree", Label: "Fork to Worktree", Shortcut: "F", Key: "F",
+			Enabled: s.Agent == agent.Claude && resumable,
+			Note:    "Claude only",
```
hayke102: **"Claude only" is a flatly wrong reason for a Claude session that just hasn't captured a resume id yet.**

`Enabled: s.Agent == agent.Claude && resumable` — two reasons, one note. A freshly started Claude session (`ClaudeSessionID` still empty) renders `Fork to Worktree — Claude only`, while the `fork` row directly above it correctly says `no session id yet` for the identical state.

Since this exact string is the example in the changelog fragment and the CLAUDE.md entry, it's worth making it honest: switch the note the way `mark_unread` does.

internal/ui/context_menu.go
```diff
@@ -0,0 +1,260 @@
...
+			return j
+		}
+	}
+	return i
+}
+
+// maxVisible is how many rows fit between the menu's top and the footer, after
+// the box chrome (2 border rows + title + hint). Keeps the dropdown on screen on
+// a short terminal; the overflow scrolls.
+func (d *ContextMenuDialog) maxVisible() int {
```
hayke102, line 124: **The chrome budget doesn't match what `View` renders.**

The comment budgets 4 rows for "2 border rows + title + hint", but `View` renders no hint line — it renders the title plus **up to two `⋮` indicator rows**, which aren't budgeted at all. So once the menu overflows, the box is taller than `maxVisible` assumed.

On a short terminal (say `height=12`, footer 1 → `bottomLimit=11`): `maxVisible` = 7, and scrolled to the middle the box renders 7 rows + title + 2 `⋮` + 2 borders = 12 rows against an 11-row budget. `Position` then flips and clamps `y` to 0 and the last row lands on the footer.

Budget 3 chrome rows + the `⋮` rows that will actually render (or subtract 2 unconditionally and accept one wasted row).

internal/ui/app.go
```diff
@@ -6405,6 +6622,20 @@ func (h *Home) dispatchCommand(id string) (tea.Model, tea.Cmd) {
 			h.actionLog.Add("delete session", s.Title, true)
 		}
 		return h, h.confirmDeleteSelected()
+	case "delete_at_cursor":
```
hayke102: Minor: `case "delete"` (a few lines above, the palette's Delete Session) is now line-for-line the non-header half of `deleteAtCursor()`. Two ids mean "delete" and one of them is scope-blind, so a future guard change to `deleteAtCursor` silently won't reach the palette entry. Could the palette's `delete` just route here too?

changelog/unreleased/context-menu.md
```diff
@@ -0,0 +1,5 @@
+---
+type: added
+highlight: true
+---
+**Press `.` for the menu.** A context menu now opens on the row under your cursor, listing exactly the actions that apply to it — a session, a worktree, a repo, and an origin each get their own list. Actions that exist but can't run right now stay visible and dimmed with the reason (`Fork to Worktree — Claude only`), and every entry shows its real keybinding, so you can keep pressing `.` or graduate to `d`/`Y`/`r` whenever you like.
```
hayke102: Nit, but it's our own rule: CLAUDE.md says *"Keep fragments concise: 1–2 sentences, lead with the user-facing change. No implementation detail, marketing tone, or exhaustive sub-points."* This is three sentences and ~70 words, enumerating per-row-kind lists, dimmed-with-reason, and the keybinding display. With `highlight: true` it also leads the release in the in-app What's New reel, where it'll be by far the longest bullet.

internal/ui/app.go
```diff
@@ -6198,6 +6260,161 @@ func (h *Home) setInfo(msg string) {
...
+		{
+			ID: "suspend_session", Label: "Suspend Session",
+			Enabled: (status == session.StatusIdle || status == session.StatusFinished) && resumable,
+			Note:    "idle or finished only",
+		},
+		{ID: "new_worktree", Label: "New Worktree Session", Shortcut: "w", Key: "w", Enabled: true},
+		{ID: "delete_at_cursor", Label: "Delete Session", Shortcut: "d", Key: "d", Enabled: true},
+	}
+	return "session: " + s.Title, items
+}
```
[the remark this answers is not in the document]
hayke102: Fixed — good catch, and confirmed against `handleKey`: the `StatusSuspended` check sits inside `case "enter"` *above* the split-mode branch (its own comment says "Overrides split mode"), so Enter resumes in both modes. `resume`'s Shortcut is now a literal `⏎` rather than `enterKey`, with a comment explaining why it can't follow the swap.

Took you up on the regression test too: `TestContextMenuResumeShortcutIsEnterInBothModes` asserts the shortcut under both `EnterMode: "attach"` and `"split"`, and also sanity-checks that `attach` *does* still follow the swap.

internal/ui/app.go
hayke102: Fixed — same issue CodeRabbit flagged independently, so it's now covered from both angles. `resume` is pinned to a literal `⏎` regardless of enter mode; `TestContextMenuResumeShortcutIsEnterInBothModes` locks it in.

internal/ui/app.go
hayke102: Fixed — you're right, and the CLAUDE.md sentence was simply wrong. "routeToModal swallows every key" is true for *keys* and says nothing about *messages*: `handleSessionCreate` (app.go:2829) and the workspace-create handler (app.go:1163) both assign `h.cursor` from async handlers, and `rebuildFlatItems` runs on the tick.

Went with capturing the row identity rather than closing the menu, since it's strictly stronger — the action then always hits the row the title *named*:
- `contextMenuTarget` (session ID / repo path / origin key) captured by `.` via `targetForCursor()`
- the `contextMenuMsg` handler calls `focusContextMenuTarget()` to re-pin the cursor before `dispatchCommand` runs, and refuses with an info message if the row is gone

Two tests: `TestContextMenuDispatchFollowsTargetNotCursor` moves the cursor under an open menu and asserts the delete confirm still carries *alpha*'s id (it fails with your exact symptom when reverted — confirms deleting beta), and `TestContextMenuRefusesWhenTargetIsGone` covers the row disappearing. CLAUDE.md now documents the real invariant.

internal/ui/app.go
hayke102: Fixed — `approve` now switches on the failing clause like `mark_unread` does: `not waiting` when the status isn't Waiting, `session not running" when it *is* waiting but the pane is dead. You're right that a note contradicting the `◐ waiting` dot beside it is worse than no note.

Also swept the same defect class while I was in there: `suspend_session` had the identical bug (a *finished* session with no resume id read `idle or finished only`, which is false) — it now distinguishes `already suspended` / `idle or finished only` / `no session id yet`. `TestContextMenuNotesNameTheRealReason` now asserts the note text for every conjunction guard, which is exactly the assertion the old test was missing.

internal/ui/app.go
hayke102: Fixed — and it was worse than one wrong string: the `fork` row directly above it already said `no session id yet` for the identical state, so the two adjacent rows disagreed. `fork_worktree` now switches: `Claude only` for a non-Claude agent, `no session id yet` for a Claude session that hasn't captured one.

Since this string was the example in both the changelog fragment and the CLAUDE.md entry, I replaced the CLAUDE.md example with the honest version and made the rule explicit there ("every guard is a conjunction, so the note must switch on which clause failed").

internal/ui/context_menu.go
hayke102: Fixed — right on both counts. The comment described a hint line I never wrote, and the two `⋮` rows weren't budgeted. Chrome is now a named `contextMenuChrome = 5` (2 border + title + up to 2 `⋮`), taking the conservative bound you suggested — one wasted row on an unscrolled menu, never an overrun.

`TestContextMenuFitsShortTerminal` renders a 14-item menu scrolled to the middle (so both indicators show) at heights 10/12/16/24 and asserts the box never exceeds its budget. Reverting the constant to 4 reproduces your numbers exactly: "height=12: box is 12 rows but only 11 are available".

internal/ui/app.go
hayke102: Fixed — merged, and renamed the palette entry to match. `case "delete", "delete_at_cursor"` now both route to `deleteAtCursor()`, so there's a single scope-follows-cursor delete behind `d`, the palette, and `.`.

Worth flagging that this is a small behavior change to an existing feature: the palette's delete previously *silently no-opped* on a repo/origin header (`confirmDeleteSelected` returns nil on `IsRepoHeader`). It now acts on the header. Since the label would then be lying, I renamed it `Delete Session` → `Delete (session / repo / worktree)`.

changelog/unreleased/context-menu.md
hayke102: Fixed — fair, it's our own documented rule and `highlight: true` made the over-length copy the first thing in the reel. Cut to two sentences, dropped the per-row-kind enumeration and the keybinding-display detail, kept the bold headline.

PR #196, opened 2026-07-12
feat(analytics): send telemetry to PostHog

hayke102:
Moves fleet's telemetry off its custom self-hosted endpoint (Next.js + Neon + Vercel) onto **PostHog**, and fixes an unrelated hooks bug that was already on the branch.

## Why

The custom dashboard worked, but every chart was hand-rolled SQL — retention cohorts and error tracking especially. PostHog reproduces all of it natively, and unlocks funnels/flags/experiments we'd otherwise have to build.

## What changed

**`feat(analytics): send telemetry to PostHog`**

The transport swaps to the `posthog-go` SDK. The public surface (`Init`/`Track`/`Gauge`/`Distribution`/`SetUserProperties`/`Heartbeat`/`Shutdown`) is **unchanged**, so none of the ~50 call sites move — only the internals of `internal/analytics`.

- The SDK owns batching/flushing → the hand-rolled queue, worker goroutine, and HTTP POST are deleted.
- A `sink` interface seam keeps mode-gating testable with **no network round-trip**.
- The SDK's logger is routed into `debuglog` — left on stdout it would corrupt the Bubble Tea display.
- The PII key sanitizer and the on-disk pending queue are untouched.

**Telemetry modes behave exactly as before:**

| mode | behaviour |
|---|---|
| `full` | identified, person profile, full event set |
| `minimal` | anonymous DAU only — `$process_person_profile=false`, so the device **still counts as a unique user** but creates no person profile and never leaks the git email |
| `off` | disabled client, nothing sent |

`FLEET_TELEMETRY_DISABLED` / `DO_NOT_TRACK` still force off.

**`fix(hooks): don't bake a `go run` binary path into Claude's hooks`**

`make run` used `go run`, which builds to a temp path and deletes it on exit. fleet persists its own executable path into `~/.claude/settings.json`, so that dead path stuck around and every later hook failed with *"No such file or directory"* — session status silently stopped updating. `make run` now builds and execs a real binary, and the installer refuses to write a `go run` path.

## On the embedded key

The PostHog **project** API key is write-only (capture-only, never reads), so it's embedded like the ingest token it replaces. `FLEET_POSTHOG_KEY` / `FLEET_POSTHOG_HOST` override it, and `Init` runs **disabled** when no key is configured. The secret *personal* API key is not in the binary.

## Verification

Not just compiled — driven end-to-end against the live PostHog project with the real client:

- ✅ anonymous (minimal) devices count as **unique users** with **no person profile**
- ✅ `Track()` no-ops in minimal mode (only the DAU heartbeat sends)
- ✅ minimal never leaks the git email — distinct_id stays the device hash even when `GitEmail` is set
- ✅ the `mode` property drives the consent / decline-rate breakdown
- ✅ a deliberately-passed `repo_name` was stripped before leaving the machine

`make build` clean, `go test -race ./...` green.

## Dashboards

Four dashboards are already built in PostHog: Overview, Engagement & Retention, Features & Agents, Health & Reliability.

## Follow-ups (not in this PR)

- Decommission the old Vercel/Neon stack
- Route `error_occurred` through `posthog.Exception` for real stack grouping
- Wire feature flags to unlock experiments

https://claude.ai/code/session_012tuMH7Wg16rYPM5mG3tP9M

## Summary by CodeRabbit

* **New Features**
  * Telemetry analytics now use PostHog while preserving existing telemetry modes and opt-out behavior.
  * Minimal telemetry remains anonymous and device-based; full telemetry supports user identification.
  * PostHog configuration can be customized through environment variables.

* **Bug Fixes**
  * Hook execution now remains reliable when launching the application with `go run`.
  * `make run` builds and runs a stable development binary, preventing temporary paths from being saved in hook configuration.

* **Documentation**
  * Updated telemetry guidance and added unreleased changelog entries for the analytics and development-run changes.


internal/analytics/analytics.go
```diff
@@ -231,48 +243,73 @@ func readGitIdentity() (name, email string) {
...
-	for item := range c.queue {
-		if err := c.post(item); err != nil {
-			debuglog.Logger.Debug("analytics post failed", "err", err)
-		}
-	}
+// phLogger routes the SDK's internal logging into fleet's debug log. Without
+// it the SDK would write to stdout/stderr and corrupt the Bubble Tea display.
+// Everything is logged at debug level — analytics is best-effort, so its noise
+// should never surface as a fleet error.
+type phLogger struct{}
```
hayke102, line 279: **This leaks the user's git email into `debug.log`, which the bug-report flow republishes into public GitHub issues.**

`phLogger.Debugf` is wired into the SDK, and posthog-go's batch loop debug-logs every prepared message:

```go
// posthog.go:1723
c.debugf("buffer (%d/%d, %d bytes) %v", len(batchData), c.BatchSize, batchSize, prepared.msg)
```

`prepared.msg` is the `APIMessage` — it carries `distinct_id`, which in `full` mode **is the git email**, plus every event property. So with `FLEET_DEBUG=1` (what `make run` sets, and what we ask reporters to run with) that lands in `~/.config/fleet/debug.log`. Then `!` → `g` embeds the last 100 lines of debug.log into the pre-filled `gh issue create --web` body → email published in a public issue.

The client this replaces was explicitly careful here — it had a comment saying *"Logging distinctID would leak email; logging deviceID is safe"* and never logged payloads at all. That guarantee is gone.

Simplest fix: make `Debugf`/`Logf` no-ops (or drop them behind a `FLEET_POSTHOG_VERBOSE` env gate). We get nothing from the SDK's per-batch buffer dump.

internal/hooks/claude_hooks.go
```diff
@@ -79,11 +80,40 @@ func fleetBinaryPath() string {
 		return "fleet"
 	}
 	if resolved, err := filepath.EvalSymlinks(exe); err == nil {
-		return resolved
+		exe = resolved
+	}
+	// Hooks outlive the process that writes them, so the path has to survive
+	// our exit. `go run` deletes its binary as soon as we quit; resolve to an
+	// installed fleet instead of leaving hooks pointing at a deleted file.
+	if isGoRunBinary(exe) {
```
hayke102, line 94: The fix is right, but it stops one file short: **`internal/chrome/install.go:74` still has the same dangling-`go run`-path bug.**

`InstallNativeMessagingHost` resolves the host binary with raw `os.Executable()` + `EvalSymlinks` — and its comment currently claims it's the *"same pattern as `hooks.GetHookCommand`"*, which stops being true with this PR. Launch via `go run ./cmd/fleet` (still possible even though `make run` no longer does it) and the NMH manifest gets written with the `<tmp>/go-build*/b001/exe/fleet` path that Go deletes on exit. Chrome then can't spawn the native host, so `p` silently loses tab reuse for the rest of that session.

Worth exporting `fleetBinaryPath()` (or moving it somewhere shared) and having `install.go` call it, so the two can't drift again.

internal/analytics/analytics.go
```diff
@@ -231,48 +243,73 @@ func readGitIdentity() (name, email string) {
...
 	}
-	req.Header.Set("Content-Type", "application/json")
-	req.Header.Set("x-fleet-token", ingestToken)
-	resp, err := c.httpClient.Do(req)
-	if err != nil {
-		return err
+	if c.mode == ModeMinimal {
+		props.Set("$process_person_profile", false)
+	}
+	if err := c.ph.Enqueue(posthog.Capture{
```
hayke102: `posthog.Enqueue` is a **blocking channel send** (`c.msgs <- prepared`, posthog.go:568 — no `default` case), and it serializes the message inline on the caller's goroutine.

The client this replaces had an explicit non-blocking drop:

```go
select {
case c.queue <- item:
default: // drop
}
```

That `default` is what made `Track` safe to call from inside Bubble Tea's `Update()` — and we do, from key handlers (`app.go:750`, `932`, `2163`, `settings.go:729`, …). Now the non-blocking property is no longer a guarantee, it's a function of the SDK's 2500-slot queue depth: if PostHog is unreachable and the batch loop stalls in `sendBatch` long enough to fill the queue, a keypress blocks and the TUI freezes.

Cheap fix: keep a small buffered channel + `select/default` in front of `Enqueue`, and drain it from a goroutine. Restores the by-construction guarantee that `Update()` can't block.

internal/analytics/analytics.go
```diff
@@ -378,28 +413,21 @@ func SyncEnabled(mode string, version string, identity Identity) {
...
-	go func() {
-		c.wg.Wait()
-		close(done)
-	}()
-	select {
-	case <-done:
-	case <-time.After(shutdownTimeout):
-		debuglog.Logger.Info("analytics shutdown timed out, dropping queued events")
+	// posthog Close respects Config.ShutdownTimeout, so this can't hang quit.
+	if err := c.ph.Close(); err != nil {
```
hayke102: The comment says this "can't hang quit" — true, it's capped at 500ms. But it's still **a blocking network flush called from inside `Update()`**, and not only on quit:

- quit: `app.go:2473`
- **every Settings-dialog close**: `app.go:953` → `settingsClosedMsg` → `SyncEnabled` → `Shutdown` → `ph.Close()`

`Close()` drains and synchronously POSTs the pending batch. With the new 5s `Interval` there's *virtually always* a pending batch, so changing a setting and pressing `esc` can freeze the TUI for up to 500ms on a slow network. Under the old design the worker had already POSTed each event as it arrived, so `Shutdown`'s `wg.Wait()` returned instantly.

Easiest: run `Shutdown()` off the Update goroutine in `SyncEnabled` (fire-and-forget `go`), keeping the synchronous path only for quit.

internal/analytics/analytics.go
```diff
@@ -40,55 +36,73 @@ const (
...
-	// shutdownTimeout caps how long Shutdown blocks waiting for the worker to
-	// drain the queue. Kept short so quitting fleet exits promptly; on a slow
-	// or unreachable network any events still queued past this are dropped.
+	// defaultPostHogHost is fleet's PostHog ingest host — the project lives in
+	// the US region. Overridable via FLEET_POSTHOG_HOST for local testing.
+	defaultPostHogHost = "https://us.i.posthog.com"
+
+	// flushInterval is how often the SDK ships a batch. fleet emits few events;
+	// a short interval keeps dashboards fresh and Close() quick on quit.
+	flushInterval = 5 * time.Second
```
hayke102: These two constants are in tension: **batch every 5s, but give the flush only 500ms on quit** (`shutdownTimeout`, just below).

The old client POSTed each event the moment it was enqueued, so quit had nothing to drain. Now a user who launches fleet and quits three seconds later has `app_started` (and the `app_active` heartbeat) still sitting in the in-memory batch — the 5s tick never fired. `Close()` then has to cover DNS + a cold TLS handshake to `us.i.posthog.com` + the POST inside 500ms; when that deadline blows, `c.cancel()` aborts the in-flight request and the events are simply dropped.

That silently undercounts exactly the launch-and-quit population the minimal-mode anonymous ping exists to measure. Either shorten `Interval` (1s) or give `ShutdownTimeout` a more realistic budget (2s) — the comment above claims the short interval keeps `Close()` quick, but it's the interval being *long* relative to session length that creates the backlog.

internal/hooks/claude_hooks.go
```diff
@@ -79,11 +80,40 @@ func fleetBinaryPath() string {
...
+		debuglog.Logger.Warn("hooks: launched via go run, resolving hook command from PATH", "exe", exe)
+		if installed, err := exec.LookPath("fleet"); err == nil {
+			return installed
+		}
+		return "fleet"
```
hayke102: This bare `"fleet"` literal gets written into `~/.claude/settings.json` (via `GetHookCommand`) **and** baked into the OpenCode plugin as `FLEET_BIN`, where it's used with `spawnSync` (`opencode_hooks.go:114`).

For a contributor who builds from source and has no installed `fleet` on `PATH`, that means every hook event fails with ENOENT/command-not-found, silently — session status just stops updating, which is the same class of breakage this PR is fixing, relocated.

At the point we reach this line, `exe` is a real path that *works right now* — it only dies later. Returning `exe` would at least keep the current session's hooks working. Better still: don't write hooks at all and log a `Warn`, so it fails loudly instead of leaving a poisoned settings.json behind.

internal/hooks/claude_hooks.go
```diff
@@ -79,11 +80,40 @@ func fleetBinaryPath() string {
...
+	dir := filepath.Dir(path)
+	if filepath.Base(dir) != "exe" {
+		return false
+	}
+	for {
+		parent := filepath.Dir(dir)
+		if parent == dir {
+			return false
+		}
+		if strings.HasPrefix(filepath.Base(dir), "go-build") {
```
hayke102: This walks every ancestor up to the filesystem root, which matches a lot more than the shape the doc comment describes.

The `go run` layout is fixed-depth — `<tmp>/go-build<rand>/b001/exe/<name>` — so any binary with an ancestor starting with `go-build` and a parent dir named `exe` (`~/go-build-tools/dist/exe/fleet`) is misclassified, and we then throw away a perfectly good absolute path in favour of `LookPath` (or the bare literal above).

An exact check is both simpler and correct:

```go
exeDir := filepath.Dir(path)              // .../b001/exe
bDir := filepath.Dir(exeDir)              // .../b001
goBuild := filepath.Dir(bDir)             // .../go-build<rand>
return filepath.Base(exeDir) == "exe" &&
	strings.HasPrefix(filepath.Base(goBuild), "go-build")
```

internal/hooks/claude_hooks.go
```diff
@@ -79,11 +80,40 @@ func fleetBinaryPath() string {
...
+			return true
+		}
+		dir = parent
+	}
+}
+
 // GetHookCommand returns the full hook command string using the current binary path.
 func GetHookCommand() string {
 	return fleetBinaryPath() + " hook-handler " + fleetHookArg
```
hayke102: The path is concatenated unquoted, and this PR makes that newly reachable: `make run` now bakes the **developer's repo path** into the hook command, where the old `go run` temp path (`/var/folders/...`) never contained spaces.

Clone into `~/My Projects/fleet` and we write:

```
/Users/x/My Projects/fleet/build/fleet hook-handler --fleet-hook
```

Claude Code runs hooks through a shell, which splits on the space and tries to exec `/Users/x/My`. Every hook fails silently → hook-based status detection (the authoritative source) stops working, all sessions stick on stale statuses, nothing surfaces in the TUI. Same unquoted path goes into the OpenCode plugin.

Shell-quote the binary path here.

internal/analytics/analytics.go
```diff
@@ -153,9 +153,20 @@ func Init(mode string, version string, identity Identity) {
...
 
-	if mode == ModeOff || isOptedOut() {
+	key := projectAPIKey
+	if v := strings.TrimSpace(os.Getenv("FLEET_POSTHOG_KEY")); v != "" {
+		key = v
+	}
+
+	// Disabled when the user opted out, or before the PostHog project key is
+	// configured — the latter keeps this file safe to land ahead of the account.
+	if mode == ModeOff || isOptedOut() || key == "" {
```
hayke102, line 181: The `key == ""` arm is now dead: `projectAPIKey` ships populated, and the env override at :157 is non-empty-guarded (`if v := strings.TrimSpace(...); v != ""`), so `key` can never be empty in production. The only thing keeping this branch alive is `gating_test.go` mutating the package var.

Ironically it doesn't even do what the comment implies for the one input that could reach it — `FLEET_POSTHOG_KEY=" "` trims to `""` and is *ignored* rather than disabling the client.

Also: `CLAUDE.md:140` still claims the client "stays disabled until a project key is configured", which is no longer true now that the key is embedded. Worth dropping the guard and fixing that line in the same pass.

internal/analytics/analytics.go
```diff
@@ -231,48 +243,73 @@ func readGitIdentity() (name, email string) {
...
+func (phLogger) Debugf(f string, a ...interface{}) {
+	debuglog.Logger.Debug("posthog: " + fmt.Sprintf(f, a...))
+}
+func (phLogger) Logf(f string, a ...interface{}) {
+	debuglog.Logger.Debug("posthog: " + fmt.Sprintf(f, a...))
+}
+func (phLogger) Warnf(f string, a ...interface{}) {
+	debuglog.Logger.Debug("posthog: " + fmt.Sprintf(f, a...))
+}
+func (phLogger) Errorf(f string, a ...interface{}) {
```
hayke102, line 286: `Errorf` and `Warnf` both funnel into `Debug`, which the default log level throws away — `debuglog` only enables `Debug` when `FLEET_DEBUG` is set (`debuglog.go:44`).

So a revoked/typo'd project key, or a 4xx from PostHog, produces **zero trace** in a normal user's `debug.log`, and `capture`'s own enqueue error is `Debug` too. Telemetry can be completely broken and no bug report would show it.

The comment above justifies this as "analytics noise shouldn't surface as a fleet error" — but the thing that protects the Bubble Tea display is writing to the *file logger* rather than stdout/stderr, not the level. `Warnf`→`Warn` and `Errorf`→`Warn` cost nothing and stay invisible to the user.

internal/analytics/analytics.go
hayke102: Fixed — confirmed at posthog.go:1723, the SDK debugf's the whole prepared message, so distinct_id (the git email in full mode) was landing in debug.log and from there into public bug reports. Debugf/Logf are now no-ops with a comment explaining why they must stay that way.

internal/analytics/analytics.go
hayke102: Fixed — Warnf/Errorf now go to Warn, so a revoked key or a 4xx is visible without FLEET_DEBUG. Audited every Errorf/Warnf call site in the SDK first: they log counts and error strings only, no payloads, so this doesn't reintroduce the leak from the sibling thread.

internal/analytics/analytics.go
hayke102: Fixed — confirmed at posthog.go:568 that c.msgs <- prepared has no default case (and it marshals on the caller's goroutine too). Client now fronts the SDK with a 256-deep drop-on-full queue drained by a background goroutine, so Track from Update() can't block. A full queue drops the event rather than freezing the UI.

internal/analytics/analytics.go
hayke102: Fixed — split Shutdown into detach() (clears the global, synchronous) and shutdownClient() (the flush). SyncEnabled now does 'go shutdownClient(detach())': the detach still runs on the Update goroutine, since Go evaluates call arguments before starting the goroutine, so Init still installs the new client — only the network flush moves off-thread. Quit keeps the blocking path, bounded at 2s.

internal/analytics/analytics.go
hayke102: Fixed — flushInterval 5s -> 1s, shutdownTimeout 500ms -> 2s. app_started and the app_active heartbeat both fire at launch, so the old pairing quietly undercounted exactly the launch-and-quit population DAU and the decline rate exist to measure. The 2s is a cap, not a fixed wait, and it's only affordable because the settings-close path is no longer synchronous.

internal/analytics/analytics.go
hayke102: Docs fixed; keeping the guard. CLAUDE.md no longer claims the client 'stays disabled until a project key is configured' — that stopped being true when the key was embedded. Left the key == "" branch in: it's five lines, TestMissingKeyDisables covers it, and it's what keeps 'no key => send nothing' true for a build with the key stripped. That's why projectAPIKey is a var rather than a const.

internal/hooks/claude_hooks.go
hayke102: Fixed — the LookPath-miss arm now returns the resolved exe path instead of the bare literal, with a warn that hooks will break once the process exits. A hook that works until you quit beats one that never resolves at all. (The other "fleet" literal, in the os.Executable() error arm, has to stay: exe is empty there, so there's genuinely nothing to fall back to.)

internal/hooks/claude_hooks.go
hayke102: Fixed — replaced the ancestor walk with an exact fixed-depth check (exe dir / b<N> step dir / go-build* dir). Added your false positive to the test table: ~/go-build-tools/dist/exe/fleet was indeed being misread as a go run temp.

internal/hooks/claude_hooks.go
hayke102: Fixed — GetHookCommand now shell-quotes the path (POSIX '\'' escaping), with a test covering a path containing a space. Verified in a real sh that the quoted command splits into exactly 3 args with argv0 intact. Note the OpenCode plugin was already safe: it JSON-marshals the path into a spawnSync argv array, so no shell is involved.

internal/hooks/claude_hooks.go
hayke102: Fixed — you were right, and the comment claiming 'same pattern as hooks.GetHookCommand' was actively false. Exported FleetBinaryPath (unquoted; GetHookCommand does its own quoting) and install.go now calls it, so the NMH manifest inherits the go-run handling. Confirmed the OpenCode plugin already routed through the same resolver.

PR #204, opened 2026-07-14
feat(ui): Shift+↑/↓ jumps between sidebar group headers

hayke102:
`Shift+↑` / `Shift+↓` jump the sidebar cursor to the previous / next group header (origin or checkout) instead of moving one row at a time.

From a session row the nearest header above is that session's own checkout header, so `Shift+↑` surfaces the group you're standing in before climbing out of it. With no header left to reach, the motion clamps to the first/last row — so it doubles as jump-to-top/bottom.

No reveal or rebuild is needed: a collapsed group still emits its header row, so every target is already in `flatItems` and landing on a collapsed group does not expand it.

## Changes
- `NextHeaderItem` + `LastSelectableItem` (`internal/ui/sidebar.go`)
- `jumpToHeader` + the `shift+up` / `shift+down` cases in `handleKey` (`internal/ui/app.go`)
- `header_jump` analytics event, help-dialog nav row, CLAUDE.md keybinding line
- Tests in `internal/ui/jump_test.go`, including one that drives the real `handleKey` switch and guards the `msg.String()` binding literal

https://claude.ai/code/session_01RThmJtRUFUCiXTRrZxJiHW

## Summary by CodeRabbit

* **New Features**
  * Added `Shift+↑` and `Shift+↓` navigation to jump between sidebar group headers.
  * Navigation skips session rows and spacer rows, stops on collapsed headers, and remains within the list boundaries.
  * Updated the keyboard shortcuts help to include the new navigation commands.

* **Documentation**
  * Added release notes describing the improved header-jump navigation.


internal/ui/sidebar.go
```diff
@@ -786,6 +786,37 @@ func FirstSelectableItem(items []SidebarItem) int {
...
+// `direction`. From a session row the nearest header above is that session's own
+// checkout header, so shift+↑ surfaces the group you're standing in before it
+// climbs out of it.
+//
+// With no header left to reach, it clamps to the edge of the list rather than
+// stalling — so the motion doubles as a jump to top/bottom. Headers are never
+// spacers and the clamp goes through First/LastSelectableItem, so the cursor
+// can't land on a gap row.
+func NextHeaderItem(items []SidebarItem, current, direction int) int {
+	for i := current + direction; i >= 0 && i < len(items); i += direction {
```
hayke102, line 823: A stale `current` teleports the cursor to the top of the list, where `NextSelectableItem` would clamp it to the end.

`rebuildFlatItems()` doesn't clamp `h.cursor` — only `syncViewport()` does — so there's a window (a shrinking rebuild on the worker tick: sessions deleted, a group collapsed) where `h.cursor >= len(flatItems)`. In that state:

- `Shift+↑` enters this loop at `i = current-1`, which is *still* `>= len(items)`, so the loop body never runs and we fall through to `return FirstSelectableItem(items)` → **cursor jumps to row 0**.
- `j`/`k` in the same state return `current`, and `syncViewport` then clamps them to the last row.

So the two navigations disagree about what a stale cursor means. Same shape on an empty list: `jumpToHeader` unconditionally sets `cursor = 0` (both `First`/`LastSelectableItem` return `0` for an empty slice) where `NextSelectableItem` is a no-op.

Clamping `current` into range before the scan fixes both:

```go
func NextHeaderItem(items []SidebarItem, current, direction int) int {
	if len(items) == 0 {
		return 0
	}
	if current < 0 {
		current = 0
	} else if current >= len(items) {
		current = len(items) - 1
	}
	for i := current + direction; i >= 0 && i < len(items); i += direction {
		...
```

internal/ui/jump_test.go
```diff
@@ -183,3 +185,184 @@ func TestSlotJumpExpandsCollapsedOrigin(t *testing.T) {
...
+// finally clamps to the top of the list instead of stalling.
+func TestHeaderJumpUpClimbsOutOfGroup(t *testing.T) {
+	h, s := headerJumpHome()
+
+	h.cursor = idxOfSession(t, h, s["a3"].ID) // inside the alpha worktree
+	want := []int{
+		idxOfCheckout(t, h, "/tmp/hj-a-wt"),   // own checkout header
+		idxOfCheckout(t, h, "/tmp/hj-a-main"), // previous checkout header
+		idxOfOrigin(t, h, "github.com/acme/alpha"),
+		FirstSelectableItem(h.flatItems), // no header above → clamp to top
```
hayke102: This clamp assertion is vacuous — it expects the same index as the step before it.

In `headerJumpHome` the alpha origin header is row **0** (no spacer precedes the first origin), so `idxOfOrigin("github.com/acme/alpha")` and `FirstSelectableItem(h.flatItems)` are both `0`. Steps 3 and 4 of `want` therefore assert the identical index, and the `// no header above → clamp to top` step adds nothing over the no-wrap check right below it.

Concretely: delete the `return FirstSelectableItem(items)` branch in `NextHeaderItem` and replace it with `return current` — this test still passes. The clamp path has no coverage.

To actually exercise it, the cursor needs to end up somewhere the clamp target differs from the last header — e.g. assert it separately from a list whose first row isn't itself a header, or seed the cursor on a session in the *first* checkout and check that a `Shift+↑` past the origin header stays put rather than wrapping. (The `Shift+↓` clamp at line 295 is the real thing — `LastSelectableItem` there is a session row, not the header the previous step landed on.)

internal/ui/sidebar.go
hayke102: Fixed — clamped `current` into range at the top of `NextHeaderItem` (plus an empty-list guard), as suggested.

One correction to the diagnosis and one addition:

**Correction:** I couldn't find a live path that reaches the stale state today. `rebuildFlatItems` indeed doesn't clamp, but every *shrinking* rebuild already fixes the cursor itself — the delete handler has an explicit `// Fix cursor.` block (app.go:4262), both filter paths reset to `FirstSelectableItem`, and a collapse leaves the cursor on the header row, which survives. So this is defense-in-depth over an invariant currently upheld by ~38 call sites, not a reproducible bug. Still worth it: the failure mode is asymmetric (j/k degrade to a no-op, Shift+↑ teleports to row 0), so a future call site that forgets to clamp fails loudly in one direction and silently in the other.

**Addition:** the same root cause bites a *negative* cursor in the other direction — `NextHeaderItem(items, -3, 1)` ran the loop zero times and fell through to `LastSelectableItem`, teleporting to the **bottom**. Both ends are covered by the new test.

internal/ui/jump_test.go
hayke102: Fixed — dropped the vacuous 4th step (the no-wrap check right below already covers "Shift+↑ on row 0 stays put"), and added `TestHeaderJumpCursorOutOfRange` to cover the clamp for real.

You were right, and it was worse than stated: `BuildFlatItems` always emits an origin header as row 0, so walking up from *any* in-range row hits a header before running off the top. The `FirstSelectableItem` branch could therefore only ever be reached with `current == 0`, where it returns 0 — i.e. `current`. The up-clamp wasn't weakly covered, it was **a no-op by construction**.

Fixing the sibling thread is what gives it observable behavior: with `current` clamped into range, an out-of-range cursor now walks up from the last row and finds a real header. The new test asserts exactly that (`NextHeaderItem(items, len+5, -1)` → the last header, not row 0), and I verified it fails on the old code via the precise mutation you suggested — delete the clamp and it goes red.

PR #205, opened 2026-07-15
Native Linux support

tokerido:
## Summary

fleet is already almost entirely portable — the codebase is pure Go, `.goreleaser.yml` already cross-compiles `linux/{amd64,arm64}`, CI already runs on `ubuntu-latest`, and several subsystems (config paths, URL-open, Chrome native-messaging paths) already branch correctly for Linux. This PR closes the remaining gaps and flips the "macOS-only" policy documented in the README/site.

**The one genuinely macOS-only subsystem was `internal/proc`** (shells out to `lsof`/`ps` to clear leftover dev daemons off a worktree before removal). It's now split by build tag: `proc_darwin.go` keeps the existing lsof/ps path untouched, `proc_linux.go` adds a native `/proc/<pid>/{cwd,exe,fd}` walk with no subprocesses. The exported API (`FindHolders`, `ForegroundCommands`, `Kill`, `Alive`) is unchanged, so every call site is unaffected.

Everything else follows the same pattern:
- **Clipboard** — `tmux copy-command` now routes through `wl-copy`/`xclip`/`xsel` (falling back to OSC 52 via `set-clipboard on` when no display-reachable tool exists), same server-option mechanism as macOS's `pbcopy`.
- **tmux compatibility** — `allow-passthrough` (added in tmux 3.3) is dropped from the batched `set-option` call on older servers instead of aborting the whole batch; Ubuntu 22.04 LTS ships 3.2a. tmux ≥ 3.3 is recommended for full feature parity; 24.04 ships 3.4.
- **Memory-pressure suspend** — Linux reads PSI (`/proc/pressure/memory`) instead of the macOS Jetsam sysctl. Low free swap only escalates a PSI *warning* to critical on Linux (never a normal/unknown reading), since Linux swap partitions are fixed-size and sit partially used on healthy boxes — unlike macOS's demand-grown swap files, where the same absolute threshold is a reliable signal on its own.
- **Diagnostics** — bug reports carry the distro `PRETTY_NAME` (`/etc/os-release`) and kernel version where macOS reports `sw_vers`.
- **Packaging** — a root `Dockerfile` (golang:1.26 → ubuntu:24.04), `.deb`/`.rpm` via GoReleaser's built-in `nfpm`, and an optional `contrib/systemd/fleet-tmux.service` user unit for keeping the tmux server alive across logins.

## Testing

- `go build`/`go vet` clean for both `GOOS=linux` and `GOOS=darwin`.
- `go test -race ./...` passes on Linux (ubuntu container, non-root user, tmux 3.5a).
- Live-validated end-to-end on WSL2 Ubuntu 24.04 (real tmux 3.4, real Claude Code sessions): session lifecycle, PTY attach/resize/detach, drawer foreground-command display, clipboard copy via xclip, and — the core scenario — worktree deletion correctly killing a leftover daemon process via the new `/proc` walk while sparing the pane's own login shell.
- A follow-up self-review caught and fixed: a login-shell (`-bash`) escaping the never-kill set on the initial Linux proc port, and the swap-pressure override described above.

## Test plan

- [ ] `go test -race ./...` on a Linux runner (already covered by existing CI, which runs on `ubuntu-latest`)
- [ ] `GOOS=darwin go build ./...` to confirm no macOS regression
- [ ] Manual smoke test on a real Linux box: create a session, open the drawer, delete a worktree with a leftover background process running in it

## Summary by CodeRabbit

* **New Features**
  * Added native Linux support, including `.deb`, `.rpm`, and Docker installation options.
  * Added optional systemd integration to keep tmux sessions available across logins.
  * Added Linux-aware worktree cleanup, process detection, clipboard integration, and idle-session suspension.
  * Added Wayland, X11, and OSC 52 clipboard fallbacks.
* **Bug Fixes**
  * Improved diagnostics with Linux distribution and kernel information.
  * Prevented automatic updates when the installation cannot be replaced.
  * Improved compatibility with older tmux versions.
* **Documentation**
  * Expanded installation, requirements, Linux support, and platform-specific behavior documentation.


internal/tmux/tmux.go
```diff
@@ -86,9 +163,57 @@ func EnsureCopyCommand() {
...
+	if goos != "linux" {
+		return "pbcopy"
+	}
+	if getenv("WAYLAND_DISPLAY") != "" && hasTool("wl-copy") {
+		return "wl-copy"
+	}
+	if getenv("DISPLAY") != "" {
+		if hasTool("xclip") {
+			return "xclip -selection clipboard -in"
+		}
```
hayke102, line 242: `xclip` without a stdout redirect is the classic tmux copy-pipe hang: xclip forks to serve the selection but never closes its inherited stdout, so tmux blocks on the pipe after every copy — the pane freezes until something else takes clipboard ownership. The tmux wiki documents exactly this and recommends redirecting (`xclip -i -selection clipboard >/dev/null`) or preferring `xsel`. Since `DISPLAY`+xclip is the default path on most X11 desktops, as written every copy-mode copy hangs the UI.

Suggest `xclip -selection clipboard -in >/dev/null` (and updating the pinned string in `clipboard_test.go`).

internal/tmux/tmux.go
```diff
@@ -86,9 +163,57 @@ func EnsureCopyCommand() {
...
+// so the routing table is testable without a real PATH or display server.
+// Wayland sessions normally export DISPLAY too (XWayland), so checking
+// WAYLAND_DISPLAY first keeps the wl-copy preference while still letting a
+// Wayland session without wl-clipboard fall through to the X11 tools.
+func clipboardCopyCommandFor(goos string, getenv func(string) string, hasTool func(string) bool) string {
+	if goos != "linux" {
+		return "pbcopy"
+	}
+	if getenv("WAYLAND_DISPLAY") != "" && hasTool("wl-copy") {
+		return "wl-copy"
```
hayke102, line 232: This checks `WAYLAND_DISPLAY`/`DISPLAY` in **fleet's** environment, but `copy-command` is executed by the **tmux server**, which has its own env — and `contrib/systemd/fleet-tmux.service` in this same PR starts that server under systemd with no display vars at all. In that setup, fleet run from a Wayland terminal passes this check and sets `copy-command wl-copy`; the server later spawns `wl-copy` without `WAYLAND_DISPLAY`, it exits non-zero, and the selection silently vanishes — the exact "strictly worse than OSC 52" failure the comment above says this check exists to prevent. Worse, having set `copy-command` also preempts the `set-clipboard` fallback the user would otherwise get.

Suggest reading the vars from the server's global environment (`tmux show-environment -g`) instead of `os.Getenv`, or propagating them (`tmux set-environment -g`) before setting `copy-command`.

internal/tmux/tmux.go
```diff
@@ -86,9 +163,57 @@ func EnsureCopyCommand() {
...
+	}
+	if copyCmd := clipboardCopyCommand(); copyCmd != "" {
+		_ = exec.Command("tmux", "set-option", "-s", "copy-command", copyCmd).Run()
+		return
 	}
+	// No local clipboard tool (headless Linux, no wl-copy/xclip/xsel): enable
+	// set-clipboard so copy-mode falls back to OSC 52 and terminals that
+	// support it (most modern Linux terminals, and anything over SSH) still
+	// get the selection.
+	_ = exec.Command("tmux", "set-option", "-s", "set-clipboard", "on").Run()
```
hayke102: The `copy-command` branch above carefully checks "only when unset" before writing, but this writes `set-clipboard on` unconditionally — clobbering a deliberate `set -s set-clipboard off` (a common hardening choice: it stops pane programs writing the user's clipboard via OSC 52), and re-clobbering it on every `EnsureCopyCommand` call (each session `Start` plus the startup bootstrap).

Same guard as copy-command: skip when `tmux show-options -sv set-clipboard` returns anything (only explicitly-set options print without `-A`, so non-empty = the user configured it).

contrib/systemd/fleet-tmux.service
```diff
@@ -0,0 +1,32 @@
...
+#   loginctl enable-linger "$USER"
+
+[Unit]
+Description=tmux server for fleet agent sessions
+
+[Service]
+Type=forking
+# start-server only: no windows are created; fleet attaches its sessions to
+# this server as usual.
+ExecStart=/usr/bin/tmux start-server
```
hayke102: Three issues that together mean this unit doesn't deliver what its header promises:

1. **It's a no-op**: `tmux start-server` with zero sessions exits immediately under the default `exit-empty on` (reproducible: `tmux -f /dev/null -L t start-server; sleep 1; tmux -L t ls` → "no server running"), while `RemainAfterExit=yes` keeps the unit reporting *active*. Needs something like `ExecStart=tmux start-server \; set -s exit-empty off`.
2. **Hardcoded `/usr/bin/tmux`** fails with 203/EXEC on Homebrew-on-Linux / Nix / built-from-source installs — systemd does no PATH lookup.
3. **`KillMode=none` is deprecated** — systemd (v242+) logs a warning on every load and may eventually ignore it, at which point stopping the unit kills the server and every agent session in it. The supported spelling of the same intent is `KillMode=process` plus an explicit no-op `ExecStop`.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN go mod download
+COPY . .
+ARG VERSION=dev
+RUN CGO_ENABLED=0 go build -trimpath \
+    -ldflags "-s -w -X main.version=${VERSION}" \
+    -o /out/fleet ./cmd/fleet
+
+FROM ubuntu:24.04
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        tmux git ca-certificates xclip locales \
```
hayke102: The documented `docker run -it --rm -v "$PWD:/repo" -w /repo fleet` flow fails on two counts:

1. **No agent binaries**: the image ships none of `claude`/`codex`/`opencode`, and session creation gates on `exec.LookPath(ag.Binary())` — so pressing `a` errors for every agent. The header comment covers the missing `gh` but not this much bigger gap; worth installing at least one agent or documenting the expectation.
2. **git "dubious ownership"**: the container runs as root over a host-uid mount, and Ubuntu 24.04's git refuses the repo — no branch/dirty detection, no worktree creation. Needs `git config --system --add safe.directory '*'` in the image (or a matching `USER`).

Also, `xclip` here is dead weight: with no `DISPLAY` in the container, `clipboardCopyCommandFor` returns `""` and OSC 52 is used regardless.

changelog/unreleased/linux-review-fixes.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+- **Copies land on Linux.** Copy-mode selections now check that a clipboard tool's display server is actually reachable (`WAYLAND_DISPLAY`/`DISPLAY`) instead of silently piping into a tool that can't connect — headless and SSH sessions fall back to OSC 52 as intended.
```
hayke102: This fragment and `linux-review-findings.md` are `type: fixed` entries for bugs that were introduced *and* fixed inside this PR — no released fleet ever had them. In the release notes they'd read as regressions in a platform that ships for the first time in the same release ("Idle-suspend works on Linux… previously Linux sessions were never protected" right under the "Native Linux support" highlight).

Suggest deleting both and folding anything still worth saying into `linux-support.md`.

tokerido:
Thanks for the thorough review, @hayke102 — all three tmux/clipboard bugs, the systemd unit, and the Dockerfile issues are real and now fixed, plus the Copilot and CodeRabbit findings. Summary of what changed (all pushed):

**Clipboard (tmux.go)**
- xclip's stdout now redirects to `/dev/null` — confirmed this was the classic tmux+xclip hang.
- The Wayland/X11 display check now reads the **tmux server's** environment (`tmux show-environment -g`) instead of fleet's own process env, so a server started headless (e.g. by the systemd unit) doesn't get a copy-command it can't actually run.
- `set-clipboard` is now guarded the same way `copy-command` already was — it no longer clobbers an explicit `set-clipboard off` on every call.

**systemd unit** — verified all three with `systemd-analyze verify` plus a live repro in a container:
- `exit-empty` is now turned off in the same chained tmux invocation; confirmed the old unit's server actually dies within ~1s of starting with zero sessions (reproduced your repro exactly), and the fixed one persists.
- `tmux` is resolved via `/usr/bin/env` instead of a hardcoded path.
- `KillMode=none` → `KillMode=process` + a no-op `ExecStop`, per your suggestion.

(Found one more bug validating the systemd fix: a bare `;` in `ExecStart=` is systemd's *own* multi-command separator and requires `Type=oneshot` — unrelated to tmux's `\;` chaining syntax, which is what's actually needed here. `systemd-analyze verify` caught it immediately.)

**Dockerfile**
- Added `git config --system --add safe.directory '*'` for the bind-mount ownership mismatch (couldn't reproduce the actual dubious-ownership error on my Windows/Docker-Desktop host — it maps bind-mounted files to root either way — but it's exactly the well-documented Docker+git≥2.35.2 gotcha, and the fix is a no-op where it doesn't apply).
- Documented the missing-agent-binary gap explicitly instead of bundling one (none of the three supported agents are apt packages, and picking one felt presumptuous) — the header now shows how to extend the image.
- Dropped `xclip`: no `DISPLAY` in a plain container, so it was dead weight.

**Changelog** — deleted `linux-review-fixes.md`/`linux-review-findings.md` (agreed they'd read as regressions in a same-release feature) and folded the real content into the single `linux-support.md` entry. Also dropped the `highlight: true` I'd set without asking — not a highlight.

**Docs** — `requirements.mdx` now mentions the Chromium NMH path alongside Chrome; `OSReleasePrettyName` now falls back to `/usr/lib/os-release` per os-release(5).

Full `go test -race ./...` passes on Linux (non-root, container) on the updated tree.

internal/tmux/tmux.go
```diff
@@ -86,9 +163,100 @@ func EnsureCopyCommand() {
...
+	// set-clipboard so copy-mode falls back to OSC 52 and terminals that
+	// support it (most modern Linux terminals, and anything over SSH) still
+	// get the selection. Same "only when unset" guard as copy-command above —
+	// otherwise this would silently override a user's deliberate
+	// `set-clipboard off` (a common hardening choice) on every call.
+	scOut, err := exec.Command("tmux", "show-options", "-sv", "set-clipboard").Output()
+	if err != nil {
+		return
+	}
+	if strings.TrimSpace(string(scOut)) != "" {
```
hayke102: This guard can never pass, so the OSC 52 fallback below is dead code.

`set-clipboard` has a **non-empty default**, unlike `copy-command`. Verified on a pristine server (`tmux -f /dev/null -L t`):

```
$ tmux -L t show-options -sv copy-command
            <- empty
$ tmux -L t show-options -sv set-clipboard
external
```

So `strings.TrimSpace(scOut) != ""` is always true, the function returns at :184, and `set-option -s set-clipboard on` (:186) never runs. That makes lines 173-186 unreachable — and `docs/linux.md:47` ("With no usable tool it sets `set-clipboard on` so tmux falls back to OSC 52") plus the changelog fragment both promise behavior that never executes.

The copy-command guard works precisely because its default *is* empty; copying the same test to an option with a real default inverts it. Since `external` is the default, treating it as "unset" is the pragmatic fix — e.g. only bail on `off`/`on` (explicitly configured), or compare against `external` rather than `""`. Worth a test that pins the real tmux default so this can't silently regress.

internal/tmux/tmux.go
```diff
@@ -86,9 +163,100 @@ func EnsureCopyCommand() {
...
+// copy-command runs as a child of the *server*, which inherits this table —
+// not fleet's env — when it execs a command. The two commonly diverge: the
+// systemd unit in this repo starts the server with no display at all, so a
+// server-launched-headless-then-attached-from-a-Wayland-terminal session
+// would otherwise pass this check on fleet's env and set copy-command to
+// wl-copy, which the server then fails to run with no WAYLAND_DISPLAY of its
+// own — silently losing the copy, worse than the OSC 52 fallback this
+// function exists to prefer when no tool can actually be reached.
+func tmuxServerGetenv(name string) string {
+	out, err := exec.Command("tmux", "show-environment", "-g", name).Output()
```
hayke102: This reads the wrong environment table — and the fix's own motivating example doesn't hold.

tmux runs `copy-command` jobs via `job_run(cmd, ..., s, ...)` (window-copy.c) → `environ_for_session(s)` (job.c), which copies `global_environ` then overlays the **session's** environment (environ.c) — session wins. `show-environment -g` is strictly the global table, and that table is frozen at server start; `update-environment` seeds `DISPLAY`/`XAUTHORITY` into the *session* env on create/attach, never into global.

So the gate is wrong in both directions:

- **Server started headless, used from X11** (the case the comment cites — this repo's unit is `WantedBy=default.target`, so it starts before the graphical session exports DISPLAY): `show-environment -g DISPLAY` is empty, fleet skips an `xclip`/`xsel` that would have worked, and downgrades to the fallback this docstring argues is inferior.
- **The converse, which is worse**: `update-environment` *removes* vars missing from the attaching client, so a server started under X11 with a session attached from an SSH client has `DISPLAY` in global but `-DISPLAY` in the session. fleet sets `xclip`, the job runs with no DISPLAY, and copies vanish silently — with no OSC 52 fallback, since `copy-command` is now set.

Suggest reading the session env (`show-environment -t <session> NAME`, falling back to global), or just using fleet's own process env — fleet's client is what attaches and seeds that session table. Note `copy-command` is a *server* option while display reachability is *per-session*, so on a mixed server no single value is correct.

Two doc nits while here: the comment's `WAYLAND_DISPLAY` example doesn't demonstrate the problem — it isn't in tmux's default `update-environment` (`DISPLAY KRB5CCNAME SSH_ASKPASS SSH_AUTH_SOCK SSH_AGENT_PID SSH_CONNECTION WINDOWID XAUTHORITY`), so it reaches neither table and skipping `wl-copy` is correct by default. And `clipboard_test.go`'s "a headless-started server has no display vars in its table" is true of the global table but false of the session table the job actually uses.

The common path (launch fleet from a terminal, fleet forks the server) is unaffected, since global == launching env there — so this is a narrow trigger, not a day-one break.

contrib/systemd/fleet-tmux.service
```diff
@@ -0,0 +1,41 @@
...
+# systemd considers the unit started, is what makes the server actually
+# persist. `env` resolves tmux via PATH rather than a hardcoded path, so this
+# also works for Homebrew-on-Linux, Nix, or from-source installs.
+ExecStart=/usr/bin/env tmux start-server \; set -s exit-empty off
+# Don't tear the server (and every agent session in it) down when the unit
+# stops — leave that to `tmux kill-server` as a deliberate act. KillMode=none
+# is deprecated (systemd >= 243 warns on load); process + a no-op ExecStop is
+# the supported way to express the same "never signal it" intent.
+RemainAfterExit=yes
+KillMode=process
```
hayke102, line 51: `KillMode=process` is not equivalent to the deprecated `KillMode=none`, so this unit still does the thing the comment above says it prevents.

With `Type=forking` and no `PIDFile=`, `GuessMainPID` defaults to yes, and the daemonized tmux server stays in the unit's cgroup (double-forking reparents onto the user manager — a child subreaper — which doesn't change cgroup membership). systemd guesses the tmux server as MainPID and `KillMode=process` SIGTERMs exactly that PID on stop. `ExecStop=/bin/true` doesn't help — systemd.service(5): after ExecStop runs, "any processes remaining for it are terminated according to the KillMode= setting." `RemainAfterExit=yes` only affects whether the unit reports active, not the stop-time kill.

So `systemctl --user stop fleet-tmux` (or `restart`, or `disable --now`) kills the server and every agent session in it. Reproduced on systemd 252 and 257 with this exact file: MainPID resolves to `tmux: server`, and after stop `tmux ls` reports "no server running".

systemd.kill(5) is explicit about the difference: "If set to `process`, only the main process itself is killed" vs "If set to `none`, no process is killed." The `none` → `process` swap is what introduced this.

Caveat: if a tmux server already exists when the unit starts, `start-server` just connects, MainPID=0, and stop is harmless — but that's exactly the case where the unit did nothing. Since it exists to own the server at login, the owning case is the normal one.

Suggested fix: `Type=oneshot` + `RemainAfterExit=yes`, and **keep** `KillMode=process`. MainPID is then 0 so there's nothing to signal, and `process` still stops the default control-group mode from sweeping the cgroup — dropping KillMode entirely alongside `oneshot` is also destructive (verified). `GuessMainPID=no` works too. Avoid `KillSignal=SIGCONT` + `SendSIGKILL=no`; it wedges the stop job indefinitely.

.goreleaser.yml
```diff
@@ -53,6 +53,28 @@ archives:
...
+      - fleet
+    package_name: fleet
+    formats:
+      - deb
+      - rpm
+    homepage: https://github.com/brizzai/fleet
+    description: TUI for managing multiple Claude Code sessions in parallel
+    maintainer: brizzai <https://github.com/brizzai/fleet>
+    license: Apache-2.0
+    bindir: /usr/bin
```
hayke102, line 68: Two packaging issues in this `nfpms` block.

**1. `bindir: /usr/bin` permanently breaks auto-update, silently and unthrottled.**

`replaceBinary` writes next to the current executable:

```go
dir := filepath.Dir(exePath)
tmp, err := os.CreateTemp(dir, "fleet-update-*")   // update.go:210-211
```

`/usr/bin` is root-owned `0755`, so an unprivileged user running the packaged binary gets EACCES. Auto-update is default-on (`IsAutoUpdateEnabled()` returns true when `AutoUpdate == nil`) and the only gate at `main.go:118` is `version != "dev"`, which a goreleaser-stamped build passes.

Two things make it worse than a one-off failure:

- **The download happens first** — `downloadAsset` (update.go:147) pulls the tarball before `replaceBinary` (update.go:152) fails.
- **The throttle is bypassed in exactly the failing case** — `saveCheckTime()` runs only at update.go:137 (already-up-to-date) and update.go:156 (after a *successful* replace). The error return at :152-154 skips it, so `last_update_check` never refreshes and `ShouldCheck()`'s 1h window stops throttling. It doesn't bite at install time (while installed == latest, :137 re-arms it); it starts the first time a newer release exists, and from then on it's a full re-download every launch, indefinitely.

The user sees nothing — `main.go:139` logs to `debug.log` only. They're pinned to a stale version with no signal why. Options: probe writability in `replaceBinary` and return a clear "installed via package manager" error (and call `saveCheckTime()` so the download stops recurring); or detect a package-managed prefix at the `main.go:118` gate; or at minimum document it and point package users at `FLEET_AUTO_UPDATE_DISABLED`.

**2. No `contents:` block — the systemd unit isn't in the package.**

`docs/linux.md` and the unit's own header both say `cp contrib/systemd/fleet-tmux.service ~/.config/systemd/user/`, which fails for anyone who installed the `.deb`/`.rpm` and has no checkout. Worth shipping it via `contents:` (e.g. to `/usr/share/fleet/`) and updating the docs to point there.

internal/proc/proc_linux.go
```diff
@@ -0,0 +1,180 @@
...
+// unique catch (a worktree file mmap'd and then closed) isn't how the target
+// daemons hold a dir. Pids owned by other users fail with EPERM and are
+// skipped — the same blindness non-root lsof has on macOS, and irrelevant
+// here: the leftover daemons we hunt were spawned by the user's own session.
+//
+// Semantics note: unlike macOS, an open fd on Linux never blocks rmdir — the
+// point of this walk is not dodging EBUSY but killing leftover dev daemons
+// (watchers, dev servers) before the worktree is removed, so they don't keep
+// respawning files or hold node_modules recreating "Directory not empty".
+func findHolders(absDir string, extra map[string]bool) ([]Holder, error) {
```
hayke102, line 25: This backend walks `/proc` with no deadline, where the darwin one bounds `lsof` with `context.WithTimeout(..., 10*time.Second)` (`proc_darwin.go:13`). Worth closing for parity.

Scoping it, because it's narrower than it first looks: of the reads here, only `os.ReadFile(base + "/cmdline")` in `commandName` (:90) can actually block — `get_mm_cmdline` → `access_remote_vm` → `mmap_read_lock_killable`. The `exe` and `cwd` readlinks in `pidHoldsDir` cannot: `proc_exe_link` resolves via RCU (`get_file_rcu`) and `proc_cwd_link` is `task_lock` + `get_fs_pwd` — no `mmap_lock` on either. And `commandName` only runs for pids that already passed the `pidHoldsDir` gate (:35-38), so the full-table scan never enters the blocking path.

To actually hang you'd need the worktree on a wedged NFS/FUSE mount, a holder process stuck under `mmap_lock`, and a writer queued behind it in the same mm. If it did happen on the quit path the symptom is a "Shutting down…" spinner that doesn't resolve — not a raw-terminal freeze: `performShutdown` runs in a command goroutine, not on Update (per its doc comment at `app.go:2461`), and the second-Ctrl+C hatch at `app.go:604` exits cleanly.

Low severity — cheapest fix is a deadline around the `cmdline` read (or the whole walk), nothing structural.

docs/linux.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Diagnostics / bug reports
+
+On Linux, reports carry `PRETTY_NAME` from `/etc/os-release` plus
+`uname -r` where macOS reports `sw_vers`.
+
+## Packaging
+
+- `Dockerfile` — golang:1.26 build stage → ubuntu:24.04 runtime with tmux,
+  git, xclip.
```
hayke102: Stale after the Dockerfile revision: the image now installs `tmux git ca-certificates locales` — `xclip` was deliberately dropped ("No display server is available in a plain container…"). This line still says "with tmux, git, xclip." Line 81's "gh+xclip recommends" is still accurate for the `.deb`, so only this one needs the edit.

While in this section, one more thing worth a line: `perfwatch.MemoryPressure()` reads host-wide `/proc/pressure/memory` and `/proc/meminfo`, neither of which is cgroup-scoped, so in a memory-capped container it reports the host's numbers (verified: against a 256 MB cgroup with `memory.swap.max=0` at 92% utilization it returns `swapFreeMB=24643`).

Not a blocker — the documented `docker run` sets no `--memory`, so host memory genuinely is the constraint, and it's unreachable in the stock image anyway (no agent CLI → zero candidates → `maybeSuspendIdleSessions` returns before probing). It only bites someone who extends the image *and* sets `--memory`, and it fails safe (host-normal → `PressureNormal` → no-op, same outcome this doc already describes for PSI-less kernels). But this file documents the probe's semantics in detail and never mentions containers, in the PR that adds the Dockerfile — so either a one-line "host-scoped" note here, or prefer cgroup v2 `memory.pressure` / `memory.max` when present.

hayke102:
Nice work on this round — the xclip redirect, the `exit-empty off` fix, `env tmux`, `safe.directory`, and folding the changelog fragments all landed cleanly. 🙏

Second pass found a few things, mostly in the *fixes* rather than the original code — the `set-clipboard` guard and the `KillMode` swap both look right and turn out not to do what they say. I triple-checked those two on real tmux and real systemd before commenting, since they're the kind of thing that's easy to argue about from the man page alone.

Nothing here is a blocker for the Linux support itself, which is in good shape. Details in the inline threads.

tokerido:
Thanks for the thorough review rounds - everything raised is now addressed, plus a few things I caught doing a self-review pass in the same spirit. All on the branch as of the latest push.

From your review:

	•	show-environment -g read the wrong table → display vars now come from fleet’s own process environment. As you showed, copy-command jobs run with the session environment, which update-environment refreshes from the attaching client — and fleet’s terminal is the client its sessions get attached from, so its env is the closest available proxy for a single server-wide option. Both failure directions (headless server + X11 client, X11 server + SSH client) are gone.
	•	Dead set-clipboard guard → resolved by deleting the fallback rather than repairing the guard, deliberately: since the default is external, tmux already sends copy-mode selections via OSC 52 with no help from us, and a deliberate set-clipboard off stays untouched. The right amount of code here turned out to be none; docs updated to describe the real behavior.
	•	Type=forking in the systemd unit → Type=oneshot + RemainAfterExit, so no MAINPID is adopted and systemctl --user stop signals nothing; KillMode=process retained to keep the cgroup sweep off the daemonized server.
	•	Packaging → the unit now ships in the .deb/.rpm at /usr/lib/systemd/user/, so systemctl --user enable --now fleet-tmux works without a checkout. And the auto-updater now probes replaceability before downloading (package installs in /usr/bin get a clear “update via your package manager” error) and records the check time once the release check succeeds — a failed replace can no longer bypass the hourly throttle and re-download every launch.

From the self-review pass:

	•	The allow-passthrough gate had the same “wrong source” bug as the env table: tmux -V is the client binary’s version, but options are interpreted by the server — and on Linux a package upgrade swaps the binary without restarting a running server, so a 3.4 binary over a live 3.2a server would have aborted the whole batched set-option. The gate now asks the running server (display-message -p '#{version}' — verified it works on a session-less server and errors cleanly, without spawning a server, when none runs) and falls back to tmux -V only when no server is up yet, which is exactly the case where the binary’s version is the right answer.
	•	CI now cross-compiles GOOS=darwin: this PR introduces the repo’s first darwin-tagged files into an ubuntu-only CI, so a macOS build break would otherwise have surfaced only at release time.
	•	The editor bundle scan is now explicitly darwin-only instead of relying on /Applications not existing — ~/Applications is real on GNUstep desktops and would have routed into open -a, which doesn’t exist on Linux. docs/linux.md already promised this; now the code enforces it.

Verified: build + vet + tests green on linux and via darwin cross-compile; the tmux probe claims were checked against a live server rather than the man page.

internal/tmux/tmux.go
```diff
@@ -86,9 +176,71 @@ func EnsureCopyCommand() {
...
+	if copyCmd := clipboardCopyCommand(); copyCmd != "" {
+		_ = exec.Command("tmux", "set-option", "-s", "copy-command", copyCmd).Run()
+	}
+}
+
+// clipboardCopyCommand returns the command line tmux should pipe copy-mode
+// selections into, or "" when no usable clipboard tool exists. macOS always
+// has pbcopy. Linux requires both the tool on PATH *and* its display server
+// reachable: wl-copy without WAYLAND_DISPLAY (or xclip/xsel without DISPLAY)
+// exits non-zero and the selection silently vanishes — strictly worse than
```
hayke102: **Correction to my own earlier review — this premise is wrong, and I'm the one who put it here.**

"exits non-zero and the selection silently vanishes" isn't what tmux does. `copy-pipe-and-cancel` passes `set_clip=1`, so tmux emits the OSC 52 selection **in addition to** running `copy-command` — the two are not alternatives. A failing tool doesn't swallow the copy.

Verified on tmux 3.5a with a `copy-command` that consumes stdin and `exit 1`:

```
copy-command ran:  YES  (received "ZZMARKERZZ")
exit status:       1
OSC 52 emitted:    YES  ^[]52;;WlpNQVJLRVJaWg==   (base64 -> ZZMARKERZZ)
```

The one case where it *is* lost is `set-clipboard off` — same test, no OSC 52 at all. So the true rule is: **`set-clipboard` (default `external`) governs OSC 52 independently of whether `copy-command` succeeds.**

This matters beyond the comment, because the false premise is now load-bearing in three places — this docstring, `docs/linux.md`'s clipboard section, and the `CLAUDE.md` bullet — and it's the stated justification for both the reverted env source and the deleted `set-clipboard` branch.

**What it means for the actual code at :215.** Reverting to `os.Getenv` is right for `DISPLAY` (tmux's default `update-environment` includes it, so fleet's env is a good proxy for the session env). It is *not* right for `WAYLAND_DISPLAY` — that's absent from the default `update-environment`, so it reaches neither table, and on the headless-server setup this PR now packages to `/usr/lib/systemd/user/`, fleet can set `copy-command wl-copy` that the server can never execute — while also skipping the `xclip` that would have worked.

Good news: given the above, that degrades to OSC 52 rather than losing the copy, so it's **low severity, not the data-loss I implied last round**. Real exposure is a `set-clipboard off` user, or a terminal that doesn't accept OSC 52. Cheapest fix is to drop the `wl-copy` branch (or gate it on `DISPLAY` too, since Wayland sessions export it via XWayland) and let `xclip`/`xsel` handle it. Worth fixing the three comments regardless — the next person to touch this will reason from them.

internal/update/update.go
```diff
@@ -196,17 +211,33 @@ func downloadAsset(assetURL string) ([]byte, error) {
...
+
+// canReplace reports whether this process can swap the binary at exePath.
+// The rename in replaceBinary needs a writable *directory*, not a writable
+// binary, so the probe is the same operation replaceBinary starts with:
+// creating a temp file next to it.
+func canReplace(exePath string) error {
+	tmp, err := os.CreateTemp(filepath.Dir(exePath), "fleet-update-probe-*")
 	if err != nil {
-		return err
+		return fmt.Errorf("%s is not replaceable by this user (package-managed install?) — update via your package manager, or turn off auto-update in Settings: %w", exePath, err)
```
hayke102: The probe-before-download fix is right, but returning an `error` for a permanent, user-unfixable condition has two side effects worth reconsidering.

`Update()` propagates this to `main.go:139`, which does `debuglog.Logger.Error("auto-update failed", ...)`, and just above it queues `analytics.EventUpdateCheck` with `"error": true`. For a `.deb`/`.rpm` user the condition never clears, so:

1. **Every Linux bug report leads with a red herring.** `debug.log` gains an ERROR line hourly, and the `!` bug-report flow pastes the last 100 lines into a public issue — so "auto-update failed" is the first thing you'll read on reports that have nothing to do with updating.
2. **Telemetry can't distinguish this from a genuinely broken updater** — a growing population of `error: true` update checks that are actually working as designed.

`docs/linux.md:100-102` already describes the intended behavior accurately — the updater "skips with a hint in `debug.log`" — which is a *skip*, not a failure. Matching the code to the doc would mean `return "", nil` here (with `saveCheckTime()` already recorded above) plus an Info/Warn log, so `error` stays false. A sentinel like `errNotReplaceable` that `main.go` special-cases would work too, if you want `fleet update` to keep printing the message to stderr.

The wording itself is good — "update via your package manager, or turn off auto-update in Settings" is exactly the right hint; it's just being routed through the failure channel.

internal/tmux/tmux.go
```diff
@@ -54,18 +56,107 @@ var (
...
+// build that is almost certainly ahead of any release.
+//
+// Memoized: Start calls this per session — a reloadAll of N dead sessions
+// would otherwise fork N redundant probes. Memoizing is conservative: the
+// binary can't change under a running fleet, so a server restarted mid-run
+// can only be newer than the memoized answer, and a stale too-old answer
+// merely skips passthrough instead of aborting the batch.
+func serverVersion() (major, minor int, ok bool) {
+	versionOnce.Do(func() {
+		out, err := exec.Command("tmux", "display-message", "-p", "#{version}").Output()
```
hayke102, line 88: Asking the running server is the right call. Two small gaps in how the answer is handled:

**1. Unparseable output fails open.** The fallback to `tmux -V` only triggers on exec *error*. If `display-message -p '#{version}'` succeeds with empty or unexpected output, `parseTmuxVersion` returns `ok=false`, and `allowPassthroughSupported` (:147-149) then returns `true` — "assume modern" — which sends `allow-passthrough` and re-arms the exact batch abort this gate exists to prevent. That default is sound for `tmux -V` (a dev build like `next-3.6` really is modern), but it's the wrong default for a server probe, where an unreadable answer means "I don't know" rather than "new". Falling through to `tmux -V` on `!ok`, not just on `err`, keeps both behaviors correct.

**2. `versionOnce` now memoizes a server-derived value.** When the probe read the binary this was stable for the process; now the first `Start()` pins whichever server happened to be live, for the rest of the run. `tmux kill-server` + a package upgrade mid-session leaves the memo stale. Mostly harmless — the common direction is dropping `allow-passthrough` unnecessarily until restart — but note `EnsureCopyCommand` right below deliberately re-checks the live server every call for the same class of reason, so the two now disagree on whether server state is cacheable.

Both are low severity; flagging mainly because #1 silently inverts the gate's safe direction.

docs/linux.md
```diff
@@ -0,0 +1,114 @@
...
```
hayke102: Still stale from last round — the section moved but this line didn't: the image installs `tmux git ca-certificates locales`, no `xclip` (dropped in the Dockerfile revision with the comment "No display server is available in a plain container"). Line 95's "gh+xclip recommends" is correct, so it's just the word `xclip` on :93.

Two other leftovers from round 2, both optional:

- **The memory-pressure section (:58-74) is still silent on scope.** `/proc/pressure/memory` and `/proc/meminfo` are host-wide, not cgroup-scoped, so in a memory-capped container the probe reports the host's numbers. It fails safe and the documented `docker run` sets no `--memory`, so this is a one-line note ("host-scoped; a `--memory` limit isn't observed") rather than a code change — but this file documents the probe in real detail and never mentions containers, in the PR that adds the Dockerfile.
- **`internal/proc/proc_linux.go` still has no deadline** where `proc_darwin.go` bounds `lsof` at 10s. I flagged it as low last round and it's untouched — fine to leave deliberately, just noting it didn't get lost silently.

One new thing worth a line in the systemd section: with `Type=oneshot` + `RemainAfterExit=yes` the unit reports `active (exited)` even after the server is gone, so following the unit's own advice (`tmux kill-server` as the deliberate teardown) leaves `systemctl --user start fleet-tmux` a silent no-op — `restart` is the one that works. That's an acceptable consequence of the fix, but it'll surprise someone.

hayke102:
Green across the board — `lint`, `test`, `vulncheck` and the changelog check all pass. 🎉

Thanks for the back-and-forth on this one, @tokerido. The systemd fix in particular was the awkward kind: `KillMode=process` reads like it means "don't kill it" and doesn't, and the working combination (`Type=oneshot` + `RemainAfterExit` while *keeping* `KillMode=process`) isn't something you'd guess from the man page. You landed it exactly right.

I pushed one small commit (accb3fe) with three follow-ups: the package-install update path now reports as a skip rather than an hourly ERROR, and I corrected the copy-command/OSC 52 model in the docstring, `docs/linux.md` and `CLAUDE.md`. That last one was my mistake originally — I told you a failing clipboard tool swallows the copy, and it doesn't: `copy-pipe-and-cancel` passes `set_clip=1`, so OSC 52 fires alongside `copy-command`. Only `set-clipboard off` actually loses it. Sorry for sending you down that path twice; the corrected version is in the code now.

Two things I left for you as judgment calls rather than changing them myself:

- The `wl-copy` branch still keys off `WAYLAND_DISPLAY` from fleet's env, which isn't in tmux's default `update-environment`. Low impact now that OSC 52 covers the gap, but it means `wl-copy` can get set where `xclip` would have worked.
- `serverVersion`'s `!ok → assume modern` default made sense for `tmux -V`; for the server probe an unreadable answer arguably means "don't know" rather than "new".

Neither blocks anything. Nice work getting Linux over the line.

PR #209, opened 2026-07-20
feat(ui): snooze a session, worktree, or repo with `z`

hayke102:
## Why

fleet's attention model runs through exactly three surfaces: the `Space` jump rotation, the status pills, and the row's status dot. There's no volume knob. A session sitting in `waiting` or `finished` keeps claiming a slot in `Space` and keeps its tally in the pills forever — even when you've consciously decided you're not touching that work today. With a large fleet, "something needs you" degrades into noise you learn to skip past.

`z` on any row — session, worktree, or origin — mutes it for `30m` / `1h` / `4h` / tomorrow 9am, or a duration you type (`15m`, `30h`, `2d`). `z` again wakes it early.

```
▾ fleet 2  ●
  ▾ master  ●
    ● ✻ Fix the drawer resize bug
    ◐ ✻ Refactor status worker  ☾ 3h      ← own snooze: dimmed, keeps its real ◐
    · ✻ Add release notes reel
  ▾ snooze *  ☾ 45m                        ← group: dim + folded, pill → clock
    ◐ ✻ Snoozed group child  ☾             ← child: bare marker, no clock
    ● ✻ Another under snoozed repo  ☾
```

Those muted sessions dropped out of both header pills and out of the `Space` rotation.

## The design decision this rests on

**Snooze is not a Status, and not Suspend.**

The obvious move was to copy the `StatusSuspended` precedent (a 28-point integration checklist). That's wrong: a snoozed session that is *waiting* is still waiting — you just don't want to be told right now. Overwriting `Status` destroys that and forces reconstructing it on wake. So snooze is an orthogonal deadline that only the **attention** surfaces consult. The status pipeline is untouched.

It's also not Suspend. Suspend answers *"free my RAM"* (automatic, memory-pressure-driven, kills tmux); snooze answers *"stop nagging me"* (manual, intent-driven, never touches the process). Keeping them orthogonal is what makes it safe to snooze a session that's actively running.

Precedence lives in exactly one place — `snoozeState()`: origin umbrella → checkout umbrella → the session's own deadline. `BuildFlatItems` resolves it once per session and stamps `SidebarItem.Snooze`, so the jump scan and the pills read the same answer instead of re-deriving it. A session's own snooze runs *independently underneath* a group's, so it survives the group waking.

**Group snooze is an umbrella, never fanned out onto sessions** — it lives in a new `snoozed_groups` table keyed in the same space as `collapsed_groups`. That's why a session created *after* the snooze is muted too, which is the difference between a group snooze and a bulk edit. Snoozing a group also collapses it; waking re-expands, symmetrically.

## Two choices that look arbitrary but aren't

**The marker is `☾` U+263E, not `⏾` U+23FE.** U+23FE ("power sleep") is the obvious pick. I parsed the macOS font cmaps directly: **Menlo does not contain it** — it would render as a fallback box in Terminal's default font, exactly the failure that got U+2B21 rejected for the agent sigils. U+263E is covered by the same Menlo faces as the existing `✻`.

**Expiry rides the ~2s Update tick, not the worker.** `maybeSuspendIdleSessions` lives on the worker only because it probes sysctl and must not block Update. Snooze expiry is clock arithmetic plus a small SQLite write — keeping it on Update means `h.groupSnooze` needs no lock.

Also worth noting: the duration picker reuses the context menu's row-anchored overlay pattern rather than being a centered modal, and `parseSnoozeDuration` is deliberately **not** `time.ParseDuration` — that accepts combos (`1h30m`) and seconds we don't offer, and has no notion of days, so its error messages would describe a syntax fleet doesn't have. Errors here are user-facing; they render live in the box.

## Testing

`make test` (`-race`, 20 packages), `make lint` (0 issues), `make build` — all green.

New coverage:
- **Precedence table** — group beats own for display; own survives the group expiring; expired deadlines mute nothing.
- **Storage round-trip** — guards the hand-maintained positional SQL in `SaveSession`/`LoadSessions`, including the neighbouring column, since a mismatch there corrupts every field after it rather than just this one. Plus migration idempotency.
- **Umbrella semantics** — a session created under a snoozed group is muted and shows no clock of its own.
- **`Space` skips muted rows** (both individually-snoozed and whole-group), and reaches them again after expiry.
- **Pills** — muted sessions drop from checkout *and* origin tallies together.
- **Width budget** — a snoozed row never renders wider than an unsnoozed one (regression test for the hardcoded `reserve` in `renderSessionItem`).
- **Duration parser** — 12 rejection cases including combos, bare numbers, seconds, and the 30d cap.
- **Dialog** — typed value beats the highlighted preset; an unparseable box refuses rather than silently using the preset; box height/width never change as you type.

## Note on unrelated dead code

`CollectGroupInfo` (`sidebar.go`) has **zero callers**. It's a second, independent status-counting path that would have silently diverged from the pills here. Left alone as pre-existing, but worth a follow-up.

https://claude.ai/code/session_01ANzhiQ3thqW2GU4ePYkEAJ

## Summary by CodeRabbit

* **New Features**
  * Added Snooze for sessions, worktrees, and repositories using the `z` shortcut, context menus, or command palette.
  * Choose preset or custom durations, with visible countdowns and automatic wake-up at expiry.
  * Snoozed items remain active but are muted from rotation and status counts; groups are visually dimmed or collapsed.
  * Press `z` again or use Unsnooze to wake items early.

* **Documentation**
  * Added Snooze guidance to the changelog and keyboard conventions.


internal/ui/snooze.go
```diff
@@ -0,0 +1,316 @@
...
+// it out of your way, and a fold is the strongest way to say "not today". The
+// user can still re-expand it.
+func (h *Home) applySnoozeUntil(sc snoozeScope, until time.Time, durationID string) {
+	if sc.session != nil {
+		sc.session.SetSnoozedUntil(until)
+		if err := h.storage.SetSessionSnooze(sc.session.ID, until); err != nil {
+			debuglog.Logger.Error("failed to persist session snooze", "id", sc.session.ID, "err", err)
+		}
+	} else {
+		h.groupSnooze[sc.groupKey] = until
```
hayke102, line 207: **`snoozed_groups` rows outlive the group they belong to.**

`snoozed_groups` shares `collapsed_groups`' keyspace by design, but only one of them gets cleaned up on delete. `forgetCollapse` (app.go) calls `clearCollapse(repoPath)` — and `clearCollapse(OriginExpandKey(origin))` when the last checkout goes — but there's no parallel `SetGroupSnooze(key, time.Time{})` anywhere on the delete path.

So: snooze `~/code/fleet-wt` for 2d → `d` on its header to remove the worktree → the collapse key is dropped, but `h.groupSnooze["~/code/fleet-wt"]` and its SQLite row survive. Create a worktree at the same path an hour later and every session under it is born muted — dimmed, skipped by `Space`, absent from the pills, with a `☾` countdown nobody set — until the original deadline lapses.

Worth adding a `forgetSnooze(repoPath)` next to `forgetCollapse` and calling it from the same places. The two tables were built to share a keyspace, so they should share a lifecycle.

internal/ui/app.go
```diff
@@ -1421,6 +1454,21 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				h.repoExpanded[key] = false
 			}
 		}
+		// Restore group snoozes, dropping any that lapsed while fleet was
+		// closed so the very first frame is already correct (waiting for the
+		// sweep would paint a stale dim group for up to one interval).
+		if snoozed, err := h.storage.LoadSnoozedGroups(); err == nil {
```
hayke102, line 1466: **Group snoozes that lapse while fleet is closed never get their auto-collapse undone.**

`maybeWakeSnoozed` treats the re-expand as part of expiry ("Symmetric with the auto-collapse in `applySnooze`"), but this startup path drops the lapsed row and returns without ever calling `setExpanded(key, true)`.

Concretely: 18:00, snooze the `fleet` origin until "Tomorrow" — `applySnoozeUntil` collapses it. Quit. Relaunch at 10:00: `until.After(now)` is false, the row is deleted, and the group stays folded permanently, with no snooze marker left to explain why. Leave fleet running over the same window and it re-expands correctly.

Since "Tomorrow" is the preset most likely to span a restart, this is probably the common case rather than the edge one. Adding `h.setExpanded(key, true)` in the else branch here would make the two paths agree.

internal/ui/snooze.go
```diff
@@ -0,0 +1,316 @@
...
+		sc.session.SetSnoozedUntil(time.Time{})
+		if err := h.storage.SetSessionSnooze(sc.session.ID, time.Time{}); err != nil {
+			debuglog.Logger.Error("failed to clear session snooze", "id", sc.session.ID, "err", err)
+		}
+	} else {
+		delete(h.groupSnooze, sc.groupKey)
+		if err := h.storage.SetGroupSnooze(sc.groupKey, time.Time{}); err != nil {
+			debuglog.Logger.Error("failed to clear group snooze", "key", sc.groupKey, "err", err)
+		}
+		h.setExpanded(sc.groupKey, true)
```
hayke102, line 243: **Waking force-expands a group that was already collapsed before the snooze.**

The comment above says "`setExpanded` is idempotent, so a group the user already re-expanded by hand is unaffected" — true, but that's the case where the user *opened* it. The uncovered case is the user who had it *closed* to begin with:

1. Collapse `brizzai/fleet` by hand (you keep it folded).
2. Snooze it 1h — `applySnoozeUntil` calls `setExpanded(key, false)`, already false, no visible change.
3. `z` to wake → `setExpanded(key, true)` and it springs open.

Snooze silently took ownership of a collapse state that predated it. Same on the expiry path in `maybeWakeSnoozed`. If you want to keep this simple, remembering whether the snooze was the thing that collapsed it (only re-expand if `applySnoozeUntil` actually changed the state) is the minimal fix.

internal/ui/app.go
```diff
@@ -5790,7 +5873,17 @@ func (h *Home) cursorBreadcrumb(bg color.Color) string {
...
 func (h *Home) statusCountsLine(bg color.Color) string {
 	counts := make(map[session.Status]int)
+	originOf, _ := h.originResolvers()
+	now := time.Now()
 	for _, s := range h.sessions {
+		// Muted sessions are absent from the attention glyphs here for the
+		// same reason they're absent from the per-header pills — otherwise the
+		// global line contradicts the sidebar it sits above. Snooze gets no
+		// counter of its own: it is a sidebar-only signal by design.
+		repo := session.GetRepoRoot(s.ProjectPath)
```
hayke102: **This re-derives the precedence rule that `BuildFlatItems` already stamped — the one thing `snoozeState`'s doc forbids.**

`snooze.go` says "This is the single place the precedence rule lives. Callers consult the result — they must not re-derive it", and CLAUDE.md repeats it ("`BuildFlatItems` resolves it once per session and stamps `SidebarItem.Snooze`, so no downstream caller re-derives it — the jump scan and the pills both read that field"). `jumpToNextAttentionSession` follows that and reads `it.Snooze.Muted`; this one doesn't.

The practical risk is exactly what the exclusion was added to prevent: change precedence later (say a session snooze that outranks its group) in `BuildFlatItems` alone, and this global pill disagrees with the sidebar directly beneath it.

There's a cost side too — `statusCountsLine` runs every `View()` frame, and `originResolvers()` takes the gitInfo lock and builds three maps sized to it on each call, plus `GetRepoRoot` + `snoozeState` per session. That's on the ~60ms What's New animation tick as well. Reading the stamped `Snooze` off `h.flatItems` fixes the correctness risk and drops the per-frame work in one move.

internal/ui/sidebar.go
```diff
@@ -597,7 +669,11 @@ func renderSessionItem(s *session.Session, width int, selected bool, slot int) s
 	if glyphRaw == "" {
 		reserve = 11
 	}
-	maxTitleLen := width - reserve - len(slotRaw)
+	// The snooze suffix comes out of the title's budget like the slot badge,
+	// rather than bumping `reserve` — it's present on only some rows. Measured
+	// with ansi.StringWidth because the marker is a multi-byte rune (len() would
+	// over-reserve by 2 and truncate snoozed titles early).
+	maxTitleLen := width - reserve - len(slotRaw) - ansi.StringWidth(snoozeRaw)
```
hayke102: **Selected rows pay for `snoozeRaw` but never render it.**

`snoozeRaw` only reaches the output on the unselected branch (`styledSnooze`); the selected branch builds `row` from symbol + glyph + title + slot and then appends its own separate `☾ snoozed 4h · z to wake` note. But `maxTitleLen` is docked here on every row.

So the cursor row of a session snoozed 4h loses 6 cells (`"  ☾ 4h"`) of title for a suffix it doesn't draw — 'Fix the drawer resize bug' truncates to 'Fix the drawer resi…' while showing no `☾` at all. Moving the cursor off makes the title *longer* and adds the marker.

Gating the subtraction on `!selected` would do it.

Smaller adjacent thing in the same function: the new `switch` makes suspended and snoozed mutually exclusive, and selected rows draw no `☾` — so a row that's both suspended *and* snoozed shows zero snooze indication while still being skipped by `Space`. Probably fine to leave, but noting it since the marker visibly flickers as the cursor passes over.

internal/ui/snooze.go
hayke102: Fixed — added `forgetSnooze(repoPath)` next to `forgetCollapse` (app.go), mirroring it exactly including the last-checkout-drops-the-origin rule, and called it from both delete sites (`unpinRepo` and the deferred-delete handler).

Confirmed your reproduction before fixing: `SetGroupSnooze` appeared only in snooze.go's apply/clear/sweep plus the startup load — nothing on any delete path.

Regression test `TestForgetSnoozeClearsGroupState` asserts the in-memory entry, the SQLite row, and the orphaned origin key are all gone, then creates a session at the same path and asserts it is *not* born muted — the symptom rather than just the mechanism.

internal/ui/app.go
hayke102: Fixed — the startup else-branch now calls `h.setExpanded(key, true)` so both wake paths agree.

You were right that this is the common case rather than the edge: "Tomorrow" is precisely the preset that spans a restart.

One note while implementing — I first wrote this as a direct `h.repoExpanded[key] = true` + storage write, on the assumption the block ran off the Update goroutine. It doesn't: it's inside `Update` itself (app.go:601), so `setExpanded` is both safe and the correct API, since it keeps the map and the persisted row in sync through one path.

internal/ui/snooze.go
hayke102: Fixed the comment; keeping the behavior — deliberate call after weighing it.

You're right that the comment was misleading: it claimed idempotency covered the case, and it only covers the user who *re-opened* the group, not the one who had it folded to begin with.

On the behavior itself, the minimal fix you suggested (remember whether the snooze collapsed it) has a catch — that bit has to **persist**, because per the thread above the common wake path is the restart. An in-memory flag is gone by then. So it costs a `collapsed_by_snooze` column on `snoozed_groups` to buy a fairly small nicety. Not worth a migration right now.

So the comment now states the limitation rather than implying coverage:

```
// Known limitation: a group that was ALREADY folded before the snooze does
// spring open on wake — snooze takes ownership of a collapse state that
// predated it. Accepted deliberately: knowing the difference means persisting
// "the snooze is what collapsed this" across restarts (the restart is the
// common wake path — see the startup reconciliation in loadSessions), i.e. a
// schema change on snoozed_groups for a fairly small nicety.
```

Happy to revisit if it turns out to annoy in practice.

internal/ui/app.go
hayke102: Fixed — but not quite the way you suggested, because reading `h.flatItems` would have regressed it.

`BuildFlatItems` omits children of collapsed groups, and this pill is deliberately fleet-wide, so sourcing it from `flatItems` would silently drop every session inside a folded group from the global count. The diagnosis was right; that particular remedy trades one disagreement for another.

Instead `rebuildFlatItems` now also stamps `h.snoozeMuted map[string]bool` over **all** `h.sessions` with a single `originResolvers()` call, and `statusCountsLine` is a map lookup. That gets both things you were after — one evaluation path for the precedence rule, and the per-frame work (three maps under the gitInfo lock, on every frame including the ~60ms What's New tick) drops to once per rebuild.

`TestSnoozeMutedCoversCollapsedGroups` pins the part that would have broken: a snoozed session hidden inside a collapsed origin is absent from `flatItems` but still marked muted.

internal/ui/sidebar.go
hayke102: Fixed — the width subtraction is now gated on `!selected`.

Verified the test actually catches it by reverting the fix: the selected row truncated to `a really quite long sess…` vs `a really quite long session ti…` unselected — the 6 cells you predicted. `TestSelectedRowKeepsFullTitleBudget` asserts both render the same title.

One correction on the framing: the selected row does show the `☾` for own-snoozed sessions — the selected branch draws its own `☾ snoozed 4h · z to wake` note outside the pill, which is what made the double-charge easy to miss. The "no ☾ at all" case is specifically the suspended-**and**-snoozed row you flagged second, where the `switch` gives the single slot to suspended.

Leaving that one as-is, agreeing with your read: only one note fits before the row overflows, and suspended is the more actionable state (it changes what Enter does).

internal/ui/snooze_dialog.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	if n <= 0 {
+		return 0, fmt.Errorf("must be more than 0")
+	}
+
+	total := time.Duration(n) * mult
+	if total > snoozeMaxDuration {
+		return 0, fmt.Errorf("max 30d")
+	}
+	return total, nil
```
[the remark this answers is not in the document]
hayke102: Good catch — fixed. Reproduced it first:

```
9999999d  -> dur=-832516h4m24s    err=<nil>
9999999h  -> dur=-248192h9m7s     err=<nil>
999999d   -> dur=-1620501h52m48s  err=<nil>
107d      -> err="max 30d"        (correct)
```

Exactly as described: the product wraps negative, a negative is not `> snoozeMaxDuration`, so the cap never fires and the dialog previews a wake time in the past. My `n <= 0` guard was checking the parsed integer, not the product.

Now bounded before multiplying:

```go
if time.Duration(n) > snoozeMaxDuration/mult {
    return 0, fmt.Errorf("max 30d")
}
return time.Duration(n) * mult, nil
```

Added the overflow inputs to the rejection table in `TestParseSnoozeDuration`, plus a blanket assertion that no rejected input ever returns a negative duration — that covers the whole class rather than the four literals.

changelog/unreleased/snooze.md
```diff
@@ -0,0 +1,6 @@
+---
+type: added
+highlight: true
+---
+
+**Snooze the noise** — `z` mutes a session, worktree, or repo so work you've decided not to touch today stops claiming a slot in the `Space` rotation and stops counting toward the header pills. Pick `30m`/`1h`/`4h`/tomorrow 9am, or type your own (`15m`, `30h`, `2d`) — the picker shows the exact wake time before you commit. The session keeps running and keeps its real status; snoozing a repo dims and folds the whole group, including sessions you create in it later. Press `z` again to wake it early.
```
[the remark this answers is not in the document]
hayke102: Half taken, half declined.

**Length — agreed and trimmed.** CLAUDE.md does say 1–2 sentences and mine ran four. Now two.

**`highlight: true` — keeping it.** This is a deliberate editorial decision, not an oversight. CLAUDE.md requires that `highlight` is never chosen unilaterally ("Never decide `highlight` on your own — always ask"), so it was put to the repo owner when the fragment was written and they explicitly chose to highlight it. That exchange isn't visible from the diff, which is why this reads as unset-by-accident here.

PR #213, opened 2026-07-21
feat(agent): add Cursor CLI as a supported agent

VainEyal:
## Summary

- Adds Cursor CLI (`cursor-agent`) as a fourth session agent alongside Claude, Codex, and OpenCode: selectable from the `A` agent picker or as `default_agent` in config/Settings.
- `internal/agent`: new `Cursor` type with binary name, display name, and launch/resume command (`cursor-agent` / `cursor-agent --resume <id>`). No fork primitive exists in `cursor-agent`, so fork-to-worktree stays Claude-only (existing guard covers it automatically).
- `internal/hooks/cursor_hooks.go` (new): installs fleet's hook command into `~/.cursor/hooks.json`, only when `cursor-agent` is on PATH. Cursor's hooks.json schema is flatter than Claude/Codex's nested `{matcher, hooks:[...]}` shape, so this is a small dedicated injector rather than a reuse of the Codex one — it does reuse the agent-neutral marker/path helpers (`isFleetHook`, `GetHookCommand`, `FleetBinaryPath`).
- Status detection is pure hook-driven (same shape as OpenCode's branch in `UpdateStatus`) — no pane-scraping. Cursor has no dedicated permission/approval hook, so `beforeShellExecution`→waiting / `afterShellExecution`→running bracket the interactive approval prompt instead.
- No dir-trust seeding needed (unlike Codex): Cursor's permission model is a global allowlist (`~/.cursor/cli-config.json`), not a per-directory trust flag.
- No known local source for reading a Cursor chat's title, so naming falls back to fleet's prompt heuristic (same as OpenCode today).
- Sidebar gets a new `✦` glyph for Cursor sessions; diagnostics reports `cursor-agent --version`.
- `CLAUDE.md` and a changelog fragment (`changelog/unreleased/cursor-agent.md`, not flagged as a highlight) are updated accordingly.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist

- [ ] Linter passes (`make lint`) — `golangci-lint` isn't installed in the environment this was built in; `go vet ./...` and `go fmt ./...` are clean
- [x] Documentation updated (`CLAUDE.md`)
- [x] Changelog fragment added in `changelog/unreleased/`

## Notes for reviewers

- The `beforeShellExecution`/`afterShellExecution` → waiting/running status mapping is inferred from Cursor's public docs, not observed against a live approval prompt — worth a manual smoke test with a real `cursor-agent` session before merging.
- Binary name is `cursor-agent` (confirmed locally; Cursor's installer also symlinks it as the more generic `agent`, which was deliberately avoided to reduce PATH-collision risk).

## Summary by CodeRabbit

- **New Features**
  - Added Cursor CLI as a selectable session agent with default-agent selection, sidebar labeling, and resume support.
  - Added automatic Cursor hook installation and updates while preserving existing settings.
  - Diagnostics now display the Cursor CLI version when detected.
- **Bug Fixes**
  - Improved Cursor status tracking and idle-state handling when hooks are unavailable.
- **Documentation**
  - Updated documentation to include Cursor support and capabilities.
- **Tests**
  - Expanded coverage for Cursor sessions, status mapping, and hook installation.


internal/hooks/cursor_hooks.go
```diff
@@ -0,0 +1,225 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/brizzai/fleet/internal/debuglog"
+)
+
+// cursorHookEvents lists the Cursor CLI hook events fleet subscribes to.
+// Cursor's payload field names match Claude's (hook_event_name, session_id,
```
hayke102, line 14: **This premise looks wrong, and most of the PR rests on it.** Cursor's hook payloads identify the chat with `conversation_id`, not `session_id` — `session_id` appears only on the session-lifecycle events, if at all.

If that's right, the chain is:

1. `hookPayload` (hook_handler.go:17) decodes only `` `json:"session_id"` `` — no alias, no fallback → `SessionID` is `""` on every Cursor hook.
2. `UpdateHookStatus` skips the whole ownership/adoption block on an empty id (session.go:425, 499, 530) → `s.ClaudeSessionID` stays empty forever.
3. `BuildLaunchCmd`'s `if o.ResumeID != ""` never fires → always bare `cursor-agent`.

So the `cursor-agent --resume <id>` branch this PR adds is dead code, and **`r` restart, Reload All, and idle-suspend→resume all silently open a brand-new empty Cursor chat**. That last one is the worst: `Suspend()`'s doc comment and CLAUDE.md both promise "the conversation survives — `ClaudeSessionID` is persisted", and the user only finds out after Enter.

Worth settling with a real payload before merge — dump one `beforeSubmitPrompt` and one `stop` from a live `cursor-agent` and check the key names. If it is `conversation_id`, the fix is small (accept both in `hookPayload`), but it needs to land with this PR rather than after.

internal/hooks/cursor_hooks.go
```diff
@@ -0,0 +1,225 @@
...
+			return false, fmt.Errorf("parse %q entries (refusing to overwrite user hooks): %w", event, err)
+		}
+		events[event] = merged
+	}
+
+	eventsRaw, err := json.Marshal(events)
+	if err != nil {
+		return false, fmt.Errorf("marshal hooks: %w", err)
+	}
+	root["hooks"] = eventsRaw
```
hayke102, line 111: **Nil-map panic that kills fleet at startup.** If `~/.cursor/hooks.json` holds a top-level `null`, `json.Unmarshal` at line 68 leaves `root` nil and returns **no error** — so the guard passes and this line panics with `assignment to entry in nil map`.

Reproduced:

```go
os.WriteFile(dir+"/hooks.json", []byte("null"), 0644)
InjectCursorHooks(dir)
// panic: assignment to entry in nil map ... cursor_hooks.go:104
```

`InjectCursorHooks` runs from `loadSessions`' tea.Cmd goroutine, so Bubble Tea's panic handler restores the terminal and exits — fleet dies on launch with a stack trace and the user can't reach the TUI to fix the file.

Lines 84-89 already handle exactly this hazard one level down (`"hooks": null`), with a test. The `else` at line 67 needs the same treatment:

```go
} else {
	if err := json.Unmarshal(orig, &root); err != nil {
		...
	}
	if root == nil {
		root = make(map[string]json.RawMessage)
	}
}
```

cmd/fleet/hook_handler.go
```diff
@@ -62,11 +63,44 @@ func mapEventToStatus(event string) string {
...
+	// No sessionStart case: unlike Claude, Cursor's initial status (see
+	// initialRunStatus in session.go) starts idle, not running — so there's
+	// nothing for sessionStart to correct, and mapping it to "finished" (as
+	// Claude's SessionStart does) would immediately flip a freshly launched,
+	// untouched session to finished before any turn ran. It falls through to
+	// the default unmapped case below; fleet subscribes to no such hook (see
+	// cursorHookEvents in internal/hooks/cursor_hooks.go).
+	case "beforeSubmitPrompt":
+		return "running"
+	case "beforeShellExecution":
```
hayke102: **The comment above says "waiting is only wrong for auto-approved commands, which resolve to running again almost immediately" — that isn't right.** `afterShellExecution` fires when the command *completes*, not when the approval clears. So an auto-approved `npm test` that takes 90s shows ◐ **waiting for the full 90s** while the agent is actually working. A denied command may never fire `afterShellExecution` at all, pinning it indefinitely.

And Cursor is on the pure-hook `UpdateStatus` branch (session.go:770) with no pane fallback, so nothing corrects it.

Two things consume that wrong state:

- `Space` rotates to it as if it needs attention.
- `quickApproveSelected` (app.go:4243) gates only on `StatusWaiting` + `IsAlive()` — **no agent gate**. Pressing `Y` sends `y` then Enter. With no approval prompt on screen, that `y` is submitted as a new user message mid-turn.

The PR description is candid that this mapping was inferred from docs rather than observed against a live prompt. Given the `Y` blast radius, I'd want it observed before merge — and if Cursor really has no distinguishable approval signal, mapping `beforeShellExecution`→`running` is the safe default (loses the waiting state, never fabricates one).

cmd/fleet/hook_handler.go
```diff
@@ -62,11 +63,44 @@ func mapEventToStatus(event string) string {
...
+		return "waiting"
+	case "afterShellExecution":
+		return "running"
+	case "stop":
```
hayke102, line 89: Cursor's `stop` payload carries `status: "completed" | "aborted" | "error"`, but this maps all three to `finished` unconditionally — no field of the payload is read.

So a turn that died on a model error or was aborted renders with the ● "done, come look" dot, and `Space` presents it alongside genuinely completed work. The user discovers it failed only after attaching.

OpenCode's adapter already distinguishes this (`session.error` → `"error"`), and `applyHookStatus` has a `case "error"` arm ready — it's just unreachable for Cursor. Reading `status` here would wire it up.

cmd/fleet/hook_handler.go
```diff
@@ -62,11 +63,44 @@ func mapEventToStatus(event string) string {
...
+		return "finished"
+	case "sessionEnd":
```
hayke102, line 98: `sessionEnd` → `"dead"` lands in `applyHookStatus`'s `case "dead"` (session.go:857 / 922), which sets `StatusError` **unconditionally** and triggers a crash dump — with no pane corroboration.

But Cursor fires `sessionEnd` when a *conversation* ends (reason `completed` / `aborted` / `user_close`), which isn't the same as the process exiting. The tmux liveness checks above have already passed at that point, so the pane is alive and sitting at its prompt.

Result: a healthy session shows a red ✕ error until the next prompt, writes a spurious crash dump, and gets swept into **Reload All Sessions** — which restarts dead/error sessions, killing a live agent.

OpenCode is the only other agent on this branch and never emits `"dead"`, so this arm was effectively unreachable before now. If Cursor's `sessionEnd` carries a reason, gate on it; otherwise a pane-death check before `StatusError` would be safer.

cmd/fleet/hook_handler.go
```diff
@@ -62,11 +63,44 @@ func mapEventToStatus(event string) string {
...
 		return ""
 	}
 }
 
+// isPromptSubmit reports whether event is a user-prompt-submission hook —
+// Claude/Codex's UserPromptSubmit, or Cursor's beforeSubmitPrompt equivalent
+// (see internal/hooks/cursor_hooks.go) — used to gate prompt-text capture and
+// prompt-count increments in handleHookHandler.
+func isPromptSubmit(event string) bool {
+	return event == "UserPromptSubmit" || event == "beforeSubmitPrompt"
```
hayke102, line 117: `isPromptSubmit` covers the two call sites in this file, but `emitHookMetrics` (internal/hooks/hook_watcher.go:258) still switches on the raw event and matches only the PascalCase `"UserPromptSubmit"` / `"Stop"`.

So for a Cursor session, `EventClaudePromptSubmitted`, `EventClaudeResponseReceived` and `MarkOnboardingMilestone(MilestoneFirstClaudeResponse)` never fire — a Cursor-only user registers in the dashboard as having installed fleet and never gotten a single agent response, which corrupts the activation funnel.

To be fair this gap is **pre-existing** — OpenCode's `session.status`/`session.idle` miss it too, so it's not introduced here. But this PR widens it, and now that `isPromptSubmit` exists the fix is nearly free: reuse it for the prompt arm and add the response-side names alongside `"Stop"`.

internal/agent/agent.go
```diff
@@ -111,6 +124,13 @@ func (t Type) BuildLaunchCmd(o LaunchOpts) string {
 		}
 	}
 
+	if t == Cursor {
```
hayke102, line 134: This branch reads `o.ResumeID` and silently drops `o.ForkID` — Codex (`codex fork <id>`) and OpenCode (`opencode --session <id> --fork`) both honor it.

The doc comment above says "no fork primitive — fork-to-worktree stays Claude-only", and that's true of **`F`** (fork-to-worktree). But plain **`f`** is a different path: `forkSelected` (app.go:3824) gates only on `ClaudeSessionID != ""`, and the context-menu entry is `Enabled: resumable` (app.go:6944) — **neither checks the agent**. It passes `parentAgent: s.Agent` straight through, so a Cursor fork reaches here with `ForkID` set and gets a bare `cursor-agent`: a new row titled `<title> (fork)` holding an empty conversation, no error.

Coupled with the `conversation_id` issue on cursor_hooks.go:14 — today `ClaudeSessionID` is probably always empty, so `f` fails at the guard with "no Claude conversation ID yet" instead. Fix the id capture and this becomes live. Either way the guard should be agent-aware, so the row dims with a real reason rather than dead-clicking (CLAUDE.md: *"the guards mirror the real handlers exactly, so a lit row can never dead-click"*).

`TestBuildLaunchCmd` adds `cursor new` and `cursor resume` cases but no fork case, so nothing catches this.

internal/session/session.go
```diff
@@ -762,11 +762,16 @@ func (s *Session) UpdateStatus() {
...
 	// hook is always trusted and no pane scraping is needed. Before the first
 	// event there is no hook — the session sits idle at its prompt.
-	if s.Agent == agent.OpenCode {
+	//
+	// Cursor CLI rides the same path: its hooks.json events (mapped in
+	// cmd/fleet/hook_handler.go) report running/waiting/finished/dead directly,
+	// with no pane scraping. Before the first event there is no hook — the
+	// session sits idle at its prompt.
+	if s.Agent == agent.OpenCode || s.Agent == agent.Cursor {
 		if !hasHook {
```
hayke102, line 983: This branch force-sets `StatusIdle` on every tick when `hasHook` is false — fine for OpenCode, but the failure mode is quieter for Cursor because there are more ways to end up with no hook file:

- hooks are injected only from `loadSessions` at startup (app.go:6670), so installing `cursor-agent` **while fleet is running** leaves the session hookless for the rest of the launch;
- if `InjectCursorHooks` returns an error, app.go:6672 only `debuglog.Logger.Error`s it — nothing reaches the UI;
- `handleSessionCreate` checks only `exec.LookPath("cursor-agent")`, so session creation still succeeds.

The user then gets a row that resets to ○ idle every 500ms forever: never Running, never Waiting on an approval, invisible to `Space` and the status pills. No error explains why — unlike a missing `claude` CLI, which surfaces as a `warning` in `loadSessionsMsg`.

Surfacing the injection error as a warning (same as the `claude` path) would make this diagnosable.

internal/ui/app.go
```diff
@@ -6665,6 +6665,13 @@ func (h *Home) loadSessions() tea.Msg {
 			debuglog.Logger.Error("opencode plugin inject failed", "err", err)
 		}
 	}
+	// Install Cursor CLI hooks too, but only if cursor-agent is present — never
+	// create ~/.cursor for users who don't have it.
+	if _, err := exec.LookPath("cursor-agent"); err == nil {
```
hayke102, line 6922: This installs into Cursor's **global** `~/.cursor/hooks.json` — which the Cursor *IDE* also reads, not just `cursor-agent`. And having the IDE installed is exactly what puts `cursor-agent` on PATH, so the `LookPath` guard selects for the affected users rather than against them.

Every agent action in the GUI then forks `fleet hook-handler`, which hits the `instanceID == ""` branch (hook_handler.go:140) and emits `log.Warn("hook-handler: no FLEET_INSTANCE_ID env var", ...)`. Warn is above the default level, so it always writes.

Two consequences:

1. `debug.log` is size-truncated, so a normal editing session evicts fleet's own history — and the **`!` bug-report flow pastes the last 100 lines into a public GitHub issue**. Reports would ship 100 lines of `no FLEET_INSTANCE_ID` instead of diagnostics.
2. `beforeShellExecution` is a blocking hook, so every terminal command in the IDE waits on a fleet process spawn.

At minimum the no-instance-id path should drop to Debug. Better: check whether Cursor supports a CLI-scoped hooks file, or bail out early when `FLEET_INSTANCE_ID` is unset before doing any work.

internal/hooks/cursor_hooks.go
```diff
@@ -0,0 +1,225 @@
...
+		fields["command"] = cmdRaw
+		patched, err := json.Marshal(fields)
+		if err != nil {
+			return nil, err
+		}
+		entries[i] = patched
+		return json.Marshal(entries)
+	}
+
+	newEntry, err := json.Marshal(cursorHookEntry{Command: currentCmd, Type: "command"})
```
hayke102, line 230: `GetHookCommand()` returns `'<path>/fleet' hook-handler --fleet-hook` — the single quotes come from `shellQuote`, and claude_hooks.go documents that contract as *"Claude runs this through a shell"*.

Nothing here establishes that Cursor does the same. If Cursor `exec`s the `command` directly (argv split, no `sh -c`), the leading `'` becomes part of the program name and every invocation fails with ENOENT.

That failure is completely silent: injection itself succeeded, so no error; Cursor is on the pure-hook path with no pane fallback, so `hasHook` never becomes true and every Cursor session sits at ○ idle forever regardless of what the agent does. The only trace is *absence* in `debug.log`.

The new tests assert `strings.Contains(e.Command, "hook-handler")`, which passes with the quotes present — so this wouldn't be caught. Worth confirming against Cursor's hook-execution docs; if it's not shell-invoked, this needs an unquoted variant.

internal/hooks/cursor_hooks.go
```diff
@@ -0,0 +1,225 @@
...
+			return nil, err
+		}
+	}
+
+	currentCmd := GetHookCommand()
+
+	for i, raw := range entries {
+		var probe cursorHookEntryProbe
+		if err := json.Unmarshal(raw, &probe); err != nil {
+			continue // not a command-hook object (e.g. a prompt hook) — leave it untouched
```
hayke102: The comment describes a branch that isn't reached. A prompt-hook entry (`{"type":"prompt","prompt":"..."}`) unmarshals into `cursorHookEntryProbe` **without error** — `Command` just decodes to `""` — so it exits via `!isFleetHook(probe.Command)` on line 194, not this `continue`. Coverage confirms line 192 is never hit by `TestInjectCursorHooksPreservesUnknownFieldsAndPromptHooks`, which is the test that seeds one.

The test passing makes the comment look load-bearing. A maintainer reads *"not a command-hook object (e.g. a prompt hook) — leave it untouched"*, concludes prompt hooks are already filtered here, and treats the line-194 guard as only about *other users' command hooks* — the natural next simplification. Drop or tighten it and the loop claims the first entry of any event as fleet's own and patches its `command` key, rewriting a user's prompt hook into `{"type":"prompt","prompt":"...","command":"'/path/fleet' hook-handler --fleet-hook"}`. The existing test still passes (entry count and `prompt` field both survive).

Either reword to describe what actually happens, or make the probe genuinely reject non-command entries (e.g. required-field check) so the comment becomes true.

CLAUDE.md
```diff
@@ -167,4 +168,4 @@ chrome-extension/                # Chrome MV3 extension (service worker, manifes
 - Chrome extension ID: `haphpcoecelhofejcklinnlbfijgdnih` (stable via `key` in manifest.json)
 - Extension commands: `open_or_focus`, `close_tab`, `create_tab_group`, `ping`
 - Service worker reconnects to native host on disconnect (2s delay)
-- Claude Code + OpenAI Codex + OpenCode, macOS + Linux (see `docs/linux.md` for the platform split)
+- Claude Code + OpenAI Codex + OpenCode + Cursor CLI, macOS + Linux (see `docs/linux.md` for the platform split)
```
hayke102: CLAUDE.md is updated in five places for Cursor, but README.md still says three agents in ten — and it's the one users actually see:

- L8 tagline: *"orchestrating Claude Code, Codex &amp; OpenCode sessions"*
- L85 requirements: the three CLIs as the "at least one"
- L95: `# 'A' — new session, pick the agent (Claude Code / Codex / OpenCode)`
- L102/104 feature heading + body, L134 resume list, L146, L159 comparison row (*"✅ Claude + Codex + OpenCode"*), L164, L166
- L178 keybindings table: `A` → "New session (pick agent: Claude Code / Codex / OpenCode)"

A Cursor CLI user reading the GitHub page concludes fleet doesn't support their agent; a user who presses `A` sees a fourth option the docs never mention. The OpenCode PR updated README in lockstep with CLAUDE.md, so this is drift introduced here rather than pre-existing.

internal/hooks/cursor_hooks.go
hayke102: **Correction — this finding is wrong, and the code here is right.** Apologies for the noise.

I installed `cursor-agent` (`2026.07.23-e383d2b`) and read the shipped bundle. Payload construction, `3143.index.js` → `executeHookForStep`:

```js
const m = d ? void 0 : (h.session_id ?? h.conversation_id),
      y = Object.assign({...t},
                        m !== void 0 && {session_id: m},
                        {hook_event_name: e, cursor_version, workspace_roots, user_email},
                        ...)
```

`t` is the call-site object — for `beforeSubmitPrompt` that's `{conversation_id: xe.getId(), generation_id, model, prompt, attachments}` (`4329.index.js`). The spread keeps `conversation_id`, and `session_id` is then added **alongside** it, populated from `session_id ?? conversation_id`.

So the emitted payload carries **both keys with the same value**, and `hookPayload.SessionID` (`json:"session_id"`) decodes fine. The comment's claim that Cursor's field names match Claude's is accurate, `ClaudeSessionID` does get captured, and `cursor-agent --resume <id>` is reachable. I had this backwards — I inferred `conversation_id` from the docs listing it as a common field, without checking that `session_id` is emitted too.

Resolving. One knock-on worth flagging: since resume IDs *are* captured, the `ForkID` gap I raised on `internal/agent/agent.go:127` is live today rather than latent — I've updated that thread.

internal/hooks/cursor_hooks.go
hayke102: **Correction — this finding is wrong; the quoting is fine.** Cursor does invoke hook commands through a shell.

From the shipped bundle (`cursor-agent 2026.07.23-e383d2b`), `3143.index.js` → `executeCommandScript`:

```js
_ = e ? `@'\n${y}\n'@ | & ${d.command}`
      : `${d.command} <<'CURSOR_HOOK_EOF'\n${y}\nCURSOR_HOOK_EOF`
```

The default transport appends a **heredoc** to the command string to feed it the JSON payload on stdin. `<<'CURSOR_HOOK_EOF'` is shell syntax — it cannot work unless a shell parses the string, so `'<path>/fleet' hook-handler --fleet-hook` is executed correctly and the `shellQuote` contract inherited from Claude holds. (The other branch is PowerShell for Windows, also shell-interpreted.)

My concern was that Cursor might `execFile` the command argv-split; it doesn't. Resolving.

internal/agent/agent.go
hayke102: **Severity update — this is live today, not latent.**

I hedged above that `ClaudeSessionID` is probably always empty for Cursor (per my `cursor_hooks.go:14` thread), so `f` would fail at the "no Claude conversation ID yet" guard rather than silently forking. **That hedge was based on my own mistake** — I've since read the `cursor-agent` bundle and confirmed the hook payload does carry `session_id`, so resume IDs *are* captured. I've retracted that thread.

Which means the path is reachable now: once a Cursor session has fired one hook, `resumable` is true, the context-menu entry lights up, `forkSelected` passes `ForkID` through, and this branch drops it and returns bare `cursor-agent` — a new empty conversation titled `<title> (fork)`, no error.

The rest of the comment stands: the guard should be agent-aware so the row dims with a reason, and `TestBuildLaunchCmd` still has no Cursor fork case.

hayke102:
@VainEyal heads up — I pushed 6 commits to this branch (maintainer edits), so **please review them before merging**. I'd rather you sign off than have me merge my own changes into your PR.

To validate the review I installed `cursor-agent` (`2026.07.23-e383d2b`), read the shipped bundle, and captured live hook payloads. That changed several conclusions — including two of my own review comments, which were wrong.

### Two of my review comments were wrong — retracted

- **`session_id` vs `conversation_id`.** I claimed Cursor keys the thread as `conversation_id` and that `--resume` was therefore dead code. Wrong. `executeHookForStep` spreads the call-site object (which carries `conversation_id`) and then adds `session_id: session_id ?? conversation_id`. A live capture confirms both keys present with identical values. **Your comment was correct.**
- **Shell quoting.** I worried `GetHookCommand()`'s single-quoted path would break if Cursor `execFile`'d the command argv-split. It doesn't — `executeCommandScript` builds ``  `${command} <<'CURSOR_HOOK_EOF'…`  ``, a heredoc, which only works under a shell.

Both threads are retracted and resolved. Sorry for the noise.

### What the 6 commits change

| Commit | |
|---|---|
| `56ffad8` | **Startup panic.** A `hooks.json` holding a bare `null` unmarshals to a nil root map *without* erroring, so `root["hooks"] = …` panicked — and since `InjectCursorHooks` runs from `loadSessions`, that killed the TUI at launch with no way back in. Same guard the events map already had. Plus the stale-path migration test and a comment fix. |
| `e5e2423` | **Shell hooks → running, never waiting.** `afterShellExecution`'s payload carries `output` and `duration` (measured 1171ms on a real command), so it fires at *completion*, not when an approval clears — waiting held the row at ◐ for the whole command. With no pane fallback, `Space` rotated to busy sessions and `Y` would send `y`+Enter into a pane showing no prompt. **`stop.status`** is now read (`completed`/`aborted`/`error` — three literal call sites in the bundle); previously a failed turn rendered identically to a successful one. Also widened `emitHookMetrics`, which matched only Claude's PascalCase names, and dropped the missing-`FLEET_INSTANCE_ID` log to Debug. |
| `4e5aeb3` | **Fork gate.** `f` checked only for a session id, not the agent, so forking a Cursor session dropped `ForkID` and launched an empty conversation. Adds `Type.SupportsFork()` and gates `forkSelected` + the context menu on it. Also surfaces a failed hook install as a startup warning — otherwise every Cursor session silently pins to idle forever. |
| `fe2d5d4` | README's eleven "three agents" mentions. |
| `e9216b4` | **`sessionEnd` is not a death.** Captured payload: `{reason: "completed", final_status, duration_ms, …}` with the process alive at its prompt. Mapping it to `dead` set `StatusError`, wrote a crash dump, and made the session eligible for Reload All Sessions — which restarts error sessions, killing a live agent. Now `finished`; real death comes from tmux pane-death, same as Codex/OpenCode. |
| `1423ddd` | Merge `master` (you were 7 behind). One conflict in `forkSelected` — my new guard vs #229's switch to the mutex-guarded `GetClaudeSessionID()`. Kept both. |

`go build ./...`, `gofmt`, and `go test ./...` all clean on the merge result.

### One thing I did NOT fix — worth your call

**Cursor CLI appears to have a per-directory trust gate**, which contradicts the comment in `cursor_hooks.go`:

> no dir-trust seeding (Cursor's permission model is a global allowlist in `~/.cursor/cli-config.json`, not a per-directory trust flag)

Launching `cursor-agent` in a fresh directory gave me:

```
Cursor Agent can execute code and access files in this directory.
Do you trust the contents of this directory?
  • Run 'agent' interactively to decide
  • Pass --trust, --yolo, or -f if you trust this directory
```

The bundle writes a trust marker (`{trustedAt, workspacePath, trustMethod: "cli-flag"}`) via an agent-store path, but I couldn't find where it persists on disk, and my follow-up probes were muddied by `--print` mode denying shell execution for unrelated reasons — so I stopped rather than guess.

Why it matters here: fleet creates a **new worktree per session**, so each one is a fresh directory. If that prompt appears, the session sits blocked on it — and because Cursor rides the pure-hook path (and I just removed the only waiting mapping), fleet will render it as **idle** rather than needing attention. That's arguably worse than the bug I fixed.

Needs one interactive run in a real tmux pane to settle. If it's real, the fix is probably an `EnsureCursorDirTrust` mirroring `EnsureCodexDirTrust`, or passing `--trust` on launch for fleet-created worktrees (which derive from an already-trusted repo).

### Also worth knowing

Hooks install into the **user-global** `~/.cursor/hooks.json`, which the Cursor **IDE** reads too — so fleet's handler is armed for every editor action, not just fleet sessions. Dropping the log to Debug stops that from poisoning `debug.log` (which the `!` bug-report flow pastes into public issues), but the per-action process spawn is inherent to a shared config file. Might be worth checking whether Cursor supports a CLI-scoped hooks path.

PR #214, opened 2026-07-22
fix(ui): keep status detection alive during an attach

hayke102:
## The bug

Every session's status froze for the entire time you were attached to a session, then caught up all at once on detach.

Reported as "took ~5s to change from waiting to running" — that 5s was the *unjam after detaching*, not a slow detector.

## Root cause

Attaching runs through `tea.Exec` (`app.go:2656`), which **suspends Bubble Tea's event loop** for the duration. And `p.msgs` is unbuffered:

```go
// bubbletea/v2@v2.0.7 tea.go:598
msgs: make(chan Msg),
```

`Program.Send` is a bare rendezvous with only `ctx.Done()` as an escape — no deadline. While suspended, nothing drains it.

The git+PR fan-out pushed each repo's result through `h.send`. During an attach that parked one goroutine per repo, each holding one of the four semaphore slots, wedging the fan-out until detach. Since the fan-out ran **inline in `statusWorkerCycle`**, it took the ~500ms fast pass down with it — and approving a permission fires no hook, so `waiting→running` is pane-detection-only and rides that pass exclusively.

The code comment the design rested on was wrong:

> `// Send is buffered + drained synchronously by Tea's main loop`

It is neither buffered nor drained while attached.

## Evidence

From one user instance:

| Time | Event |
|---|---|
| ~15:31:09 | Attach. Last `detectStatus` line. |
| 15:31:09 → 15:47:18 | **Detection blackout** — attached ~16 min |
| 15:47:18 | Detach |
| 15:47:25 | Cycle completes: `duration_ms=975955` (16m16s) |

Across 18 captured stall dumps:

| Signal | Count |
|---|---|
| Worker blocked in `refreshAllGitAndPR` | 18 / 18 |
| All 4 slots held by goroutines stuck in `h.send` | **16 / 18** |
| Event loop suspended in `tea.Exec` → `waitForDetach` | **16 / 17** |
| Slots held by actual `gh` calls | 2 |

## Changes

- **Decouple** the git+PR fan-out onto its own `gitWorker` goroutine, so a slow cycle can no longer starve the fast pass.
- **Publish via `writeGitInfo`** (lock-free COW, safe from any goroutine) instead of `h.send`. The new `gitRepaintMsg` carries no payload — only a repaint hint — so dropping it costs a repaint, never correctness.
- **Skip `gitWorkerCycle` while attached**: the sidebar isn't on screen, and the detach path (`statusUpdateMsg`) rebuilds it wholesale.
- **Gate the stall watchdog on `isAttaching`**. It had been firing on every attach over 90s — 18 false dumps in a single day of normal use, enough noise to hide a real one.

## Tests

Three AST guards in `worker_cadence_test.go`. Which goroutine a call runs on isn't reachable from a runtime assertion — the damage only appears when a real `tea.Exec` suspends the loop — so these pin the call sites:

- `TestStatusWorkerCycleDoesNotRunGitFanout` — fan-out stays off the fast pass's goroutine
- `TestGitFanoutPublishesWithoutTheUpdateLoop` — results go through `writeGitInfo`, not a payload-carrying message
- `TestWorkerWatchdogIgnoresAttach` — watchdog and `gitWorkerCycle` both consult `isAttaching`

Each guard has a positive arm as well as a negative one, so it can't pass vacuously.

`go test -race ./...` — all 23 packages pass.

## Not addressed

`app.go:1109` also sends `gitInfoUpdateMsg` through the loop and can park a goroutine during an attach. It's a one-shot `tea.Cmd` for a single repo rather than a per-cycle fan-out, so it can't wedge anything — flagged, deliberately out of scope.

## Verification

Build + `-race` across all packages. Not yet exercised against a live fleet; the real check is attaching for a couple of minutes and confirming statuses keep moving.

https://claude.ai/code/session_01Fv8jX7xW1iMZ51dJUWrVtc

## Summary by CodeRabbit

* **Bug Fixes**
  * Session statuses now continue updating while attached, preventing prolonged freezes and applying missed updates immediately after attachment.
  * Git and pull request refreshes no longer block status updates or the interface during attachment.
  * Watchdog alerts no longer report false stalls while attachment is in progress.

* **Tests**
  * Added coverage to verify status and repository refreshes run independently and remain non-blocking.

### DIFF
--- a/changelog/unreleased/status-freeze-during-attach.md
+++ b/changelog/unreleased/status-freeze-during-attach.md
@@ -1,0 +1,5 @@
+---
+type: fixed
+---
+
+**Statuses no longer freeze while attached.** Every session's status stopped updating for as long as you were inside a session and only caught up on detach — one measured freeze lasted 16 minutes, so `Space` came back with nothing waiting. Status detection now runs on its own goroutine and keeps working throughout.
--- a/changelog/unreleased/status-freeze-during-attach.md
+++ b/changelog/unreleased/status-freeze-during-attach.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**Statuses no longer freeze while attached.** Every session's status stopped updating for as long as you were inside a session and only caught up on detach — one measured freeze lasted 16 minutes, so `Space` came back with nothing waiting. Status detection now runs on its own goroutine and keeps working throughout.
+**No more frozen statuses.** Session statuses used to stop updating the whole time you were attached and only catch up once you detached — one measured freeze ran 16 minutes, so `Space` came back with nothing waiting. They now keep updating while you're attached.
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -156,12 +156,16 @@ type (
 	// finishes (or hits its deadline). It dismisses the boot splash and
 	// hands control to the steady-state status worker.
 	bootstrapDoneMsg struct{}
-	// gitInfoUpdateMsg carries a per-repo refresh result from a worker
-	// goroutine back to Update — Update is the sole writer of
-	// h.gitInfoCache, so workers push proposed updates through this msg
-	// instead of taking workerMu and mutating the map directly. nil info
-	// is allowed (signals "no data yet"); Update overwrites whatever was
-	// in the cache for `repo`.
+	// gitInfoUpdateMsg carries a per-repo refresh result to Update, which
+	// applies it via writeGitInfo. nil info is allowed (signals "no data
+	// yet"); Update overwrites whatever was in the cache for `repo`.
+	//
+	// Update is NOT the only writer of h.gitInfoCache: the git+PR fan-out
+	// writes it directly (writeGitInfo is a lock-free COW swap, safe from
+	// any goroutine) because routing payloads through Update blocks on
+	// Tea's unbuffered channel for the length of an attach. Readers that
+	// span several h.gitInfo() loads across one decision must therefore
+	// take a single snapshot and index it — see confirmDeleteOrigin.
 	gitInfoUpdateMsg struct {
 		repo string
 		info *git.RepoInfo
@@ -312,6 +316,16 @@ type Home struct {
 	// responsive (it waits synchronously on the per-cycle git+PR fan-out).
 	lastWorkerCycleNano  atomic.Int64
 	workerCycleStartNano atomic.Int64
+	// Same pair for gitWorker. The git+PR fan-out is the blocking work the
+	// watchdog exists to catch, and it no longer runs on statusWorker — without
+	// its own stamps a permanently wedged gitWorker would leave branch/dirty/PR
+	// frozen while workerHeartbeat() still read healthy off statusWorker's.
+	lastGitCycleNano  atomic.Int64
+	gitCycleStartNano atomic.Int64
+	// attachStartedAt is when the current tea.Exec attach began (0 when not
+	// attached). The watchdog subtracts this window instead of skipping its
+	// check, so an attach can't manufacture a stall — nor hide a real one.
+	attachStartedAt atomic.Int64
 
 	// lastTmuxStatusBar tracks the most recent (status, theme) tuple applied
 	// to each session's tmux status bar so the worker can skip no-op
@@ -413,6 +427,7 @@ type Home struct {
 
 	// Background worker for async status/git/PR updates.
 	statusTrigger         chan struct{} // buffered(1), triggers worker
+	gitTrigger            chan struct{} // buffered(1), triggers gitWorker out of band
 	priorityStatusUpdates chan string   // buffered, session IDs with fresh hook changes — drained before round-robin
 	// workerMu now protects only h.sessions (worker snapshots the slice at
 	// the top of each cycle; Update mutates on add/remove/restore). The
@@ -426,9 +441,12 @@ type Home struct {
 
 	// program is the running tea.Program, injected by cmd/fleet/main.go via
 	// SetProgram before p.Run(). Worker goroutines push state updates back
-	// to Update via h.send(msg) so Update remains the sole writer of model
-	// fields — no lock contracts to honor at call sites. Nil during tests
-	// that drive Update directly; send() is a no-op in that case.
+	// to Update via h.send(msg) so Update remains the writer of model fields
+	// — no lock contracts to honor at call sites. The exception is
+	// gitInfoCache, which the git+PR fan-out writes directly via its atomic
+	// COW swap; h.send blocks for the whole of an attach, which is too long
+	// to hold a fan-out goroutine. Nil during tests that drive Update
+	// directly; send() is a no-op in that case.
 	program *tea.Program
 
 	startTime time.Time // app start time for uptime tracking
@@ -514,6 +532,7 @@ func NewHome(storage *session.StateDB, cfg *config.Config, version string, ident
 		errorHistory:           NewErrorHistory(50),
 		actionLog:              NewActionLog(100),
 		statusTrigger:          make(chan struct{}, 1),
+		gitTrigger:             make(chan struct{}, 1),
 		priorityStatusUpdates:  make(chan string, 256),
 		ctx:                    ctx,
 		cancel:                 cancel,
@@ -588,25 +607,17 @@ func (h *Home) writeGitInfo(mutate func(m map[string]*git.RepoInfo) bool) {
 }
 
 // send pushes a message to the Update loop from a background goroutine.
-// In production h.program is wired by main.go; tests skip that, so we
-// fall back to applying state-mutating messages directly. writeGitInfo
-// is safe to call from any goroutine — it COWs the underlying map and
-// atomic-swaps; the test asserts end state, not the path it took.
+// In production h.program is wired by main.go; tests skip that, leaving this a
+// no-op — every remaining sender already applies its own state and uses the
+// message purely as a repaint/toast hint.
+//
+// Callers on the status or git worker MUST guard this with isAttaching:
+// program.Send is a rendezvous on Tea's unbuffered msgs channel, and tea.Exec
+// suspends the loop that drains it for the whole of an attach.
 func (h *Home) send(msg tea.Msg) {
 	if h.program != nil {
 		h.program.Send(msg)
 		return
-	}
-	switch m := msg.(type) {
-	case gitInfoUpdateMsg:
-		if m.repo != "" {
-			h.writeGitInfo(func(next map[string]*git.RepoInfo) bool {
-				next[m.repo] = m.info
-				return true
-			})
-		}
-	default:
-		_ = m
 	}
 }
 
@@ -776,6 +787,9 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		h.rebuildFlatItems()
 		h.enqueuePriorityUpdates(changed)
 		// Also trigger full background refresh for pane captures, git, etc.
+		// Both workers: git skipped every cycle during the attach we just left,
+		// so branch/dirty/PR are as stale as the attach was long.
+		h.triggerGitRefresh()
 		select {
 		case h.statusTrigger <- struct{}{}:
 		default:
@@ -1115,7 +1129,9 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		refresh := func() tea.Msg {
 			return gitInfoUpdateMsg{repo: repoPath, info: git.RefreshGitInfo(repoPath)}
 		}
-		// Trigger PR refresh for new branch.
+		// Trigger PR refresh for new branch. Git lives on its own worker now, so
+		// this needs gitTrigger — statusTrigger would only wake pane detection.
+		h.triggerGitRefresh()
 		select {
 		case h.statusTrigger <- struct{}{}:
 		default:
@@ -1435,6 +1451,11 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		if !h.booted {
 			h.booted = true
 			go h.statusWorker()
+			// Empty fleet still needs git/PR refresh: this path never emits
+			// bootstrapDoneMsg, so without this gitWorker would never start and
+			// branch/dirty/PR would stay blank for the whole process lifetime.
+			// No bootstrap probe runs here, so the repoLastHotAt reasoning holds.
+			go h.gitWorker()
 		}
 		if len(msg.items) > 0 {
 			analytics.Track(analytics.EventOnboardingFirstLaunch, map[string]interface{}{
@@ -2665,6 +2686,7 @@ func (h *Home) attachSession(s *session.Session) tea.Cmd {
 	}
 
 	h.isAttaching.Store(true)
+	h.attachStartedAt.Store(time.Now().UnixNano())
 	// Record the attached session so the memory-pressure sweep never hibernates
 	// the pane the user is sitting in (LastAccessedAt doesn't refresh mid-attach).
 	// Guarded by workerMu — the worker reads it.
@@ -2677,6 +2699,7 @@ func (h *Home) attachSession(s *session.Session) tea.Cmd {
 		// CRITICAL: Clear isAttaching before returning the message.
 		// Prevents race where View() returns empty string after detach.
 		h.isAttaching.Store(false)
+		h.attachStartedAt.Store(0)
 		h.workerMu.Lock()
 		h.attachedSessionID = ""
 		h.workerMu.Unlock()
@@ -3078,13 +3101,18 @@ func (h *Home) confirmDeleteHeader(item SidebarItem) tea.Cmd {
 // reach every checkout even when the origin group is collapsed — collapsed
 // checkouts aren't present in flatItems.
 func (h *Home) checkoutsForOrigin(origin string) []string {
+	return h.checkoutsForOriginIn(h.gitInfo(), origin)
+}
+
+// checkoutsForOriginIn resolves against a caller-supplied snapshot — see originOfIn.
+func (h *Home) checkoutsForOriginIn(m map[string]*git.RepoInfo, origin string) []string {
 	if origin == "" {
 		return nil
 	}
 	seen := make(map[string]struct{})
 	var out []string
 	add := func(repo string) {
-		if repo == "" || h.originOf(repo) != origin {
+		if repo == "" || originOfIn(m, repo) != origin {
 			return
 		}
 		if _, ok := seen[repo]; ok {
@@ -3114,7 +3142,15 @@ func (h *Home) checkoutsForOrigin(origin string) []string {
 // the origin_delete_removes_worktrees setting (default on); the main repo's
 // folder is always kept.
 func (h *Home) confirmDeleteOrigin(item SidebarItem) tea.Cmd {
-	checkouts := h.checkoutsForOrigin(item.OriginKey)
+	// One git snapshot for the whole decision. The git+PR fan-out swaps this
+	// map from its own goroutine, so re-loading per lookup would let the set of
+	// checkouts and the per-checkout destroy decisions come from two different
+	// generations — the dialog counts N worktrees while a different set gets
+	// `git worktree remove`. Most reachable right after boot or just after a
+	// worktree is created, when OriginKey/IsWorktreeRepo are still settling.
+	gitSnap := h.gitInfo()
+
+	checkouts := h.checkoutsForOriginIn(gitSnap, item.OriginKey)
 	if len(checkouts) == 0 {
 		return nil
 	}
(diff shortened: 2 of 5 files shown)


internal/ui/app.go
```diff
@@ -1399,6 +1416,9 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			h.syncViewport()
 		}
 		go h.statusWorker()
+		// Started here, not earlier: gitWorker and the bootstrap probe both
+		// touch repoLastHotAt unlocked, and bootstrap is done by now.
+		go h.gitWorker()
```
hayke102, line 1437: **`gitWorker` never starts on the empty-fleet boot path.**

It's started only here, in the `bootstrapDoneMsg` handler — but that message is only emitted when `bootstrapRepoSet()` is non-empty. The empty-fleet path (fresh install, or every session deleted + every repo unpinned) takes the `len(repos) == 0` branch in `Init` and boots via `discoveryMsg` at :1437, which starts `statusWorker` alone.

On that path `refreshAllGitAndPR` now has **no periodic caller at all** (the only other call sites are one-shot bootstrap and `branchCheckoutMsg`), for the whole process lifetime: repo headers never show a branch name, the dirty `*` never appears or clears, PR badges never appear. Before this PR the fan-out was inline in `statusWorkerCycle`, which `discoveryMsg` *does* start — so this is a regression, not a pre-existing gap.

Starting it in the `discoveryMsg` branch too should be enough; the `repoLastHotAt` reasoning in the comment above still holds there, since that path has no bootstrap probe to race with.

internal/ui/app.go
```diff
@@ -5282,31 +5302,81 @@ drainPriority:
 	// which is why it lives here on the worker, not the Update loop.
 	h.maybeSuspendIdleSessions(sessions)
```
hayke102, line 5355: **The status worker still makes one blocking `h.send` during an attach** — the exact wedge this PR fixes, just through a different message.

`maybeSuspendIdleSessions` → `h.send(sessionsSuspendedMsg{n: n, auto: true})` at :3639. With `session_suspend_mode` at its default `light`, the ~20s sweep suspends an idle session mid-attach; `tea.Exec` has the loop suspended, nothing drains `p.msgs`, and this goroutine parks inside `statusWorkerCycle` until detach. Every session's status freezes — the changelog symptom ("`Space` came back with nothing waiting") reproduces verbatim.

Neither new test covers it: the `isAttaching` gates landed only in `gitWorkerCycle` and `workerWatchdog`. Same treatment as the fan-out would close it — gate the send, or make `sessionsSuspendedMsg` a data-free repaint hint like `gitRepaintMsg` (the suspend already mutated `Status` under the worker's own lock, so the message carries nothing the tick can't re-derive).

internal/ui/app.go
```diff
@@ -5575,11 +5653,34 @@ func (h *Home) refreshAllGitAndPR(repos []string, maxParallel int, deadline time
...
+					return true
+				})
+
+				// Update still owns the sidebar rebuild. The message carries no
+				// data — the cache write above already published it — so if this
+				// is dropped or delayed the cost is a late repaint, never stale
+				// state. Skipped while attached: the sidebar isn't on screen, and
+				// the detach path rebuilds it wholesale.
+				if !h.isAttaching.Load() {
+					h.send(gitRepaintMsg{structural: structural})
```
hayke102: **Check-then-act:** the `Load()` and the `Send` aren't atomic with respect to `isAttaching`.

The tick fires a few ms before the user presses Enter. The top-level gate at :5351 passes, all four in-flight repo goroutines read `isAttaching == false` here, then `attachSession` stores `true` and `tea.Exec` suspends the loop. All four block in `Program.Send`, each still holding a `sem` slot, so `wg.Wait()` at :5698 doesn't return until detach — the same parked-goroutine shape the comment above describes, just through a narrower window.

Secondary cost: `SavePRCacheRow` at :5688 sits *behind* the blocked send, so a fleet killed during a long attach loses those PR cache rows and re-fires `gh` for them next launch. Moving the storage write above the send fixes that half regardless.

The window can't be fully closed while the transport is `program.Send`, but it doesn't need to be: the hint is droppable by construction, and the ~2s tick rebuilds the sidebar anyway.

internal/ui/app.go
```diff
@@ -5575,11 +5653,34 @@ func (h *Home) refreshAllGitAndPR(repos []string, maxParallel int, deadline time
...
+				// This must NOT go through h.send. Bubble Tea's msgs channel
+				// is UNBUFFERED (v2 tea.go: `msgs: make(chan Msg)`) and Send
+				// is a bare rendezvous with no deadline — while tea.Exec has
+				// the loop suspended for an attach, nothing drains it. Sending
+				// here parked one goroutine per repo for the whole attach,
+				// each holding a semaphore slot, wedging the fan-out until
+				// detach. Compare inside the mutation so the before/after read
+				// is atomic with the write.
+				var structural bool
+				h.writeGitInfo(func(next map[string]*git.RepoInfo) bool {
```
hayke102, line 5773: **This makes a worker goroutine the first production writer of `gitInfoCache` outside `Update`.**

The COW swap itself is safe. What breaks is the atomicity that Update-goroutine readers taking *several* `h.gitInfo()` loads across one logical decision were relying on — which held on master, where every production `writeGitInfo` (:1156, :1269, :1386, :1527) ran inside an Update handler.

`confirmDeleteOrigin` (:3116) is the sharp case, and its own comment at :3127 states the dependency: the snapshot exists "so the ~2s gitInfo refresh can't flip a worktree classification between the user ticking the box and execution." It takes a fresh load per repo via `checkoutsForOrigin` → `originOf`, then another per checkout via `repoIsWorktree` (:3135) and `h.gitInfo()[repo]` (:3141). With four fan-out goroutines swapping the map at arbitrary times, a repo whose `OriginKey` resolves `local:<dir>` → `github.com/o/r`, or whose `IsWorktreeRepo` flips false→true on first refresh, can leave `checkouts` and `targets`/`destroyDirs` built from two different generations — the dialog counts N checkouts while a different set gets `git worktree remove`, or one is silently left behind. Most reachable in the first seconds after boot, or right after a worktree is created.

Cheapest fix is one load per decision: `m := h.gitInfo()` at the top of `confirmDeleteOrigin` and index that everywhere below. The render path has the same shape (`h.gitInfo()[item.RepoPath]` per row at :5986 vs `originResolvers` at :6239).

Also worth updating the two doc comments that now assert something false: :160 ("Update is the sole writer of h.gitInfoCache, so workers push proposed updates through this msg instead of … mutating the map directly") and :429.

internal/ui/app.go
```diff
@@ -5282,31 +5302,81 @@ drainPriority:
...
+// gitWorker refreshes per-repo git + PR state on its own goroutine and cadence.
+//
+// This is deliberately NOT part of statusWorkerCycle. The fan-out is unbounded:
+// N repos at 4-wide, each costing a serial git refresh plus up to two 15s `gh`
+// calls, and every result has to reach the Update loop. Sharing a goroutine
+// with the ~500ms fast pass meant a slow fan-out starved the active-session
+// pane re-checks — and a permission grant fires no hook, so waiting→running is
+// pane-only. A 16-minute cycle once froze every session's status for its whole
+// duration.
+func (h *Home) gitWorker() {
```
hayke102, line 5377: **The fan-out moved to this goroutine but the instrumentation didn't follow it.**

`gitWorkerCycle` sets no liveness stamps, and `workerWatchdog` only watches `workerCycleStartNano`/`lastWorkerCycleNano` — both written exclusively by `statusWorkerCycle` (:5101-5105).

`workerWatchdog`'s own doc comment names the failure it exists to catch — *"the worker waits synchronously on its per-cycle git+PR fan-out"* — and that wait now lives here, unwatched. If a `gh`/`git` call outlives its timeout, `refreshAllGitAndPR`'s `wg.Wait()` (deadline=0 path) blocks forever, `gitWorker` never ticks again, and branch/dirty/PR badges freeze permanently — while `workerHeartbeat()` in the bug report looks perfectly healthy because `statusWorker` keeps stamping every 500ms. That's precisely the diagnostic blind spot the 18-minute `gh` wedge motivated adding the watchdog for.

A second pair of stamps for this cycle would restore coverage. `workerStallThreshold`'s 90s doc comment is also now sized around work that no longer runs on the worker it guards.

internal/ui/app.go
```diff
@@ -5335,6 +5405,14 @@ func (h *Home) workerWatchdog() {
 		case <-ticker.C:
 		}
 
+		// An attach suspends the Tea loop for as long as the user is in the
+		// session, so any worker latency measured across one says nothing about
+		// a wedge. Dumping there produced a stall report for every attach over
+		// 90s — normal usage, buried in false positives.
+		if h.isAttaching.Load() {
```
hayke102: This `continue` runs before the stall computation, so the watchdog does **no checking at all** for the duration of an attach — not just no false positives.

A genuine `statusWorker` wedge that starts 30s into a 20-minute attach (say a tmux capture against the attached pane) freezes every session's status — the very symptom this PR is about — and the dump lands only after detach, with the goroutine state from the window that mattered already gone.

The false-positive problem is real and worth fixing; the version that keeps coverage is subtracting the attach window from the measured latency — stamp `attachStartedAt` on the way in and re-base `last`/`start` against it — rather than skipping the tick.

internal/ui/app.go
```diff
@@ -5282,31 +5302,81 @@ drainPriority:
...
+// duration.
+func (h *Home) gitWorker() {
+	ticker := time.NewTicker(tickInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-h.ctx.Done():
+			return
+		case <-ticker.C:
```
hayke102, line 5386: **`statusTrigger` no longer reaches git at all.** This select listens only to its own ticker, and `statusTrigger` has exactly one consumer — `statusWorker` at :4922 — which no longer touches git.

So both explicit "refresh git now" triggers are now no-ops:

- `branchCheckoutMsg` (:1118-1122), pushed under the comment *"Trigger PR refresh for new branch"*
- `statusUpdateMsg` (:779-783), *"Also trigger full background refresh for pane captures, git, etc."* — the detach path

Check out a branch from the dialog while a fan-out is mid-flight (per `workerStallThreshold`'s own comment a single repo can legitimately take ~70s) and the new branch's PR badge waits out that fan-out plus the next tick instead of refreshing immediately. Same for the branch/dirty refresh expected right after detach. A second trigger channel drained here would restore both — at minimum the two call-site comments are now describing something that doesn't happen.

internal/ui/worker_cadence_test.go
```diff
@@ -144,3 +147,131 @@ func TestHeavyCycleDue(t *testing.T) {
...
+// So the fan-out must publish through writeGitInfo (a lock-free COW swap, safe
+// from any goroutine) and treat the message as a repaint hint only.
+func TestGitFanoutPublishesWithoutTheUpdateLoop(t *testing.T) {
+	got := callees(t, "refreshAllGitAndPR")
+	if !got["writeGitInfo"] {
+		t.Error("refreshAllGitAndPR no longer calls writeGitInfo — results are " +
+			"presumably routed through h.send again, which blocks for the whole " +
+			"attach on Tea's unbuffered channel")
+	}
+	if got["gitInfoUpdateMsg"] {
```
hayke102: **This assertion can never fire.** `callees` records names off `*ast.CallExpr` only (`case *ast.Ident` / `case *ast.SelectorExpr` on `call.Fun`); `gitInfoUpdateMsg{...}` passed to `h.send` is an `*ast.CompositeLit` **argument**, never a call.

Re-add `h.send(gitInfoUpdateMsg{repo: r, info: info})` next to the `writeGitInfo` call and `callees("refreshAllGitAndPR")` returns `map[send:true writeGitInfo:true]` — the first assertion is still green and this one is still false. The exact regression the test names goes back into production with the suite passing.

Matching on composite-literal type names in `callees`, or asserting `!got["send"]` for this function, would give it teeth.

internal/ui/worker_cadence_test.go
```diff
@@ -144,3 +147,131 @@ func TestHeavyCycleDue(t *testing.T) {
...
+	}
+}
+
+// TestWorkerWatchdogIgnoresAttach: an attach suspends the Tea loop for as long
+// as the user is in the session, so worker latency measured across one says
+// nothing about a wedge. Without this gate the watchdog fired on every attach
+// longer than workerStallThreshold — 18 stall dumps in a single day of normal
+// use, which is enough noise to hide a real one.
+func TestWorkerWatchdogIgnoresAttach(t *testing.T) {
+	if !mentions(t, "workerWatchdog", "isAttaching") {
```
hayke102: `mentions` returns true on the first `*ast.Ident` named `isAttaching` anywhere in the body, so this passes for an inverted, dead, or no-op reference — it pins that the identifier is *present*, not that the gate is right.

Verified by mutation, both assertions:

- invert :5412 to `if !h.isAttaching.Load() { continue }` — watchdog now dumps on every non-attached tick and goes silent during the one case it was gated for — **test passes**
- invert :5351 to `if !h.isAttaching.Load() { return }` — fan-out runs *only* during attaches, i.e. never usefully — **test passes**
- even `_ = h.isAttaching` satisfies both

Neither regression in the doc comment is actually guarded. A behavioral test is cheap here: set `isAttaching`, run one `gitWorkerCycle`, assert the fan-out didn't fire.

internal/ui/app.go
```diff
@@ -5575,11 +5653,34 @@ func (h *Home) refreshAllGitAndPR(repos []string, maxParallel int, deadline time
 					}
 				}
 
-				// Push the refresh result to Update — Update is the sole
-				// writer of h.gitInfoCache. Send is buffered + drained
-				// synchronously by Tea's main loop, so the next worker
-				// cycle's carry-forward read sees this update.
-				h.send(gitInfoUpdateMsg{repo: r, info: info})
+				// Publish the result by writing the cache directly. It is a
```
hayke102, line 5759: Small follow-on: replacing this send made `send()`'s `case gitInfoUpdateMsg:` branch at :601 unreachable. `grep 'h\.send(' internal/ui` now returns only `sessionsSuspendedMsg` (:3639), `gitRepaintMsg` (:5682) and three drawer messages.

That branch existed purely so the test-mode path (`h.program == nil`) of *this* fan-out still populated the cache — which the direct `writeGitInfo` below now does for real (`TestRefreshAllGitAndPR_AllReposPopulated`, app_test.go:160, goes through the real write). The remaining `gitInfoUpdateMsg` producer at :1116 is a `tea.Cmd` return value, not an `h.send`, so it doesn't keep it alive.

changelog/unreleased/status-freeze-during-attach.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Statuses no longer freeze while attached.** Every session's status stopped updating for as long as you were inside a session and only caught up on detach — one measured freeze lasted 16 minutes, so `Space` came back with nothing waiting. Status detection now runs on its own goroutine and keeps working throughout.
```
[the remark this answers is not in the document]
hayke102: Fixed — headline down to 4 words (`**No more frozen statuses.**`) and the goroutine detail is gone. Kept the 16-minute number rather than taking the suggestion verbatim, since the guide also asks for numbers over adjectives.

internal/ui/app.go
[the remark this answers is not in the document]
hayke102: Fixed — `go h.gitWorker()` added to the `discoveryMsg` branch. Confirmed the regression: `app.go:1574` takes the `len(repos) == 0` path, which never emits `bootstrapDoneMsg`, so `refreshAllGitAndPR` had no periodic caller at all on a fresh install.

hayke102: Fixed — started from `discoveryMsg` too. Your `repoLastHotAt` note is right: that branch never calls `bootstrapGitInfo`, so there's no probe to race with.

internal/ui/app.go
```diff
@@ -5575,11 +5653,34 @@ func (h *Home) refreshAllGitAndPR(repos []string, maxParallel int, deadline time
...
+				})
+
+				// Update still owns the sidebar rebuild. The message carries no
+				// data — the cache write above already published it — so if this
+				// is dropped or delayed the cost is a late repaint, never stale
+				// state. Skipped while attached: the sidebar isn't on screen, and
+				// the detach path rebuilds it wholesale.
+				if !h.isAttaching.Load() {
+					h.send(gitRepaintMsg{structural: structural})
+				}
```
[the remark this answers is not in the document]
hayke102: Partially fixed, and the remainder is deliberate. `SavePRCacheRow` now runs *before* the send. The check-then-send window is left open: closing it needs a transport other than `program.Send`, and the cost is bounded — this is the decoupled git worker, so status detection is unaffected, and the hint is droppable by construction (the cache write above already published the data). Documented inline rather than papered over.

internal/ui/app.go
hayke102: Fixed the half that matters — `SavePRCacheRow` moved above the send, so a fleet killed mid-attach keeps its PR cache rows instead of re-firing `gh` for every repo next launch. Agreed on the rest: the window stays open, for the reasons you gave. Comment updated to say so explicitly.

internal/ui/worker_cadence_test.go
```diff
@@ -144,3 +147,131 @@ func TestHeavyCycleDue(t *testing.T) {
...
+		t.Error("refreshAllGitAndPR no longer calls writeGitInfo — results are " +
+			"presumably routed through h.send again, which blocks for the whole " +
+			"attach on Tea's unbuffered channel")
+	}
+	if got["gitInfoUpdateMsg"] {
+		t.Error("refreshAllGitAndPR sends gitInfoUpdateMsg — that carries the payload " +
+			"through the Update loop, reintroducing the blocking rendezvous. Write the " +
+			"cache directly and send gitRepaintMsg as a hint instead.")
+	}
+}
```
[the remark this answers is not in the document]
hayke102: Fixed — switched to `mentions()`. You were right, and I verified it empirically before and after: `callees(refreshAllGitAndPR)["gitInfoUpdateMsg"]` was `false` even with the send present. Mutation-tested the fix — re-adding `h.send(gitInfoUpdateMsg{...})` now fails the test.

internal/ui/worker_cadence_test.go
hayke102: Fixed via `mentions()`. Note `!got["send"]` wouldn't have worked as written — `h.send(gitRepaintMsg{…})` is a legitimate remaining call, so `send` is `true` today and that assertion would fail immediately. Mutation-verified the `mentions` version goes red on exactly the revert you describe.

internal/ui/app.go
hayke102: Fixed — this was the real miss. Confirmed `:3639` sits inside `maybeSuspendIdleSessions` (`:3550`) on the status worker, so the original freeze reproduced verbatim through a different message. Gated on `isAttaching`, taking your framing: `Suspend()` already wrote `Status` under `workerMu`, so skipping costs only the toast and the tick re-derives the sidebar.

internal/ui/app.go
hayke102: Fixed — `confirmDeleteOrigin` now takes one `h.gitInfo()` snapshot and indexes it throughout, via new `originOfIn`/`repoIsWorktreeIn`/`checkoutsForOriginIn` helpers (the plain versions delegate with a fresh load). Both doc comments corrected: `:159-160` now states the fan-out writes the cache directly and that multi-load readers must snapshot; `:429` likewise. Left the render path alone — it renders per row rather than deciding across rows, so a straddled refresh costs a frame, not a wrong `git worktree remove`.

internal/ui/app.go
hayke102: Fixed — `gitWorkerCycle` now stamps its own `gitCycleStartNano`/`lastGitCycleNano` pair, `workerWatchdog` loops over both workers (dumping per-worker with the name in the log line), and `workerHeartbeat`/snapshots surface `git_stalled` + `git_cycle_in_flight_for` separately. Your point about the heartbeat reading *healthy* during a git wedge was the deciding one — that's worse than no coverage.

internal/ui/app.go
hayke102: Fixed your way — `attachAdjustedStall` subtracts the attach window instead of skipping the tick, with `attachStartedAt` stamped at both attach sites (session and drawer shell). Added a table test for the arithmetic, including the case the plain skip got wrong: a 95s stall with 20s of attach stays a measurable 75s rather than vanishing.

internal/ui/app.go
hayke102: Fixed — added a buffered(1) `gitTrigger` drained in `gitWorker`'s select, plus `triggerGitRefresh()` wired into both call sites you named (`branchCheckoutMsg` and the detach path). Both comments now describe something that happens. The detach case matters more than I'd realised, since `gitWorkerCycle` skips entirely during an attach — branch/dirty was as stale as the attach was long.

internal/ui/worker_cadence_test.go
hayke102: Fixed — added `TestGitWorkerCycleSkipsWhileAttaching`, a behavioral test with both arms (attached ⇒ cache stays empty; detached ⇒ cache populates, so it can't pass on a cycle that never runs). Ran your mutations against it: inverting the gate fails, and deleting it fails. The watchdog arm now pins `attachAdjustedStall` and the git stamps by name, backed by `TestAttachAdjustedStall` for the polarity.

internal/ui/app.go
hayke102: Fixed — deleted the branch. Confirmed the type is still live at `:1116` as a `tea.Cmd` return, so only the `send()` case went. `send()`'s doc comment now says what it actually does (no-op in tests) and warns that worker callers must guard on `isAttaching`.

Issue #157, opened 2026-06-29
sessions shows completed when subagents are still running


1jonahg1:
### Description
sessions shows completed when subagents are still running

### Steps to Reproduce (last 20 actions)
| Time | Action | Detail | Result |
|------|--------|--------|--------|
| 12:20:08 | open bug report |  | ok |

### Diagnostics
- **Version**: 2.12.1
- **macOS**: 26.5.1 (arm64)
- **tmux**: tmux 3.6a
- **Claude CLI**: 2.1.195 (Claude Code)
- **Codex CLI**: codex-cli 0.130.0
- **gh CLI**: gh version 2.89.0 (2026-03-26)
- **Sessions**: 14

### Terminal Environment
- **TERM**: `xterm-256color`
- **Terminal**: iTerm.app 3.6.11
- **COLORTERM**: truecolor
- **TUI size**: 253x63
- **LANG**: en_US.UTF-8
- **tmux default-terminal**: `tmux-256color`
- **tmux mouse**: off

<details><summary>Debug Log (last 100 lines)</summary>

```
time=2026-06-29T12:07:54.468+03:00 level=INFO msg="hook-handler: writing status" instance=21297a7f-1782664159 event=UserPromptSubmit status=running claudeSession=f8c50dcf-d690-43e5-8662-ead444ef5b30
time=2026-06-29T12:07:58.286+03:00 level=INFO msg="status changed (hook)" session=21297a7f-1782664159 title=front-end-code old=idle new=running hookStatus=running hookAge=4.286s
time=2026-06-29T12:09:05.995+03:00 level=INFO msg="hook-handler: writing status" instance=5d98b845-1782655929 event=UserPromptSubmit status=running claudeSession=d8aaa449-b144-49e2-8c82-051670b2c798
time=2026-06-29T12:09:28.757+03:00 level=INFO msg="hook-handler: writing status" instance=5d98b845-1782655929 event=UserPromptSubmit status=running claudeSession=d8aaa449-b144-49e2-8c82-051670b2c798
time=2026-06-29T12:09:30.381+03:00 level=INFO msg="status changed (hook)" session=5d98b845-1782655929 title="Investigate BRZ-2126 implementation requirements" old=idle new=running hookStatus=running hookAge=2.375s
[...]

PR #215, opened 2026-07-22
feat(ui): report a wrong status from `!`

hayke102:
## Summary

Lets a user report a status misdetection with the evidence attached, instead of a description we can't act on.

fleet had both halves of this workflow and no bridge between them. `Shift+D` freezes everything needed to diagnose a misdetection; `!` filed an issue carrying none of it. Issue #157 — *"sessions shows completed when subagents are still running"* — came in through `!` with a one-line description, a steps-to-reproduce table whose only row was `open bug report`, and 100 lines of *global* debug log belonging to other sessions. It never says **which session**. The fix had to come from reproducing it in-house.

`!` now asks what's being reported, and each kind files with its own label and a body sized to it:

```
┌ Report ────────────────────────────────────┐
│  What are you reporting?                   │
│  ❯ Wrong status  "4208" shows · idle       │
│    Bug           something else broke      │
│    Feature       an idea or request        │
└────────────────────────────────────────────┘
```

The resulting issue leads with the signals table — which is what #157 needed and never had:

| Signal | Value |
|--------|-------|
| hook status | `waiting` |
| pane detected | `running` |
| TUI showed | `waiting` |
| **mismatch** | **true** |
| worker stalled | true |
| transcript advanced past hook | true |

### Design notes

**The capture fires at the keypress, not at submit.** Status is a moving target: by the time a description is typed the session may have self-corrected, and reading it then would freeze a state nobody is complaining about. The shown status and title are read once, at the same instant.

**The picker is a stage of the existing `BugReportDialog`, not a new dialog** — so it inherits every registration (`modalOpen`, `routeToModal`, `renderBody`, `SetSize`) and adds no integration points.

**The body is two tiers.** The signals table always ships and holds no user content. The screen excerpt and session-filtered debug log ship behind `ctrl+p` — on by default, previewed in the dialog directly above the toggle, so the default is visible rather than silent.

**Privacy is enforced structurally.** The body is built by allowlisted field reads off the snapshot map, never a marshal of it: `snapshot.json` embeds `hook.file_contents.user_prompt`, the reporter's verbatim prompt, which the dialog never shows them. `TestBuildStatusReportBody_NeverLeaksUserPrompt` guards it. The governing rule — *nothing ships that the dialog didn't show or the checkbox didn't name* — is also why opting out drops the **global** debug log the shared diagnostics block would otherwise append, rather than removing one log and smuggling in a wider one.

**Two choices that look arbitrary but aren't.** The expected-status cycler uses `↑↓` rather than the `←→` this codebase gives cyclers elsewhere, because the description input holds focus and `←→` are its caret keys — taking them would cost fixing a typo mid-string. And it starts unset and refuses submit (same rule as `SnoozeDialog`'s unparseable box), because every candidate default is either the disputed status itself or a guess put in the reporter's mouth.

**Sizing.** Real snapshots measure: `snapshot.json` 1.5–2.6 KB, `pane_clean.txt` 2.9–8.5 KB, `debug_tail.txt` 6–17 KB — all well inside GitHub's 65 KB body limit, so no attachment path is needed. The `!` capture skips the frozen transcript copy `Shift+D` keeps (0.5–2.7 MB, and `!` fires one on every press); `readClaudeLogStat` derives the `claude_log` block in a streaming pass either way, so it costs no signal.

### Drive-by fixes

- Index-out-of-range crash the picker had when made visible without `Show`
- Analytics `Cmd` reached back into dialog state a later `Show` would reassign — now built eagerly

## Type of Change

- [x] New feature

## Checklist

- [x] Documentation updated (`CLAUDE.md`)
- [x] Changelog fragment added in `changelog/unreleased/` (`added`, not a highlight)

## Not verified

Dialogs were exercised through their render functions and the body through unit tests, but **the real TUI was not driven end-to-end** — that needs a live tmux session, and doing it against a running fleet would create real sessions in the state DB. Unconfirmed by execution:

- that a `!` capture omits `claude_session.jsonl` on disk while `Shift+D` still writes one
- the async `reportSnapshotMsg` → `SetSnapshot` path under real timing

**No test submission was made** — that would file a real public issue. To exercise the live `gh` path, point `--repo` at a scratch repo locally.

## Deliberately out of scope

- No clipboard/`--web` fallback when `gh` is missing; the dead-end stands
- No timeout on `gh issue create` (`bugreport.go`), despite being a plausible cause of open #160 *"stuck on creating issue status"* — it's on the path this feature depends on and deserves its own PR
- No `status-detection` label, no `Shift+D` documentation, no snapshot-dir pruning (80 MB across 1,240 dirs locally, essentially all transcript)

https://claude.ai/code/session_01MfX4j6JxDFMwCiT5cQeUTa


internal/ui/bugreport.go
```diff
@@ -76,23 +129,92 @@ func (d *BugReportDialog) SetSize(w, h int) {
 	d.height = h
 }
 
+// SetSnapshot installs an async status capture. Ignored unless the dialog is
+// still open on the invocation that asked for it — a capture landing after the
+// user closed and reopened would describe a session they are no longer
+// reporting on.
+func (d *BugReportDialog) SetSnapshot(snap snapshotResult) {
```
hayke102, line 141: The doc comment above promises the capture is "ignored unless the dialog is still open on **the invocation that asked for it**", but nothing here identifies the invocation — `d.visible` + a non-empty title are both true for *any* open status form.

`!` on session A (capture-pane + transcript scan + goroutine dump + 5 file writes — easily hundreds of ms), `esc`, cursor to B, `!` again: A's capture lands into B's form. The issue is then titled/headed with B's title and shown status, while the signals table, screen excerpt and debug log all come from A — and A's screen content gets published even though this dialog invocation never previewed it.

Cheapest fix: stamp the capture with the session id (or a generation counter bumped in `Show`) and drop it if it doesn't match `d.status`.

internal/ui/bugreport.go
```diff
@@ -101,34 +223,117 @@ func (d *BugReportDialog) Update(msg tea.Msg) (*BugReportDialog, tea.Cmd) {
...
-	if d.renderStats != "" {
-		body += "\n" + d.renderStats
+	var body string
+	label := "bug"
+	if d.kind == kindFeature {
+		// A feature request carries prose and versions only. The error history,
+		// action log, and debug log are reproduction context for a defect; on an
+		// idea they are noise the maintainer has to scroll past.
+		label = "enhancement"
+		body = "## Feature Request\n\n### Problem\n" + description + "\n\n" +
```
hayke102: `description` goes into the body raw — no home-dir sanitize. Both sibling paths do it (`FormatMarkdownWithDesc` and `buildStatusReportBody`), so a feature request is the one report kind that publishes `/Users/<name>/...` verbatim if the reporter mentions a path ("add a way to open /Users/alice/code/foo without leaving fleet").

Same gap at line 291: `truncate(desc, 60)` builds the wrong-status **title** off the raw description while the body below it is sanitized — so the leak survives on the one line GitHub shows in search results.

internal/ui/statusreport.go
```diff
@@ -0,0 +1,457 @@
...
+	if len(lines) > n {
+		lines = lines[len(lines)-n:]
+	}
+	return lines
+}
+
+// submitBlocker names what still stops submission, or "" when ready. The dialog
+// footer renders this, so a disabled Enter always says which field it wants
+// instead of silently doing nothing.
+func (f *statusReportForm) submitBlocker(desc string) string {
```
hayke102, line 128: `submitBlocker` gates on the description and the expected-status pick, but not on `f.captured`. The capture is async, so Enter can land while the form still shows `capturing…` — and `buildStatusReportBody` (line 384) then short-circuits to `_Status snapshot unavailable — the capture failed or the session had no live tmux pane._` with no signals table at all.

That files the exact undiagnosable report this feature exists to prevent, *and* tells the maintainer a false cause (nothing failed — the user was just fast). Worse, the footer said "enter Submit" with no hint the press was premature.

A `case !f.captured && f.snap.err == nil: return "Capturing the session's status…"` arm would reuse the machinery that's already there.

internal/ui/app.go
```diff
@@ -6440,6 +6456,30 @@ func (h *Home) setError(err error) {
...
+	s := h.selectedSession()
+	h.bugReport.Show(h.version, len(h.sessions), h.errorHistory, h.actionLog,
+		h.width, h.height, &h.renderStats, time.Since(h.startTime), s)
+	analytics.Track(analytics.EventBugReportOpened, nil)
+
+	if s == nil {
+		return h, nil
+	}
+	hb := h.workerHeartbeat()
+	return h, func() tea.Msg {
```
hayke102, line 6484: This fires for every `!` press with a session under the cursor — before the user has picked a kind, and regardless of whether they ever submit. `captureStatusSnapshot` isn't read-only: it `MkdirAll`s `~/.config/fleet/snapshots/<ts>_<title>/` and writes pane_raw, pane_clean, debug_tail, snapshot.json and a full-process `runtime.Stack(buf, true)` dump. The `copyTranscript=false` flag only skips the JSONL.

So: press `!`, pick "Feature", file an idea → a multi-hundred-KB snapshot dir is left behind, with the reporter's verbatim prompt inside snapshot.json (`hook.file_contents.user_prompt`). Same for `!` then immediate `esc`. Before this PR only `Shift+D` wrote there, which is what makes "read the newest snapshot" a usable debugging instruction — now the newest dir is likely a cancelled dialog rather than the deliberate `D` capture.

Worth splitting: capture the in-memory evidence for the dialog on `!`, persist the dir only on submit (or only for `kindStatus`).

internal/diagnostics/diagnostics.go
```diff
@@ -234,7 +260,7 @@ func (r *Report) formatMarkdown(description string) string {
 	b.WriteString("\n")
 
 	// Debug logs.
-	if r.RecentLogs != "" {
+	if includeLogs && r.RecentLogs != "" {
```
hayke102, line 264: `includeLogs` gates the debug log but not the Config block right below (line ~270), even though the new doc comment says the function renders "optionally the debug log **and config**".

`buildStatusReportBody` calls `FormatEnvironmentMarkdown(false)` when the reporter unticks `ctrl+p`, so config.json still ships — including `default_project_path` (e.g. `~/work/acme-confidential`) and `editor`. The dialog never showed the config and the checkbox never named it, which is precisely the rule CLAUDE.md states for this flow: *nothing ships that the dialog didn't show or the checkbox didn't name.*

internal/ui/statusreport.go
```diff
@@ -0,0 +1,457 @@
...
+		total := len(f.paneExcerpt(statusReportPaneLines))
+		b.WriteString("  " + checkbox + " " + sectionStyle.Render("include screen + logs") +
+			"    " + dimStyle.Render("ctrl+p"))
+		b.WriteString("\n")
+		for _, ln := range lines {
+			b.WriteString(dimStyle.Render("   │ " + truncate(ln, innerWidth-6)))
+			b.WriteString("\n")
+		}
+		if more := total - len(lines); more > 0 {
+			b.WriteString(dimStyle.Render(fmt.Sprintf("   +%d more lines, plus this session's debug log", more)))
```
hayke102: This note renders unconditionally — `f.includeContent` is consulted only for the `[x]`/`[ ]` glyph above. With the box unticked the reporter sees `[ ] include screen + logs` and, two lines under it, "+34 more lines, plus this session's debug log", which the body will *not* contain.

That's the one ambiguity the preview-above-the-toggle design was built to remove. Guarding the excerpt loop + this note on `f.includeContent` (or dimming them further and saying "not included") keeps the checkbox honest.

internal/ui/statusreport.go
```diff
@@ -0,0 +1,457 @@
...
+		b.WriteString(dimStyle.Render("  capturing…"))
+		b.WriteString("\n")
+	default:
+		lines := f.paneExcerpt(statusReportPreviewLines)
+		total := len(f.paneExcerpt(statusReportPaneLines))
+		b.WriteString("  " + checkbox + " " + sectionStyle.Render("include screen + logs") +
+			"    " + dimStyle.Render("ctrl+p"))
+		b.WriteString("\n")
+		for _, ln := range lines {
+			b.WriteString(dimStyle.Render("   │ " + truncate(ln, innerWidth-6)))
```
hayke102: `truncate` (bugreport.go:573) is byte-based (`s[:maxLen-3]`). Pane lines from Claude/Codex are dense with 3-byte box-drawing runes (`│ ╭ ─ ❯`), so this both cuts at roughly a third of the intended visible width and can split a rune mid-sequence — the excerpt the reporter is asked to approve won't look like their screen.

Same helper is used on the session title (lines ~182, ~246) and on the picker hint at ~204, where a narrow dialog (`dialogWidth` clamps to 40 → `maxLen` 16) can land the cut inside `StatusSymbolRaw`'s `●` and emit invalid UTF-8 into lipgloss.

`ansi.Truncate` is already the codebase's answer here — sidebar.go:689 even documents why ("cell-width-aware and rune/ANSI-safe — it won't split a rune").

CLAUDE.md
```diff
@@ -143,6 +144,10 @@ chrome-extension/                # Chrome MV3 extension (service worker, manifes
...
 - Bug report: `!` key opens dialog showing error history, action log, system diagnostics; `g` opens GitHub issue with pre-filled markdown via `gh issue create --web`
```
hayke102: This bullet is now contradicted by the one added directly beneath it. It claims `!` opens a dialog "showing error history, action log, system diagnostics" and that `g` opens the issue via `gh issue create --web` — there is no `g` key and no `--web` path (`createGitHubIssue` uses `--body-file`), and a feature request deliberately carries no error history or action log.

A future agent reading this file gets two mutually exclusive descriptions of the same key with no way to tell which is current. Delete line 146.

internal/ui/bugreport.go
hayke102: Fixed in 3a58252 — captures now carry a session id (`snapshotResult.sessionID`) and `SetSnapshot` drops any that doesn't match the form's.

You were right that the comment promised something the code never did. Added `TestSetSnapshot_RejectsCaptureForADifferentSession`, which asserts both halves: a capture for A is refused by B's form, and B's own still installs.

internal/ui/app.go
hayke102: Fixed in 3a58252 — went further than either option you offered: `!` now never persists, not even on submit.

The capture takes a `persist` flag and `!` passes false, so it touches the filesystem nowhere — no dir, no `snapshot.json`, and no `writeGoroutineDump`. That last one mattered most: `runtime.Stack(all=true)` pauses every goroutine, which has no business on a hot interactive key.

Persisting on submit was tempting but bought nothing: the issue body is built from the returned values, never read back off disk, so the filed report is byte-identical. The only files that exist solely on disk are `pane_raw.txt` and `goroutines.txt`, neither of which is ever published. So `snapshots/` goes back to being purely your `D` captures and `ls -lt | head` keeps working.

internal/ui/statusreport.go
hayke102: Fixed in 3a58252 — added the `!f.captured && f.snap.err == nil` arm you sketched, returning "Capturing the session's status…".

Also routed the Enter handler through `submitBlocker` instead of its own `expectedStatus()` check, so the key and the footer can't drift apart — that split was how the gap opened in the first place.

Kept the distinction you implied: a capture that genuinely failed (`err != nil`) still submits, since there the "snapshot unavailable" line is true and the reporter's account still has value. Two tests, one per arm.

internal/diagnostics/diagnostics.go
hayke102: Fixed in 3a58252 — Config is gated on `includeLogs` now, so the code does what the comment already claimed.

`default_project_path` being confidential in its own right is the part I'd missed: a client name in a directory name leaks the engagement, not just the layout. Added `TestBuildStatusReportBody_ContentOffAlsoDropsConfig`.

internal/ui/statusreport.go
hayke102: Fixed in 3a58252 — the excerpt loop and the note are both gated on `f.includeContent` now. Unticked renders a single dim `│ not included — signals only` in their place, so the section doesn't collapse and shift the rows below it.

Agreed this was the design defeating itself: the whole argument for on-by-default was that the preview makes the state legible, and a preview that outlives its checkbox makes it less legible than no preview at all. Test asserts the unticked render carries neither the pane text nor the "+N more lines" promise.

internal/ui/statusreport.go
hayke102: Fixed in 3a58252 — `ansi.Truncate` at all four sites this PR added: pane lines, both session titles, and the picker hint.

The narrow-dialog case you spotted was the nastier one — a 40-column clamp putting the cut inside `StatusSymbolRaw`'s `●` and feeding invalid UTF-8 to lipgloss. `TestStatusForm_ExcerptTruncationIsRuneSafe` renders a box-drawing-heavy pane and asserts both `utf8.ValidString` and that the line survives near its intended visible width.

Left the pre-existing `truncate` uses in bugreport.go (error messages, action details) alone as out of scope.

internal/ui/bugreport.go
hayke102: Fixed in 3a58252 — added a `sanitizeHome` helper covering the feature body and both titles, and folded the three inline copies (`formatErrors`, `formatActions`, and the closure in `buildStatusReportBody`) into it.

The title point is the one I'd have kept missing: sanitizing a body while the title publishes the path verbatim is worse than not sanitizing, because it reads as handled. Fixed the `openGitHubIssue` title too — it predates this PR, but it's the same expression and leaving half the leak would have been indefensible.

CLAUDE.md
hayke102: Fixed in 3a58252 — deleted. The `g` key and `--web` claims were already wrong before this PR, but putting a contradicting bullet directly beneath them is what made it actively misleading.

Also refreshed the surrounding bullets for the fixes in this round: the id-matched capture, the single `submitBlocker` gate, `persist=false` on `!`, `ansi.Truncate`, and `sanitizeHome`.

PR #217, opened 2026-07-22
fix(ui): stop the status worker from eating hook transitions

hayke102:
Diagnosed from a `D` snapshot: every signal said the session was running — hook `running`, pane detected `running`, Claude's own status file `busy` with a live pid, transcript advanced 12s past the hook — while the sidebar showed `idle`.

## Root cause

`syncHookStatuses` returns the sessions whose hook changed, but it computes that list by diffing against the session's **in-memory** hook state and then overwrites it. A transition is therefore **consume-once**: whichever caller syncs first eats it.

| Caller | Result |
|---|---|
| `hookChangedMsg` (UI) | → `enqueuePriorityUpdates` ✅ |
| `statusUpdateMsg`, post-detach (UI) | → `enqueuePriorityUpdates` ✅ |
| `statusWorkerCycle`, every ~500ms | **discarded** ❌ |

During an attach `tea.Exec` suspends the Update loop, so `hookChangedMsg` never runs and the worker is the sole observer — it consumed every hook transition and dropped it. `statusUpdateMsg`'s post-detach catch-up sync, which exists precisely to cover the attach window, then found nothing changed and enqueued nothing. Idle sessions aren't in the fast pass (`fastPassSessions` covers Running/Waiting/Starting only), so the session fell back to the round-robin: `ceil(N/statusRoundRobin) * tickInterval` ≈ **26s at 65 sessions**.

## Measured

Same session, same pipeline, 50 seconds apart:

```
17:44:44.145  hook-handler writes status=running     (attached)
17:45:08.457  status changed idle→running   hookAge=24.446s   ← 24.3s
17:45:33.454  hook-handler writes status=waiting     (detached)
17:45:33.564  status changed running→waiting hookAge=564ms    ← 110ms
```

The only differentiator is whether the Update loop was suspended.

## Regression from #214

Before c13286f the worker wedged in `refreshAllGitAndPR` for the duration of an attach, so it never got to consume the change and the detach-time sync caught it. Keeping the worker alive across attaches is right, but it turned that safety net into a guaranteed no-op.

## Fix

Bind the result and merge it into `priorityIDs`, which is built later in the same cycle — no added latency. Merged directly rather than pushed through `priorityStatusUpdates` because that channel drops on full, and a send-then-drain round trip through our own queue buys nothing.

## Test

AST-based, matching the guards #214 added. The damage needs a real attach (Tea suspended, worker still cycling), and `UpdateStatus` needs a live tmux pane to reach `Running` at all — neither is reachable from a `ui` unit test, since `Session.paneCapturer` is package-private to `internal/session`.

Both arms mutation-verified:
- restoring the bare `h.syncHookStatuses(sessions, true)` → red on the discard check
- `_ = hookChanged` → red on the merge check

`go test -race ./internal/ui/ ./internal/session/` passes.

https://claude.ai/code/session_01Fv8jX7xW1iMZ51dJUWrVtc

## Summary by CodeRabbit

* **Bug Fixes**
  * Status rows now update much faster after detaching—typically within about 100 ms instead of up to 25 seconds.
  * Hook status changes are prioritized immediately, reducing delays caused by normal update scheduling.

* **Tests**
  * Added coverage to verify hook status changes are correctly prioritized.

### DIFF
--- a/changelog/unreleased/status-lag-after-detach.md
+++ b/changelog/unreleased/status-lag-after-detach.md
@@ -1,0 +1,5 @@
+---
+type: fixed
+---
+
+**Status updates right after you detach.** Prompting a session and detaching left its row on the old status for up to ~25 seconds. It now updates in ~100ms.
--- a/changelog/unreleased/status-lag-after-detach.md
+++ b/changelog/unreleased/status-lag-after-detach.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**Status updates right after you detach.** Prompting a session and detaching left its row on the old status for up to ~25 seconds. It now updates in ~100ms.
+**Fresh status on detach.** Even after the freeze was fixed, a session you prompted while attached kept its old status for up to ~26 seconds after you detached. It now updates in ~100ms.
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -5240,14 +5240,24 @@ func (h *Home) statusWorkerCycle() {
 	// 3. Sync hook status (fast: in-memory map lookups; worker may resolve a
 	// session-id rotation here — off the UI loop, so its transcript I/O is safe).
 	//
-	// The returned IDs MUST be carried into this cycle's priority set below.
-	// syncHookStatuses diffs against the session's in-memory hook state and then
-	// overwrites it, so a transition is consume-once: whichever caller syncs first
-	// eats it. During an attach tea.Exec suspends the Update loop, so hookChangedMsg
-	// never runs and this worker call is always the first to observe — dropping the
-	// result here left the session on the round-robin (≈26s at 65 sessions), and the
-	// post-detach catch-up sync in statusUpdateMsg found nothing left to enqueue.
+	// The returned IDs MUST be carried into this cycle's priority set, and the merge
+	// below stays adjacent to this call on purpose. syncHookStatuses diffs against
+	// the session's in-memory hook state and then overwrites it, so a transition is
+	// consume-once: whichever caller syncs first eats it. During an attach tea.Exec
+	// suspends the Update loop, so hookChangedMsg never runs and this worker call is
+	// always the first to observe — dropping the result here left the session on the
+	// round-robin (≈26s at 65 sessions), and the post-detach catch-up sync in
+	// statusUpdateMsg found nothing left to enqueue.
+	//
+	// Nothing may come between the consume and the merge: the cycle's defer recover()
+	// swallows a panic from anything in between (step 3b reads transcripts and writes
+	// SQLite), and the transitions this call already ate would be gone for good —
+	// the same failure reached a different way.
 	hookChanged := h.syncHookStatuses(sessions, true)
+	priorityIDs := make(map[string]bool, len(hookChanged))
+	for _, id := range hookChanged {
+		priorityIDs[id] = true
+	}
 
 	// 3b. Auto-name: generate title for ONE session per cycle (heavy cadence).
 	// Priority: manual (R key) > the agent's own title (Claude custom-title, then
@@ -5305,10 +5315,10 @@ func (h *Home) statusWorkerCycle() {
 		}
 	}
 
-	// 4. Priority updates first — sessions whose hook file just changed.
+	// 4. Priority updates first — sessions whose hook just changed, seeded above
+	// from this cycle's own sync and topped up here from the UI path's queue.
 	// These bypass round-robin so the UI reflects fresh hook status within
 	// ~100ms of the hook firing (vs. up to (N/statusRoundRobin)*tickInterval seconds).
-	priorityIDs := make(map[string]bool)
 drainPriority:
 	for {
 		select {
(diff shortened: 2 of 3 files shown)


internal/ui/app.go
```diff
@@ -5310,6 +5318,12 @@ drainPriority:
 			break drainPriority
 		}
 	}
+	// Hooks this cycle's own sync (step 3) observed. Merged directly rather than
+	// pushed through priorityStatusUpdates: the channel drops on full, and a
+	// send-then-drain round trip through our own queue buys nothing.
+	for _, id := range hookChanged {
```
hayke102: The merge itself is right, but it lands these sessions in the priority loop, which runs *before* `seedActiveCaptures`. The priority loop sets `processed[s.ID] = true` (5341), `fastPassSessions` skips anything already processed, so hook-changed sessions never reach the batched `BatchCapturePanes` at 5356 — each one shells out to its own `capture-pane` inside `UpdateStatus`, with `captureCacheTTL` (400ms) already stale relative to the 500ms cycle.

Before this PR those ids were dropped and the sessions rode the fast pass, i.e. one batched capture. Worst case is startup: `HookWatcher.Start` → `loadExisting()` seeds every session's hook file before the first cycle, so the first `syncHookStatuses` reports *every* session as changed and the whole fleet forks a `capture-pane` serially inside one ~500ms cycle. Same shape on any hook burst (`Ctrl+K` → Reload All Sessions fires SessionStart for everything at once).

Cheapest fix: collect the priority sessions into a slice and `h.seedActiveCaptures(prio)` before the loop at 5334, so they read warm caches like the fast pass does.

internal/ui/app.go
```diff
@@ -5239,7 +5239,15 @@ func (h *Home) statusWorkerCycle() {
...
-	h.syncHookStatuses(sessions, true)
+	//
+	// The returned IDs MUST be carried into this cycle's priority set below.
+	// syncHookStatuses diffs against the session's in-memory hook state and then
+	// overwrites it, so a transition is consume-once: whichever caller syncs first
+	// eats it. During an attach tea.Exec suspends the Update loop, so hookChangedMsg
+	// never runs and this worker call is always the first to observe — dropping the
+	// result here left the session on the round-robin (≈26s at 65 sessions), and the
+	// post-detach catch-up sync in statusUpdateMsg found nothing left to enqueue.
+	hookChanged := h.syncHookStatuses(sessions, true)
```
hayke102, line 5256: Consume-once, but the result then travels ~75 lines before it's merged (5324) — through step 3b's transcript reads and SQLite writes. Anything that panics in between is caught by the cycle's `defer recover()`, and the transitions this call already ate are gone for good: `statusUpdateMsg`'s post-detach catch-up sync finds nothing left to enqueue and the session falls back to the round-robin. That's the exact failure this PR removes, just reached a different way.

Hoisting `priorityIDs := make(map[string]bool)` above step 3b and merging right here would keep the consume and the merge adjacent (the channel drain can still happen later into the same map), so the invariant is local instead of spanning the function.

internal/ui/worker_cadence_test.go
```diff
@@ -349,3 +349,101 @@ func mentions(t *testing.T, fnName, ident string) bool {
...
+	// 2. Binding it is not enough: it has to reach the priority set. Without this arm
+	// a log-only use (or `_ = hookChanged`) would satisfy the check above while the
+	// changed sessions still starved on the round-robin.
+	fed := false
+	ast.Inspect(fn, func(n ast.Node) bool {
+		rng, ok := n.(*ast.RangeStmt)
+		if !ok {
+			return true
+		}
+		if id, ok := rng.X.(*ast.Ident); !ok || id.Name != bound {
```
hayke102, line 470: This arm is weak in one direction and brittle in the other.

**Weak** — it only asserts that a `range hookChanged` loop *mentions* `priorityIDs` somewhere in the function: not that it assigns into it, and not that it runs before the loop that consumes `priorityIDs`. Both were mutation-verified on this branch: (a) replacing the body with a read-only `if priorityIDs[id] { … }`, and (b) moving the whole merge *below* the `for _, s := range sessions { if !priorityIDs[s.ID] … }` consumer — each keeps `go test -run TestStatusWorkerFeedsHookChangesIntoPriority` green while fully restoring the ≈26s lag. (b) is the likely one: a future reorder of steps 4/5 reintroduces the shipped bug with a green suite and a test comment claiming it's pinned.

**Brittle** — it pins one syntactic shape, so a behaviour-preserving refactor (`h.enqueuePriority(hookChanged)`, `maps.Copy`, renaming the map) turns CI red with a message asserting the fix was removed, inviting the next person to just weaken the guard.

Asserting an actual assignment into `priorityIDs` plus position (`merge.Pos() < consumingRange.Pos()`) closes the weak side. Minor, same test: `bound` (407-416) is overwritten by each matching assignment, so a second `syncHookStatuses` call here would leave only the last one checked.

changelog/unreleased/status-lag-after-detach.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Status updates right after you detach.** Prompting a session and detaching left its row on the old status for up to ~25 seconds. It now updates in ~100ms.
```
hayke102: Two nits. The hook is six words — CLAUDE.md and `changelog/README.md` both cap it at 2–4 (`**Status updates on detach.**` fits).

And this ships alongside `status-freeze-during-attach.md` ("**No more frozen statuses.** … They now keep updating while you're attached"), so the What's New reel shows two overlapping status-staleness bullets covering the same window — a reader can't tell whether it's two fixes or one restated. Either merge them or name the distinct cause here (the worker eating the hook transition, not the attach itself).

internal/ui/app.go
hayke102: Fixed in 29159d0 — priority sessions are now collected into a `prio` slice and seeded through `h.seedActiveCaptures(prio)` before the update loop, so they read warm caches like the fast pass does.

One correction: the startup burst is not new. `loadExisting()` notifies before the first worker cycle, the UI path syncs and calls `enqueuePriorityUpdates` with all 65 ids (channel cap 256), and the worker drained them into the same priority loop — so that fork storm predates this PR. What this PR newly regressed is the steady state: active sessions with a hook change now leave the batched path. The fix covers both.

internal/ui/app.go
hayke102: Fixed in 29159d0 — `priorityIDs` is now declared and filled immediately after the `syncHookStatuses` call, above step 3b. The channel drain still fills the same map later, so the invariant is local: nothing runs between the consume and the merge. Comment updated to say why nothing may.

internal/ui/worker_cadence_test.go
hayke102: Fixed in 29159d0. Both mutations you named were reproduced red before the change and are now caught:

- read-only body (`if priorityIDs[id]`) → the guard requires an `*ast.AssignStmt` whose LHS is an `*ast.IndexExpr` over `priorityIDs`, not a mention
- merge relocated below the consumer → the guard compares positions, asserting the merge precedes the map’\s first *read* (index reads are distinguished from writes by position set)

Also fixed the `bound` overwrite: a second `syncHookStatuses` binding now fails with a message to extend the guard, rather than silently checking only the last.

On brittleness — I kept the structural assertion but worded the failure around the invariant ("every id must land in priorityIDs before that map is read") and told the reader to update the guard rather than weaken it if the merge moves into a helper or `maps.Copy`.

changelog/unreleased/status-lag-after-detach.md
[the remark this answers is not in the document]
hayke102: Fixed in 29159d0. Headline is now 3 words and the figure is aligned to ~26s:

> **Fresh status on detach.** Even after the freeze was fixed, a session you prompted while attached kept its old status for up to ~26 seconds after you detached. It now updates in ~100ms.

Took the constraints rather than the literal suggestion — "Faster post-detach updates" drops the second person the guideline also asks for, and the rewrite had to carry the de-duplication in the neighbouring thread.

hayke102: Fixed in 29159d0. Went with differentiating rather than merging — `status-freeze-during-attach.md` is #214's artifact and already merged, and the two symptoms genuinely shipped as separate fixes. This fragment now names its own:

> **Fresh status on detach.** Even after the freeze was fixed, a session you prompted while attached kept its old status for up to ~26 seconds after you detached. It now updates in ~100ms.

"Even after the freeze was fixed" makes the pair read as a sequence rather than a restatement, and the headline is down to 3 words.

PR #221, opened 2026-07-28
fix(status): detect activity lines without a token counter

hayke102:
## What

`isWhimsicalActivity` required `tokens` plus `· ↓`/`· ↑`, so activity lines carrying neither were invisible to `detectRunning`:

```
✽ Improvising… (51s · almost done thinking with xhigh effort)
✢ Marinating… (33s)
```

`spinnerChars` covers only part of Claude's rotating glyph set — `·` (U+00B7) and `✻` (U+273B) are absent — so a frame rendering one of those glyphs on a counter-less line matched **neither** signal, fell through to the prompt check, and read as finished mid-turn.

Anchor on the line's *shape* instead: leading glyph → ellipsis-terminated word → parenthesised duration → trailing `)`. Verified against 37/37 real activity lines scraped from `~/.config/fleet/snapshots`.

The leading glyph is a **new** requirement the old matcher didn't have. It's what separates a live indicator from prose closing on a duration, and it matters because `normalizeForHash` shares this function to *strip* lines — over-matching there would blind content-change detection.

I left `spinnerChars` alone deliberately: adding a bare `·` prefix match would false-positive on markdown bullets in scrollback, and the shape check covers every glyph without enumerating a set Claude keeps extending.

## How this surfaced

A `D` snapshot showed `pane_detected: running` / `tui_shows: idle`, `mismatch: true`, on a session that was demonstrably working. The log had it flipping 300ms apart:

```
18:38:01.643  matched spinner char ✽  "✽ Osmosing… (27s · thinking more with xhigh effort)"  → running
18:38:01.946  matched prompt ❯ linesFromBottom=4                                              → finished → idle
```

Same line, glyph rotated off the covered set.

Worth being clear about scope: this flap was the **secondary** cause. The primary one was the hook layer going dead — every hook dropped because its claude `session_id` had been rejected as foreign (a parked-job handoff that `sessionRotationVerdict` can't currently see). That fix is **not** in this PR; it's held pending a second sighting to tell us which signal generalizes. The pane only became the sole authority *because* hooks were down.

## Also: make a silent rejection visible

`case negCached` returned `false` with no log line, so a session whose hook layer had gone dead looked identical in the log to one that was simply quiet. Identifying that state cost a full pipeline trace.

Now logged at INFO, throttled to once a minute per session (the worker re-offers the same hook every ~500ms, so per-attempt would spam):

```
hook dropped: claude session id still rejected id=… owner=0dce251e foreign=bb8feefb
  dropped=running frozenHook=finished frozenHookAge=4h48m36s
```

`frozenHookAge` is the age of the hook being **kept**, not the one dropped — a large value beside a fresh `dropped=running` is the signature. Throttled rather than once-only because `debug.log` rotates: during this investigation it truncated twice inside four minutes, which is exactly how the original rejection line disappeared.

## Behavior change worth a look

`pane_finished_compacting.txt` flips `StatusFinished` → `StatusRunning` (expectation only, fixture file untouched). Its old expectation encoded a *limitation*, not a desired behavior — its own comment said the PreCompact hook was what kept the session running. The pane now matches `· Compacting conversation… (2m 27s)` directly. `debug-status/SKILL.md` updated, since it claimed the pane shares the compaction blind spot.

## Tests

Full suite green with `-race`; `make build` clean.

Every false-positive guard fixture survived unchanged: quoted crashdumps, prose mentioning `· ↓`/`tokens`, CLI spinners in scrollback, quoted dock rows.

One existing test input changed, flagging rather than burying it: `TestDetectStatus/whimsical_pattern` fed a glyphless `Clauding… (53s · ↓ 749 tokens)`, which no real capture looks like. Given a `·` and joined by three new cases — both previously-missed formats plus a prose negative.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented brief “idle” flickers when an activity indicator appears without a token counter.
  * Improved detection of `/compact` sessions so they stay marked as running during active compaction.
  * Expanded recognition of “whimsical” activity indicators, reducing false idling across formatting variants.
* **Documentation**
  * Updated status-detection guidance and added an unreleased changelog entry describing the fixes.
* **Tests**
  * Refreshed and expanded golden and fixture coverage for compaction and whimsical activity detection scenarios.


internal/session/session.go
```diff
@@ -1679,21 +1710,46 @@ func detectRunning(recentLines []string, _ string, log *slog.Logger) Status {
...
+	if !strings.HasSuffix(lower, ")") {
+		return false
+	}
+	// The duration must follow the activity word, not merely appear somewhere on
+	// the line — otherwise prose quoting a timing figure could satisfy the check.
+	ellipsis := strings.Index(lower, "…")
+	if ellipsis < 0 {
+		return false
+	}
+	return hasWhimsicalDuration(lower[ellipsis:])
```
hayke102: **The widened predicate now matches Claude's Bash tool-progress row, which can mask a live permission prompt.**

With `tokens` + `· ↓`/`· ↑` gone, the surviving conditions are: leading non-ASCII byte, trailing `)`, an `…`, and a `(Ns` after it. `⎿  Running… (9s · timeout 5m)` satisfies all four — and that line is not hypothetical, it appears verbatim in your own captures (`~/.config/fleet/snapshots/2026-06-02T16-19-34_cache-stale-while-revalidate-c/pane_clean.txt:44`). Replaying old vs new `isWhimsicalActivity` over the 92 `snapshots/*/pane_raw.txt` files, it is one of exactly 4 new-only matches.

Because `detectRunning` runs before `detectWaiting`, that row outranks an on-screen permission menu. Injecting it into `testdata/pane_waiting_permission_3opt.txt`, `pane_waiting_exitplanmode_approve.txt` and `pane_waiting_askuserquestion_checkbox_focus.txt` at 4, 8 and 14 lines from the bottom flips all three `waiting` → `running` on this branch, and all three stay `waiting` on master.

Downstream that's not just a wrong dot: `applyHookWaiting` sees `paneStatus == Running`, never sets `waitingPaneConfirmed`, and `waitingHookRenderBackstop` then overwrites Status to Running — so `Space` never rotates to the session and it sits unanswered until someone attaches. Parallel/backgrounded bash ticking under a prompt is exactly when this happens.

A cheap fix that keeps the token-counter-less lines working: exclude the tool-result gutter (`⎿`) and `⏺` prefixes, or require the glyph to be a single rune followed by a space + capitalised activity word, rather than accepting any non-ASCII first byte.

internal/session/session.go
```diff
@@ -1679,21 +1710,46 @@ func detectRunning(recentLines []string, _ string, log *slog.Logger) Status {
...
 //
 // Used by detectRunning (status detection) and normalizeForHash (content hashing).
 func isWhimsicalActivity(line string) bool {
-	lower := strings.ToLower(strings.TrimRight(line, " \t"))
-	return strings.HasSuffix(lower, ")") &&
-		strings.Contains(lower, "tokens") &&
-		(strings.Contains(lower, "· ↓") || strings.Contains(lower, "· ↑")) &&
-		hasWhimsicalDuration(lower)
+	trimmed := strings.TrimSpace(line)
+	if trimmed == "" || trimmed[0] < 0x80 {
```
hayke102: **The mandatory leading glyph can miss the frames this change is meant to catch.**

Claude animates the leading cell by cycling glyphs. On any frame where that cell captures as a space — blank phase, a wide glyph split at the pane edge, or a future glyph-less counter form — `TrimSpace` leaves the line starting at `C` of `Clauding`, `trimmed[0] < 0x80` returns false, and detection falls through to the prompt check: mid-turn session reads finished.

That's the same frame-to-frame flicker this commit is fixing, relocated from "glyph missing from `spinnerChars`" to "glyph missing from the frame". Worth noting the old test asserted exactly the glyph-less form (`"Clauding… (53s · ↓ 749 tokens)"`) and was edited to add a `· ` prefix rather than kept alongside — so nothing covers it now.

Not necessarily wrong to require the glyph (it's what fences off prose), but if the `tokens` + `· ↓` anchor were kept as an *alternative* to the glyph, both shapes would match and the widening above would be narrower too.

internal/session/session.go
```diff
@@ -1679,21 +1710,46 @@ func detectRunning(recentLines []string, _ string, log *slog.Logger) Status {
...
+// only surfaces when the hook layer is also down, but then it flips the status
+// frame-to-frame as the glyph rotates.
+//
+// The leading glyph is required (and deliberately not enumerated — Claude keeps
+// adding them): it is what separates a live indicator from prose that happens to
+// end in a parenthesised duration. Every glyph observed so far is non-ASCII, so a
+// non-ASCII first byte stands in for "starts with a glyph, not a sentence".
 //
 // Used by detectRunning (status detection) and normalizeForHash (content hashing).
 func isWhimsicalActivity(line string) bool {
```
hayke102, line 1784: **This branch lacks the `waiting for` exclusion its sibling spinner branch carries.**

`detectRunning`'s spinner loop deliberately skips a spinner-prefixed line containing `waiting for` (session.go:1681-1683) so the agent-team box `✢  Waiting for team lead approval` isn't read as running. `isWhimsicalActivity` has no equivalent, and its glyph gate is satisfied by the box's leading `│` (U+2502).

So a team-waiting line carrying an elapsed counter — `│ ✢ Waiting for team lead approval… (2m 13s)` — matches. `detectRunning` precedes `detectWaiting`'s team-box check (session.go:1823-1830), so the team check never runs: the session shows Running, `Space` skips it, and `Y` refuses to approve it (`app.go:4251`).

I didn't find that exact rendering in the fixtures (the pinned format at session_test.go:66 is boxed and counter-less), so this is the weaker of the two masking cases — but it's the same class as the exclusion that already exists two branches up, and one `strings.Contains(lower, "waiting for")` guard closes it.

internal/session/session.go
```diff
@@ -449,7 +458,23 @@ func (s *Session) UpdateHookStatus(hs *HookStatus, resolveRotation bool) bool {
 				debuglog.Logger.Info("adopting forked claude session",
 					"id", s.ID, "parent", owner, "new", hs.SessionID)
 			case negCached:
-				return false // already rejected this (owner, foreign) pair
+				// Already rejected this (owner, foreign) pair. Re-announce it on a
+				// throttle: this branch is the steady state of a session whose hook
+				// layer has gone dead, and dropping every hook in silence is what
+				// made that state cost a full pipeline trace to identify.
+				if time.Since(rejectLoggedAt) >= rotRejectLogInterval {
```
hayke102: **`rotRejectLoggedAt` is never reset, so this diagnostic can go silent exactly when it's needed — and it's read and written under different locks.**

1. **Not reset on pair change.** `negCacheRotation` (session.go:609-617) overwrites `rotRejectOwner`/`rotRejectForeign` but leaves `rotRejectLoggedAt` alone, and `clearHookState` doesn't touch it either. A session that rejects id A, logs it, then rotates and neg-caches a different id B within the same minute suppresses the line for B for up to 60s. That's precisely the window where a user hits `!` and files a wrong-status report — and the published debug log then contains no evidence hooks are being dropped, which is the whole-pipeline-trace cost this logging exists to remove. The stamp is per-Session but the state it throttles is per-(owner, foreign) pair.

2. **Split locking.** `rejectLoggedAt` is snapshotted under `s.mu.RLock()` at line 438; the stamp is written under a separate `s.mu.Lock()` at 466. The UI-path caller (`syncHookStatuses` with `resolveRotation=false`, `app.go:5017`, on the Update goroutine) and the worker-path caller can both pass the check before either writes, and log the same drop twice.

3. **Zero-value first hit.** The first `negCached` hit right after a neg-cache logs immediately, duplicating the "ignoring foreign claude session" line emitted moments earlier at 504.

Resetting the stamp inside `negCacheRotation` fixes (1) and (3); folding the check + stamp into one `s.mu.Lock()` fixes (2).

internal/session/session_test.go
```diff
@@ -47,7 +47,16 @@ func TestDetectStatus(t *testing.T) {
...
+		// Every activity line Claude renders opens with a rotating glyph (37/37 real
+		// captures), and the glyph is what separates a live indicator from prose that
+		// happens to close on a parenthesised duration — so it is required here.
+		{"whimsical pattern", "· Clauding… (53s · ↓ 749 tokens)\n", StatusRunning},
+		// The glyph set rotates wider than spinnerChars covers (`·` and `✻` are absent
+		// from it), so these two must be carried by isWhimsicalActivity alone. Neither
+		// has a token counter — requiring one made them read as finished mid-turn.
+		{"whimsical thinking without token counter", "· Improvising… (51s · almost done thinking with xhigh effort)\n", StatusRunning},
+		{"whimsical bare duration", "✻ Marinating… (33s)\n", StatusRunning},
+		{"prose with duration but no glyph not matched", "we cut cold start… (2s) after the rewrite\n❯\n", StatusFinished},
```
hayke102: **This case never reaches the glyph gate — the one guard that replaced the removed token anchor is untested.**

`"we cut cold start… (2s) after the rewrite"` ends in `e`, so `isWhimsicalActivity` returns false at the `HasSuffix(")")` check and `trimmed[0] < 0x80` is never consulted. Confirmed by mutation: deleting `|| trimmed[0] < 0x80` leaves every test in `internal/session` green (deleting the `…` requirement *does* fail the suite).

The case name promises coverage that doesn't exist, so a future reader will trust it, and a later simplification can drop the gate with a fully green run — after which every idle pane whose scrollback ends a line in `)` reads as Running.

A one-word edit fixes it: end the prose in `)`, e.g. `"we cut cold start… (2s)"` — no glyph, still `Finished`, and now the gate is the only thing making it pass.

changelog/unreleased/activity-line-detection.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Busy sessions stop flickering to idle.** Status detection missed activity lines without a token counter (`✽ Improvising… (51s · thinking with xhigh effort)`), so a working session could read idle for a frame and flip back. Mid-`/compact` sessions now read running too.
```
hayke102: **The example line was already detected before this change.**

`✽` is in `spinnerChars` (session.go:1494), and `detectRunning`'s spinner loop matches any of the bottom-10 lines whose trimmed prefix is one of them — so `✽ Improvising… (51s · thinking with xhigh effort)` already returned `StatusRunning` on master. The real gap is glyphs *outside* `spinnerChars`, i.e. `·` and `✻`, which is what the new test cases correctly use.

As written, a user who hits the actual bug won't recognise it from the release notes, and the wrong example gets cited back during triage. Swapping the glyph to `·` or `✻` makes it honest.

Second, smaller thing: the fragment ends with "Mid-`/compact` sessions now read running too." — a distinct behaviour change trailing off the flicker fix, against CLAUDE.md's "one idea per bullet". It's a nice change and deserves its own bold headline rather than the tail of this one.

.claude/skills/debug-status/SKILL.md
```diff
@@ -132,7 +132,7 @@ Read it as:
...
 - **No `waiting` value exists.** A permission prompt still reads `idle`/`busy` here, so this **corroborates Running vs Idle but cannot source Waiting** — use hooks/pane for waiting.
-- **Blind to `/compact`.** Compaction is not counted as `busy`, so the file stays `idle` through a multi-minute compaction — the same blind spot the hook, pane, and transcript all share. If the symptom is "idle during compaction," this file won't rescue it; the fix is detection-side (a `PreCompact` hook → running).
+- **Blind to `/compact`.** Compaction is not counted as `busy`, so the file stays `idle` through a multi-minute compaction — a blind spot it shares with the hook and the transcript. If the symptom is "idle during compaction," this file won't rescue it. The **pane** does see it: `detectRunning` matches the `· Compacting conversation… (2m 27s)` activity line, so a session reading idle mid-compaction points at the pane layer (or at nothing reaching it), not at compaction being undetectable.
```
hayke102: **Worth naming the position window here** — the claim is true but conditionally.

"the pane *does* see it" holds only when the compaction line is within the bottom 20 non-empty lines (`whimsicalN := min(20, len(recentLines))`). During a long compaction, queued messages and chrome rendered beneath it can push it out, and then the pane genuinely *can't* see compaction — at which point this doc sends the debugger hunting a pane-layer bug that isn't there.

One clause ("…matches the activity line **when it's still within the bottom 20 lines**") keeps the guidance actionable in the case where it fails.

internal/session/session.go
hayke102: Fixed in 69040b2 — this was the one that mattered, and I reproduced all of it: the row is at `snapshots/2026-06-02T16-19-34_cache-stale-while-revalidate-c/pane_clean.txt:44`, injecting it 4 lines from the bottom flips all three waiting fixtures, and the downstream chain is exactly as you described (`waitingPaneConfirmed` never set → 5s backstop makes `trustPane` true → `s.Status = StatusRunning`).

Took the structural-glyph exclusion: `⎿ ⏺ │ ├ └`. Enumerating those is safe in a way enumerating the spinner set isn't — gutter and box chrome is stable, the spinner set is what churns. Note capitalisation alone would *not* have caught this one (`Running` is capitalised), so both clauses were needed.

Guarded by `bash tool gutter row not matched` in `session_test.go`, and mutation-verified: deleting the exclusion turns the suite red.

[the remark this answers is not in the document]
hayke102: Fixed in 69040b2. Both reproduce — `trimmed[0] < 0x80` was a byte test, so any accented word passed it.

Adopted the rune decode and the letter/digit rejection as asked. One deviation: I did **not** require the ellipsis to terminate the first activity token. That would reject `· Compacting conversation… (2m 27s)`, where it terminates the second word — and this PR deliberately makes compaction read running. Word count can't separate that from `· quoted example…` either (both two words). Used **capitalisation** instead, which rejects the prose case and keeps the two-word activity line.

Both your examples are now guards (`accented prose not matched`, `lowercase prose behind a glyph not matched`) and each is mutation-verified to fail if its clause is dropped.

internal/session/session.go
hayke102: Fixed in 69040b2 — took the union.

Being straight about the split: I couldn't confirm the failure modes (blank glyph phase, glyph clipped at the pane edge) — all 37 real activity lines I have carry a glyph. The coverage loss was real though: I replaced the glyph-less case instead of keeping it, so that shape was both unmatched and untested.

The union is the better design regardless of whether those frames occur. Shape A (`tokens` + `· ↓`/`· ↑`) is the proven anchor and stays; shape B only has to carry counter-less lines, which is what lets it take a strict gate without risking a false negative — and that strict gate is what closed the three false positives in the sibling threads. So your suggestion narrowed the widening exactly as you predicted.

Glyph-less case restored as `whimsical token counter without glyph`. One note: the `waiting for` guard is on shape B only — putting it on A would narrow long-standing behaviour, which the union exists to preserve.

internal/session/session.go
hayke102: Fixed in 69040b2 — added the guard, mirroring the spinner branch.

Your read on severity was right: the `│` you cited is now excluded as box chrome anyway, so the boxed rendering you constructed is caught by that instead. But the guard still earns its place for an *unboxed* glyph-prefixed waiting line, which the box exclusion wouldn't touch.

That distinction matters for testing — the box case can't isolate the clause. So the guard is `unboxed team-waiting line not matched` (`· Waiting for team lead approval… (2m 13s)`), which nothing else rejects; mutation-verified. The boxed form is covered separately.

Scoped to shape B only, since widening it to the token-counter shape would narrow behaviour that predates this PR.

internal/session/session.go
hayke102: Fixed in 69040b2 — all three confirmed, and (1) is the one that actually bites.

(1) and (3) pull opposite ways on first-hit behaviour, and your remedy resolves it: stamping in `negCacheRotation` means the initial event is always covered by the `ignoring foreign claude session` line firing on the same cycle, so the throttled line becomes a pure still-happening heartbeat rather than a duplicate — and a newly rejected pair can't inherit a spent interval. Also zeroed it in `clearHookState` so a restart re-arms; the stamp should follow the pair it throttles, and that's the third place the pair is written.

(2) folded into one `s.mu.Lock()`.

internal/session/session.go
```diff
@@ -449,7 +458,23 @@ func (s *Session) UpdateHookStatus(hs *HookStatus, resolveRotation bool) bool {
...
+					s.rotRejectLoggedAt = time.Now()
+					s.mu.Unlock()
+					// frozenHookAge is the diagnostic: it is the age of the hook we are
+					// KEEPING, not of the one we just dropped. A large value next to a
+					// fresh dropped status is the signature of this failure.
+					debuglog.Logger.Info("hook dropped: claude session id still rejected",
+						"id", s.ID, "owner", owner, "foreign", hs.SessionID,
+						"dropped", hs.Status, "frozenHook", frozenHook,
+						"frozenHookAge", time.Since(frozenHookAt).Truncate(time.Second))
+				}
```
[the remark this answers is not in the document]
hayke102: Fixed in 69040b2 — check and stamp now happen under one `s.mu.Lock()`.

Confirmed the race is reachable rather than theoretical: `app.go` calls `syncHookStatuses` from both the worker and the UI path, so two goroutines genuinely do reach this branch for the same session.

Same defect as @hayke102's point 2 on the thread at line 465, fixed by the same edit — that thread also covers two things this one didn't: the stamp was never reset when the rejected pair changed, and the first hit duplicated a line logged moments earlier.

internal/session/session_test.go
hayke102: Fixed in 69040b2 — and reproducing your mutation turned up more than the one case.

Confirmed: deleting `|| trimmed[0] < 0x80` left the suite green, while deleting the `…` requirement failed it. Ending the prose in `)` fixes that. But running the mutation across *every* clause showed two more were covering for each other — dropping the letter/digit check stayed green because the whitespace-separator check caught the same input, and vice versa. And your suggested edit had a side effect: making the prose end in `)` removed the only case testing the trailing-`)` check, so that clause went green too.

So the guards are now one-clause-per-case, each verified to fail alone:

| clause | isolating case |
|---|---|
| trailing `)` | `· Marinating… (33s) and then some prose` |
| `…` present | (existing) |
| duration after `…` | (existing) |
| letter/digit first rune | `1 Compiling… (2s)` |
| gutter chrome | `⎿  Running… (9s · timeout 5m)` |
| whitespace separator | `·Marinating… (33s)` |
| capitalised word | `· quoted example… (2s)` |
| `waiting for` | `· Waiting for team lead approval… (2m 13s)` |
| shape-A token branch | `Clauding… (53s · ↓ 749 tokens)` |

The letter/digit one needed a single-character opener — the separator check tests the rune right after the first, so nothing longer reaches that clause.

changelog/unreleased/activity-line-detection.md
hayke102: Fixed in 69040b2 — you're right, and this was my error rather than a wording problem.

Verified both: `✽` (U+273D) and `✢` (U+2722) are in `spinnerChars`, so both lines I cited already returned `StatusRunning` on master via the spinner branch. The genuinely uncovered glyphs are `·` (U+00B7) and `✻` (U+273B) — which my *test cases* used correctly, so the code was right and only the prose was wrong.

It had spread to four places. Fixed the fragment and the `isWhimsicalActivity` doc comment here; the commit message is immutable without a rewrite, so I've posted a correction on the PR description.

Also split compaction into its own bullet per "one idea per bullet", and took the headline to 2–4 words in second person (covers @coderabbitai's format thread).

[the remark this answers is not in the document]
hayke102: Fixed in 69040b2. Checked against CLAUDE.md: headline was six words and the body opened with "Status detection missed…" rather than second person, so both hold. Rewritten as two bullets with 2–4 word headlines in second person.

`highlight` stays absent — that was settled before the PR went up, same call you'd have recommended.

One correction to the fragment you didn't flag, from @hayke102's thread on the same line: the example glyph `✽` is already in `spinnerChars`, so that line was never broken. The example is now `·`, which is the glyph that actually was.

.claude/skills/debug-status/SKILL.md
hayke102: Fixed in 69040b2 — qualified with the window.

You're right that the unconditional phrasing was the harmful part: a doc that sends a debugger hunting a pane-layer bug that isn't there is worse than one that says nothing. The sentence now names `whimsicalN := min(20, len(recentLines))` and tells the reader to check where the line actually sits before concluding the pane layer is broken.

hayke102:
## Correction to the description above

The two examples I used in the original description and commit message were **wrong**, caught by review:

```
✽ Improvising… (51s · almost done thinking with xhigh effort)
✢ Marinating… (33s)
```

`✽` (U+273D) and `✢` (U+2722) are **both already in `spinnerChars`**, so both of those lines already returned `StatusRunning` on master via the spinner branch. They were never the bug.

The glyphs actually absent from `spinnerChars` are **`·` (U+00B7)** and **`✻` (U+273B)**. Read the description's examples with those substituted. The test cases in the diff used the correct glyphs throughout — only the prose was wrong.

## What changed after review (69040b2)

Ten threads, all verified rather than taken at face value; all ten reproduced.

The serious one: removing the `tokens` anchor made `⎿  Running… (9s · timeout 5m)` match, and since `detectRunning` runs **before** `detectWaiting`, a bash task ticking under a live permission menu returned Running and the menu was never read. Downstream, `waitingPaneConfirmed` never got set, the 5s backstop made `trustPane` true, and `Status` was pinned to Running — so an unanswered prompt dropped out of the `Space` rotation and `Y` refused it. Injecting that row 4 lines from the bottom flipped all three waiting fixtures.

The predicate is now a **union** rather than a replacement:

| | shape |
|---|---|
| **A** (restored) | `tokens` + `· ↓`/`· ↑` — the proven anchor, unchanged from master |
| **B** (new) | glyph + capitalised activity word, for counter-less lines |

Keeping A means nothing that matched before can stop matching, which is what lets B take a strict gate: first rune not a letter/digit, not gutter chrome (`⎿ ⏺ │ ├ └`), whitespace then a capitalised word, no `waiting for`.

Capitalisation rather than "the ellipsis must terminate the first token" — the latter also rejects `· Compacting conversation… (2m 27s)`, and word count can't separate that from `· quoted example…`.

Also fixed the throttle from the first commit: single-lock check-and-stamp, stamped in `negCacheRotation` so a newly rejected pair can't inherit a spent interval, zeroed in `clearHookState`.

Test guards are now **one clause per case, mutation-verified** — dropping any single condition turns the suite red. Three clauses were covering for each other before, and my own fix to the prose case had silently removed the only test of the trailing-`)` check.

Still out of scope, as agreed: the parked-job rotation fix (the *primary* cause of the original report). The new throttled log line makes the next occurrence one grep.

internal/session/session.go
```diff
@@ -1678,22 +1730,89 @@ func detectRunning(recentLines []string, _ string, log *slog.Logger) Status {
...
-		(strings.Contains(lower, "· ↓") || strings.Contains(lower, "· ↑")) &&
-		hasWhimsicalDuration(lower)
+	trimmed := strings.TrimSpace(line)
+	lower := strings.ToLower(trimmed)
+	if !strings.HasSuffix(lower, ")") {
+		return false
+	}
+	// The duration must follow the activity word, not merely appear somewhere on
+	// the line — otherwise prose quoting a timing figure could satisfy the check.
+	ellipsis := strings.Index(lower, "…")
```
hayke102, line 1799: This gate sits **above** the shape-A branch, so shape A now requires an `…` too — which narrows it.

On master:
```go
return strings.HasSuffix(lower, ")") &&
	strings.Contains(lower, "tokens") && ...
```
no ellipsis anywhere. So `Clauding (53s · ↓ 749 tokens)` matched before and returns `false` here.

That's the exact frame the doc comment says shape A exists for — "the shape that still matches if a frame captures with its glyph cell blank or clipped" — and it makes "nothing that matched before stops matching" false. Such a frame now falls through `detectRunning` to the prompt check and reads **finished** mid-turn, and `normalizeForHash` stops stripping the line.

The new `whimsical token counter without glyph` test still contains an `…`, so nothing catches it.

Either move the ellipsis gate inside the shape-B branch, or fix the comment to say shape A was deliberately narrowed and why.

internal/session/session.go
```diff
@@ -1678,22 +1730,89 @@ func detectRunning(recentLines []string, _ string, log *slog.Logger) Status {
...
+// do: one non-alphanumeric glyph, whitespace, then a capitalised activity word.
+//
+// The glyph itself is not matched against a list — that set rotates and grows.
+// But "starts with any non-ASCII byte" is too weak a stand-in: it admits accented
+// prose (`Éclair… (2s)`), gutter rows (`⎿  Running… (9s)`), and lowercase prose
+// behind a real glyph (`· quoted example… (2s)`). Requiring the following word to
+// be capitalised is what separates an activity word (Osmosing, Compacting,
+// Marinating) from a sentence — and unlike "the ellipsis must terminate the first
+// token" it still admits the two-word `Compacting conversation…`.
+func hasActivityGlyphPrefix(s string) bool {
```
hayke102, line 1821: The only rejections here are letters, digits, `RuneError`, and the five gutter glyphs — so plain ASCII punctuation qualifies as "the glyph":

```
- Ran the suite… (12s)      → true
* Marinating… (33s)         → true
> Marinating… (33s)         → true
# Marinating… (33s)         → true
```

Claude renders its own responses as markdown, so a bullet or blockquote ending in a parenthesised duration sits inside `detectRunning`'s bottom-20 window. `detectRunning` runs before `detectWaiting`/`detectFinished`, so an idle session pins to Running until that line scrolls past 20 non-empty lines.

The comment argues capitalisation is what separates an activity word from prose — but a markdown bullet has capitalisation too. Requiring the glyph to be non-ASCII (or excluding `-*>#+•`) closes it without re-enumerating the rotating spinner set.

internal/session/session.go
```diff
@@ -1678,22 +1730,89 @@ func detectRunning(recentLines []string, _ string, log *slog.Logger) Status {
...
+	if !strings.HasSuffix(lower, ")") {
+		return false
+	}
+	// The duration must follow the activity word, not merely appear somewhere on
+	// the line — otherwise prose quoting a timing figure could satisfy the check.
+	ellipsis := strings.Index(lower, "…")
+	if ellipsis < 0 || !hasWhimsicalDuration(lower[ellipsis:]) {
+		return false
+	}
+	if strings.Contains(lower, "tokens") &&
```
hayke102: Shape A is evaluated before `hasActivityGlyphPrefix` is ever reached, so the `activityGutterGlyphs` exclusion doesn't cover gutter rows that carry a counter:

```
  ⎿  Running… (12s · ↓ 3.4k tokens)   → true   (standard Task/Agent row)
  ⎿  Running… (9s · timeout 5m)       → false  (bash row — the one the new test covers)
```

Not a regression — master had no gutter check at all. The issue is that the comment at L1738-1744 states the exclusion is load-bearing against "a bash task ticking under a live permission menu" masking Waiting, and the new test asserts that guarantee, but it only holds for counter-less rows.

Moving the gutter check above the shape-A branch would make the code match the claim; otherwise the comment should say the protection is scoped to counter-less rows.

internal/session/golden_test.go
```diff
@@ -39,7 +39,7 @@ var goldenTests = []struct {
 	{"pane_running_background_agents_multi.txt", StatusRunning, "three in-flight background Explore agents in the dock (`◯ Explore … · ↓ tokens`) while the lead is parked → running"},
 	{"pane_finished_quoted_dock_row.txt", StatusFinished, "idle pane whose conversation quotes a dock row verbatim (`◯ Explore … · ↓ tokens`) above the chrome — must NOT read as running (position guard: dock only scanned in bottom 3 lines)"},
 	{"pane_waiting_permission_with_dock.txt", StatusWaiting, "permission menu with live dock rows below it — waiting must win over the dock's running (detectWaiting precedes the dock check)"},
-	{"pane_finished_compacting.txt", StatusFinished, "mid-`/compact` pane — reads finished from the prompt/queued-message `❯`s (pane detection can't see compaction; the PreCompact hook is what keeps this session running)"},
+	{"pane_finished_compacting.txt", StatusRunning, "mid-`/compact` pane — the compaction activity line (`· Compacting conversation… (2m 27s)`) is matched now that the activity check no longer requires a token counter, so the pane reads running on its own instead of leaning entirely on the PreCompact hook"},
```
hayke102: Question rather than a defect: this flip assumes the compaction line disappears once compaction ends.

In the fixture `· Compacting conversation… (2m 27s)` is at row 25 of 37 — ordinary transcript content ~12 non-empty lines up, not a live-only region. If Claude leaves it in scrollback post-`/compact`, `detectRunning` (which precedes `detectFinished`) keeps returning Running until enough output pushes it past the bottom-20 window — an idle-at-prompt session showing a running dot.

Worth a look at a real post-compact pane; if it does persist, a `pane_finished_post_compaction.txt` golden would pin it (the fixture set only covers mid-compaction today).

Minor: the file is still named `pane_finished_compacting.txt` while now asserting `StatusRunning`.

internal/session/session.go
hayke102: Fixed in 52ed923 — you're right on every point, including that my own justification defeated itself.

Confirmed against master: no ellipsis anywhere, and `hasWhimsicalDuration` scanned the whole line. `Clauding (53s · ↓ 749 tokens)` → `true` on master, `false` on the branch. So "nothing that matched before stops matching" was false, and the reason A exists — a frame whose glyph cell captures blank or clipped — was gated behind a *different* single character such a frame may equally lack.

Both gates moved inside shape B. Shape A is now master's predicate verbatim, and the comment states that every tightening lives in B and why the ellipsis in particular must not be hoisted. Pinned by `whimsical token counter without ellipsis`.

Behavioural impact was nil in practice — all 37 activity lines I have carry an `…` — but the claim was published, so it should be true.

internal/session/session.go
hayke102: Fixed in 52ed923 — verified all five prefixes (`-`, `*`, `>`, `#`, `+`) matched.

This one I regressed myself: the original version's `trimmed[0] < 0x80` required non-ASCII, and I lost that property swapping the byte test for a rune decode. Restored as an explicit `glyph < utf8.RuneSelf` rejection.

Your hit on my reasoning lands too — I argued capitalisation separates activity words from prose, and a markdown bullet is capitalised, so it was never sufficient alone.

One thing I did beyond the ask: with non-ASCII required, the letter/digit rejection is only reachable by something like `É Marinating… (2s)`, so rather than write a test for an input that cannot occur I **deleted the clause**. `Éclair… (2s)` stays rejected by the separator check — nothing follows the first rune but more letters. Net one fewer condition than before.

Guards added for all four markdown shapes; mutation-verified that dropping the non-ASCII clause turns the suite red.

internal/session/session.go
hayke102: Took the second option in 52ed923 — scoped the comment, left the check where it is.

Confirmed the premise: `⎿  Running… (12s · ↓ 3.4k tokens)` → `true`, `⎿  Running… (9s · timeout 5m)` → `false`. My comment did claim a general protection and only delivers it for counter-less rows, so that was overclaiming.

Why not move the check: I scanned all 92 captured panes and found **zero** counter-carrying gutter rows — unlike the bash row in the previous round, which you cited at an exact path and line and I could confirm. So the hole is structurally real but unobserved, while hoisting the check would narrow shape A, which is exactly what your sibling thread objects to, and would make a session read finished if a foreground Task's gutter row were ever the only running indicator. No data either way on that, since the shape isn't in the corpus.

Trading a documented, master-equivalent hole for an undocumented behaviour change on a constructed example seemed the wrong direction. The comment now names the limit explicitly and says it was left deliberately, with the reasoning — so if a real capture turns up, it's a one-line change with the rationale already written down.

internal/session/golden_test.go
hayke102: Checked, and the assumption holds — 52ed923.

Corpus scan: 1 pane carries `Compacting conversation…`, 2 carry `Compacted (ctrl+o to see full summary)`, and **none carries both**. Claude replaces the activity line with the `⎿  Compacted` result row rather than leaving it in scrollback — same lifecycle as every other activity line. So the post-compact pane reads finished on its own and the flip is safe.

Caveat: that's 2 post-compact panes, not a large sample. I didn't add `pane_finished_post_compaction.txt` — I'd have had to hand-write a pane for a state I have no evidence occurs, and a fabricated golden asserting a behaviour nobody has observed is worse than no golden. The reasoning is recorded in the golden table entry instead, so if one ever shows up the expectation is easy to revisit.

Fixture renamed to `pane_running_compacting.txt` — good catch, it's asserted StatusRunning since the previous commit.

PR #222, opened 2026-07-28
fix(status): read past oversized lines in Claude transcripts

hayke102:
## What

Every transcript walk used a `bufio.Scanner` capped at 1MB per line. A Scanner reports an over-long line by returning `false` from `Scan()` and setting `ErrTooLong` — **indistinguishable from a clean EOF unless `Err()` is checked, and none of the four callers checked it.** Each therefore answered from whatever it had read *before* the long line, and answered confidently.

An entry carrying a pasted image or a large tool result is a single JSON line, so this is ordinary transcript content, not corruption. **4 of 554 local transcripts hold a line over 1MB**, the worst 2% of the way into the file.

## How it surfaced

A `D` snapshot showed a session as `finished` while every other signal said running: pane showing `✢ Topsy-turvying… (3m 59s · ↓ 12.5k tokens)`, transcript growing every 1–3s, Claude's own status file `busy`.

That transcript is 12,826 lines with its first oversized entry at line 4,341:

```
lastLeadTranscriptTimestamp() → 2026-07-12 14:26:47   ← 16 days stale
true last lead entry          → 2026-07-28 12:31:53
```

Both `conversationActivePastHook` conditions fail on that stale value, so the tiebreaker for the between-bursts frame — where a working agent's activity line is momentarily absent and the persistent `❯` box reads as a finished idle prompt — silently answered `false`. One such frame flipped a mid-turn session to finished:

```
15:32:04.867  ✻ Topsy-turvying… (3m 42s · … · thought for 8s)   ← thinking block ends
15:32:05.417  matched prompt ❯ linesFromBottom=3                 ← one frame, no activity line
15:32:05.510  hook says waiting but pane shows idle prompt, overriding to finished
```

Recovery was slow for a second reason: `fastPassSessions` carries only Running/Waiting/Starting, so on flipping to Finished the session dropped from the 500ms pass to the 2s round-robin — 14s of silence in the log before the capture.

## Blast radius

The same scanners back `sessionRotationVerdict`, where truncation biases every signal the same direction — a timestamp that looks un-advanced, a parent link that looks absent, a uuid that looks missing, all reading as *"not a rotation"*.

## The fix

One `forEachTranscriptLine` helper on `bufio.Reader`, which has no line limit, replacing all four scanner sites. Lines past an explicit 8MB cap are **skipped and the walk continues** — losing one entry is recoverable (its neighbours are usually seconds away), whereas abandoning the walk turns the entire rest of the file into "doesn't exist". Read errors are returned **and logged**: a missing transcript stays quiet (routine — a session may not have written one yet), while permission errors and mid-read I/O failures log at Warn with the path, so a partial walk is visible instead of returning a stale answer that reads as authoritative.

## Tests

Two cases that place the interesting entries **after** an oversized line and assert each reader still finds them — timestamps and `transcriptContainsUUID` (the rotation path) — plus the over-cap skip and an unterminated final line.

Run against a reverted implementation they fail, reproducing the real-world symptom exactly:

```
lastLeadTranscriptTimestamp = 2026-07-12 14:26:47, want 2026-07-28 12:31:53
transcriptContainsUUID missed a uuid located after the oversized line
```

Full suite green with `-race`; `vet`, `gofmt`, build clean.

## Deliberately not included

- **The fast-pass change.** Keeping a session on the 500ms pass briefly after a Running→Finished flip would cut the recovery lag, but that's an amplifier rather than the cause, and it has its own CPU cost — worth deciding on its own.
- **Any claim this closes the parked-job rotation bug.** Those transcripts have no line over 93KB, so that diagnosis stands independently and remains open.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed transcript processing for long/large sessions so oversized transcript entries no longer stop processing early.
  * Improved reliability when transcripts contain very large lines, including proper handling of the final entry even without a trailing newline.
* **Tests**
  * Added coverage ensuring transcript scanning continues after oversized lines.
  * Added coverage for missing transcript and error reporting when the final JSONL line is unterminated.

### DIFF
--- a/changelog/unreleased/transcript-oversized-line.md
+++ b/changelog/unreleased/transcript-oversized-line.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**Long transcripts stop lying.** A session whose conversation contained a very large entry — a pasted image, a big tool result — could show as finished while it was still working. fleet read the transcript only up to that entry; it now reads all of it.
+**Large transcripts read fully.** A conversation entry over 1MB — a pasted image, a big tool result — used to stop fleet mid-file, so a session that was still working showed as finished. You now get the right status whatever the entry size.
--- a/changelog/unreleased/transcript-oversized-line.md
+++ b/changelog/unreleased/transcript-oversized-line.md
@@ -1,0 +1,5 @@
+---
+type: fixed
+---
+
+**Long transcripts stop lying.** A session whose conversation contained a very large entry — a pasted image, a big tool result — could show as finished while it was still working. fleet read the transcript only up to that entry; it now reads all of it.
--- a/internal/session/claude_name.go
+++ b/internal/session/claude_name.go
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/brizzai/fleet/internal/debuglog"
 )
 
 // claudeProjectDirReplacer mirrors Claude Code's per-cwd transcript dir naming.
@@ -75,7 +77,7 @@ func ReadClaudeSessionName(claudeSessionID, projectPath string) string {
 	jsonlPath := filepath.Join(projectDir, claudeSessionID+".jsonl")
 
 	var lastCustom, lastAI string
-	forEachTranscriptLine(jsonlPath, func(line string) bool {
+	_ = forEachTranscriptLine(jsonlPath, func(line string) bool {
 		// Quick check before JSON parsing. Matches both "custom-title" and
 		// "ai-title"; the Type check below filters any incidental hits.
 		if !strings.Contains(line, "-title") {
@@ -278,7 +280,7 @@ func transcriptParentLink(path string) string {
 		return ""
 	}
 	var link string
-	forEachTranscriptLine(path, func(line string) bool {
+	_ = forEachTranscriptLine(path, func(line string) bool {
 		if !strings.Contains(line, "logicalParentUuid") {
 			return true
 		}
@@ -319,9 +321,18 @@ const transcriptLineCap = 8 << 20 // 8MB
 // uuid reported absent — and every consumer treats them as authoritative. That
 // silently disabled the between-bursts tiebreaker in conversationActivePastHook and
 // biased sessionRotationVerdict toward rejecting genuine rotations.
+// Callers that only need a best-effort answer discard the returned error; the
+// helper has already logged anything beyond a routine missing file.
 func forEachTranscriptLine(path string, fn func(line string) bool) error {
 	f, err := os.Open(path)
 	if err != nil {
+		// A missing transcript is routine — the session may not have written one
+		// yet, or belongs to an agent that keeps none. Anything else (permissions,
+		// a bad path) leaves callers returning a zero value that reads exactly like
+		// "nothing happened", so it must not pass in silence.
+		if !errors.Is(err, os.ErrNotExist) {
+			debuglog.Logger.Warn("transcript: open failed", "path", path, "err", err)
+		}
 		return err
 	}
 	defer f.Close()
@@ -353,6 +364,11 @@ func forEachTranscriptLine(path string, fn func(line string) bool) error {
 			if errors.Is(err, io.EOF) {
 				return nil
 			}
+			// Partial walk. Whatever the caller returns is derived from a prefix of
+			// the file, which is precisely the failure this helper exists to end —
+			// so say so rather than letting a stale answer look authoritative.
+			debuglog.Logger.Warn("transcript: read failed, result is partial",
+				"path", path, "err", err)
 			return err
 		}
 	}
@@ -366,7 +382,7 @@ func transcriptContainsUUID(path, uuid string) bool {
 		return false
 	}
 	found := false
-	forEachTranscriptLine(path, func(line string) bool {
+	_ = forEachTranscriptLine(path, func(line string) bool {
 		if !strings.Contains(line, uuid) {
 			return true
 		}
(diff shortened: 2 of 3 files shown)


changelog/unreleased/transcript-oversized-line.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Long transcripts stop lying.** A session whose conversation contained a very large entry — a pasted image, a big tool result — could show as finished while it was still working. fleet read the transcript only up to that entry; it now reads all of it.
```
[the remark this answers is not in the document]
hayke102: Fixed — fair hit on both counts.

The adjective one stung: `a very large entry` where the actual threshold (1MB) was already sitting in the code, which is exactly the substitution CLAUDE.md's guide warns against.

Now reads:

> **Large transcripts read fully.** A conversation entry over 1MB — a pasted image, a big tool result — used to stop fleet mid-file, so a session that was still working showed as finished. You now get the right status whatever the entry size.

4-word headline, concrete `1MB`, closes in second person.

internal/session/claude_name.go
```diff
@@ -286,32 +279,99 @@ func transcriptParentLink(path string) string {
...
+	r := bufio.NewReaderSize(f, 64*1024)
+	var buf []byte
+	overCap := false
+	for {
+		chunk, err := r.ReadSlice('\n')
+		if errors.Is(err, bufio.ErrBufferFull) {
+			// Mid-line: accumulate until the cap, then drop the rest of this line.
+			if !overCap {
+				if len(buf)+len(chunk) > transcriptLineCap {
+					overCap, buf = true, nil
```
hayke102, line 380: The over-cap branch drops the entry and the walk still returns `nil`, so the caller can't tell it read a truncated view. That's a narrow version of the bug this PR exists to kill: a pasted screenshot is one base64 entry that can exceed 8MB, and if it happens to be the last one, `lastLeadTranscriptTimestamp` returns the *previous* entry's timestamp — and every consumer treats it as authoritative. That's exactly the "stale but confident" answer the doc comment above says must not pass in silence.

The read-failure path 30 lines down already gets a warn; this one is silent. A `debuglog.Logger.Warn("transcript: line over cap, entry skipped", "path", path)` here would at least make it greppable when a wrong-status report comes in.

internal/session/claude_name.go
```diff
@@ -286,32 +279,99 @@ func transcriptParentLink(path string) string {
...
+// helper has already logged anything beyond a routine missing file.
+func forEachTranscriptLine(path string, fn func(line string) bool) error {
+	f, err := os.Open(path)
+	if err != nil {
+		// A missing transcript is routine — the session may not have written one
+		// yet, or belongs to an agent that keeps none. Anything else (permissions,
+		// a bad path) leaves callers returning a zero value that reads exactly like
+		// "nothing happened", so it must not pass in silence.
+		if !errors.Is(err, os.ErrNotExist) {
+			debuglog.Logger.Warn("transcript: open failed", "path", path, "err", err)
```
hayke102: This warn sits on a per-cycle path: `ReadClaudeSessionName` re-reads the transcript every fast cycle during `agentNameFreshPollWindow`, then roughly every 30s after that, per session. A transcript that is unreadable for a non-ENOENT reason — EACCES, or EIO from a project dir on a stale network mount — therefore logs once per session per cycle for the life of the process.

`debug.log`'s last 100 lines are what the bug-report flow pastes into a public issue, so a steady drip here can push out the diagnostics this warn was added to preserve. `rotRejectLogInterval` in `session.go` throttles precisely this shape (repeated log for an unchanged state) and would fit here.

internal/session/claude_name.go
hayke102: Fixed in 02049c9 — and you're right that it's the same bug at a smaller scale. I replaced a silent 1MB abort with a silent 8MB skip, two lines under a comment saying such an answer must not pass in silence. The inconsistency with the read-failure branch below was the tell.

Now warns: `transcript: entry over cap, skipped` with the path and the cap, throttled per (reason, path) since the same oversized line recurs on every scan.

Scale, for the record: the largest line across 554 local transcripts is 1.35MB, so nothing reaches 8MB today. Kept the cap where it is rather than raising it on speculation — the log is what would let us move it on evidence.

One thing I did **not** do: return an error for a truncated walk. Every caller discards the error that already exists, so it would buy nothing today, and I'd rather not add a contract nobody consumes. If you want callers to be able to distinguish it, say so — it's a one-liner plus a sentinel.

internal/session/claude_name.go
hayke102: Fixed in 02049c9. Verified at the call site and it's worse than every 30s: `app.go:5278` sets `recheck = 0` while `AgentSessionName == ""` and the session is inside `agentNameFreshPollWindow`, so it's every worker cycle. Combined with sticky conditions — EACCES doesn't clear itself — that's a permanent drip, not a burst.

The self-defeating part is the sharp one: `debug.log`'s tail is what the `!` flow publishes, so an unthrottled warn evicts the diagnostics it was added to preserve.

Worth noting the tempting shortcut is wrong. Demoting to Debug fixes this but breaks your sibling thread — `debuglog` only enables Debug under `FLEET_DEBUG` (`debuglog.go:45-49`), so the line would be missing from exactly the bug reports it's for. Throttled Warn is the level that serves both.

`warnTranscript` now caps at one line per minute per (reason, path), same interval and shape as `rotRejectLogInterval`. Per-pair rather than global, so one noisy transcript can't mask a second — guarded by `TestWarnTranscriptThrottles`, which I mutation-checked: deleting the throttle turns it red.

Issue #226, opened 2026-07-30
yoooooooo why the fork dont work


tonytimo:
### Description
yoooooooo why the fork dont work

### Steps to Reproduce (last 20 actions)
| Time | Action | Detail | Result |
|------|--------|--------|--------|
| 17:00:26 | open bug report |  | ok |
| 17:00:20 | attach session | Review comments and images from discussion (fork) | ok |
| 17:00:05 | attach session | Review comments and images from discussion | ok |
| 16:59:49 | attach session | Review comments and images from discussion (fork) | ok |
| 16:59:39 | delete session | Review comments and images from discussion (fork) | ok |
| 16:59:27 | attach session | Review comments and images from discussion (fork) | ok |
| 16:59:19 | attach session | Review comments and images from discussion | ok |
| 16:59:13 | delete session | Review comments and images from discussion (fork) | ok |
| 16:59:04 | attach session | Review comments and images from discussion (fork) | ok |
| 16:59:02 | delete session | Review comments and images from discussion (fork) | ok |
| 16:58:25 | attach session | Review comments and images from discussion (fork) | ok |
| 16:58:04 | attach session | Review comments and images from discussion | ok |
| 16:56:53 | attach session | Plan implementation strategy | ok |
| 16:56:24 | attach session | Investigate missing capability errors across test data | ok |
| 16:56:22 | attach session | Investigate missing capability errors across test data | ok |
| 16:55:57 | attach session | Plan implementation strategy | ok |
| 16:55:12 | attach session | Plan implementation strategy | ok |
| 16:46:11 | attach session | Plan implementation strategy | ok |
| 16:45:48 | attach session | Plan implementation strategy | ok |
| 16:45:07 | attach session | Review comments and images from discussion | ok |

### Diagnostics
- **Version**: 2.20.0
- **macOS**: 26.4.1 (arm64)
- **tmux**: tmux 3.6a
- **Claude CLI**: 2.1.220 (Claude Code)
- **Codex CLI**: codex-cli 0.142.0
- **gh CLI**: gh version 2.88.1 (2026-03-12)
- **Sessions**: 24

### Terminal Environment
- **TERM**: `xterm-256color`
- **Terminal**: iTerm.app 3.6.11
- **COLORTERM**: truecolor
- **TUI size**: 192x71
- **LANG**: he_IL.UTF-8
- **tmux default-terminal**: `tmux-256color`
- **tmux mouse**: off

[...]

Issue #156, opened 2026-06-25
fork isn't working (regressed)


1jonahg1:
### Description
fork isn't working (regressed)

### Steps to Reproduce (last 20 actions)
| Time | Action | Detail | Result |
|------|--------|--------|--------|
| 12:06:59 | open bug report |  | ok |
| 12:06:44 | attach session | Build Tier Restriction Engine backend (PR 1) (fork) | ok |
| 12:06:21 | delete session | Build Tier Restriction Engine backend (PR 1) | ok |
| 12:06:18 | attach session | Build Tier Restriction Engine backend (PR 1) (fork) | ok |
| 12:06:13 | delete session | Build Tier Restriction Engine backend (PR 1) | ok |
| 12:06:05 | attach session | Build Tier Restriction Engine backend (PR 1) (fork) | ok |
| 12:04:18 | attach session | Build Tier Restriction Engine backend (PR 1) | ok |
| 12:01:43 | attach session | https://github.com/brizzai/brizzai/pull/3824 | ok |
| 12:01:31 | attach session | Build Tier Restriction Engine backend (PR 1) | ok |
| 11:59:56 | attach session | Debug issue and create failing test | ok |
| 11:59:17 | attach session | Build Tier Restriction Engine backend (PR 1) | ok |
| 11:59:08 | attach session | HIGH case security risk analysis with examples | ok |
| 11:57:39 | attach session | Build Tier Restriction Engine backend (PR 1) | ok |
| 11:54:59 | attach session | Debug issue and create failing test | ok |
| 11:54:15 | attach session | Build Tier Restriction Engine backend (PR 1) | ok |
| 11:52:49 | attach session | HIGH case security risk analysis with examples | ok |
| 11:50:30 | attach session | Debug issue and create failing test | ok |
| 11:50:08 | attach session | https://github.com/brizzai/brizzai/pull/3824 | ok |
| 11:48:09 | attach session | https://github.com/brizzai/brizzai/pull/3822 | ok |
| 11:40:38 | attach session | Build Tier Restriction Engine backend (PR 1) | ok |

### Diagnostics
- **Version**: 2.12.0
- **macOS**: 26.5.1 (arm64)
- **tmux**: tmux 3.6a
- **Claude CLI**: 2.1.191 (Claude Code)
- **Codex CLI**: codex-cli 0.130.0
- **gh CLI**: gh version 2.89.0 (2026-03-26)
- **Sessions**: 19

### Terminal Environment
[...]

Issue #142, opened 2026-06-21
forked session is opened with the wrond session id


tonytimo:
### Description
forked session is opened with the wrond session id

### Steps to Reproduce (last 20 actions)
| Time | Action | Detail | Result |
|------|--------|--------|--------|
| 09:46:29 | open bug report |  | ok |
| 09:46:18 | attach session | brizz-2 | ok |
| 09:46:05 | attach session | brizz-2 | ok |
| 09:46:00 | attach session | brizz-2 (fork) | ok |
| 09:45:45 | attach session | brizz-2 | ok |
| 09:45:35 | attach session | brizz-2 (fork) | ok |
| 09:45:26 | attach session | brizz-2 (fork) | ok |
| 09:45:15 | delete session | brizz-2 (fork) | ok |
| 09:44:36 | attach session | brizz-2 (fork) | ok |
| 09:40:37 | attach session | Simplify document management system for v1 (fork) | ok |
| 09:40:32 | delete session | Move to master and pull | ok |

### Diagnostics
- **Version**: 2.9.0
- **macOS**: 26.4.1 (arm64)
- **tmux**: tmux 3.6a
- **Claude CLI**: 2.1.185 (Claude Code)
- **Codex CLI**: codex-cli 0.135.0
- **gh CLI**: gh version 2.88.1 (2026-03-12)
- **Sessions**: 39

### Terminal Environment
- **TERM**: `xterm-256color`
- **Terminal**: iTerm.app 3.6.9
- **COLORTERM**: truecolor
- **TUI size**: 281x71
- **LANG**: he_IL.UTF-8
- **tmux default-terminal**: `tmux-256color`
- **tmux mouse**: off

<details><summary>Debug Log (last 100 lines)</summary>

```
time=2026-06-18T17:39:56.813+03:00 level=INFO msg="overridden waiting hook but pane shows idle prompt, settling to finished" session=b6b7d0cd-1781617265 title="elide files channel traceloop spans"
time=2026-06-18T17:40:00.819+03:00 level=INFO msg="overridden waiting hook but pane shows idle prompt, settling to finished" session=a5d0be20-1781708765 title="bulk session export parallelize"
time=2026-06-18T17:40:10.806+03:00 level=INFO msg="overridden waiting hook but pane shows idle prompt, settling to finished" session=8cbebb27-1781531587 title="Test all queries in the UI manually"
[...]

PR #229, opened 2026-08-02
fix(status): decide conversation ownership by reporting process

hayke102:
A session pins itself to the agent session_id that claimed it, and a hook from a
different id has to prove it continues that conversation before being adopted.
When the proof fails the id is neg-cached — and nothing undoes that. Ownership is
released on one condition only: a `dead` hook *for the owner id*. An in-process
rotation (`/clear`) never sends one, so a rotation we could not prove leaves the
session pinned to a conversation nobody is in, for the rest of its life.

Everything downstream reads that id. Status freezes (every hook from the live
conversation is dropped), `r` resumes the abandoned conversation, and `f`/`F`
fork it — which is how the state gets reported: three times now, as "the fork
doesn't work" (#142, #156, #226). #226 is the first report carrying the
fork-initiation logging from #151, and it is conclusive: the source session fires
UserPromptSubmit under `fb8a487e` at 17:00:09, and eight seconds later fleet
forks `c6e65cc4`.

The proof fails for reasons that have nothing to do with the conversation.
Before #222 a transcript entry over 1MB — a pasted image — ended the scan early
and biased every rotation signal toward "not a rotation"; a pair that stays
undecidable past the retry cap is neg-cached on a timeout rather than on
evidence.

## Ask the process, not the transcript

Transcript recency cannot separate the two cases it is being asked about. A
nested `claude` spawned inside a tool call inherits FLEET_INSTANCE_ID and fires
hooks under its own conversation id while its parent, blocked on that call,
writes nothing — the same trace a rotation leaves. No margin fixes that; it only
sets how long the wrong answer takes.

A hook command runs as a direct child of the agent that fired it, so the handler
records `os.Getppid()` in the status file. Ownership then asks which *process* is
reporting:

  - same pid, new session id -> the owner itself under a new name. `/clear`
    starts a fresh conversation inside the running process. A rotation.
  - different pid, owner gone -> the conversation that held this session ended;
    whatever reports now succeeded it.
  - different pid, owner alive -> two agents at once. Reject.

Note the first row, because it is the one the reported bug lives in and the one
a liveness check gets backwards: an in-process `/clear` leaves the owner's pid
very much alive.

Fails closed. A recycled pid reads as "owner alive" and leaves the session
frozen, which is today's behaviour — so a wrong answer costs nothing already
lost, where the opposite bias would hand a session's status and resume id to a
scratch conversation.

The transcript predicate survives as the fallback for status files written
before the field existed, corrected to measure how long the owner has been
silent rather than how far the new conversation ran past it. The difference is
load-bearing: six minutes of work after a `/clear` then a pause never advances a
"ran N past the owner" measure beyond six minutes, so that session stayed frozen
for good — and a short piece of work after `/clear` is the common shape. The
fallback self-clears on the next hook from the current handler.

## Also

  - The neg-cache is keyed by foreign id. One slot meant an eval harness's
    stream of sub-agent ids evicted the rotated pair, so it paid a full
    two-transcript verdict every ~500ms cycle — the cost the cache exists to
    avoid — and its recovery clock was reset before it could fire.
  - `r` heals, not just fork. Restart and respawn resolve the id before
    `clearHookState()` wipes the state it is derived from.
  - `f` and `F` reach the action log. #226's steps-to-reproduce lists eleven
    attaches and four deletes and not one of the five forks it is about.

## Verified against Claude 2.1.220

  - a hook's `getppid()` is the `claude` process (probed before building on it);
  - `--resume <unknown-id> --fork-session` exits without firing SessionStart, so
    the forks in #226 resumed a real conversation — just a dead one;
  - a fork reports its OWN id at SessionStart, where 2.1.191 reported the
    parent's. The `forkParentID` adoption branch is now dead code against
    current Claude; left in place as it remains correct for older CLIs.

Tests cover both directions and are mutation-checked: deleting the same-pid
branch fails the in-process rotation cases, forcing liveness to report "gone"
fails the three nested-agent guards, and restoring the old
"how far the new session ran" predicate fails the short-post-`/clear` case.

### DIFF
--- a/changelog/unreleased/fork-stale-conversation.md
+++ b/changelog/unreleased/fork-stale-conversation.md
@@ -1,0 +1,4 @@
+---
+type: fixed
+---
+**Forks resume the live conversation.** Forking a session that had rotated to a fresh conversation (a `/clear`, typically) could copy the one it left behind, so the fork opened work you finished an hour ago. It now follows the conversation the session is actually in.
--- a/changelog/unreleased/fork-stale-conversation.md
+++ b/changelog/unreleased/fork-stale-conversation.md
@@ -1,4 +1,4 @@
 ---
 type: fixed
 ---
-**Forks resume the live conversation.** Forking a session that had rotated to a fresh conversation (a `/clear`, typically) could copy the one it left behind, so the fork opened work you finished an hour ago. It now follows the conversation the session is actually in.
+**Forks follow the conversation.** Forking a session that had started a fresh conversation (after `/clear`, typically) could copy the one it left behind, so you got back work you finished an hour ago. You now fork whatever the session is actually in.
--- a/changelog/unreleased/frozen-conversation-id.md
+++ b/changelog/unreleased/frozen-conversation-id.md
@@ -1,4 +1,4 @@
 ---
 type: fixed
 ---
-**Stuck sessions unstick themselves.** A session could pin itself to a conversation that had already ended and then ignore every status update for the rest of its life — a frozen dot, and `r` reopening the wrong conversation. It now notices the old one has gone quiet and moves on.
+**Stuck sessions recover.** A session could pin itself to a conversation that had already ended and then ignore every status update for the rest of its life, so you got a frozen dot and an `r` that reopened the wrong conversation. Both come back the moment the old conversation's process is gone.
--- a/changelog/unreleased/frozen-conversation-id.md
+++ b/changelog/unreleased/frozen-conversation-id.md
@@ -1,0 +1,4 @@
+---
+type: fixed
+---
+**Stuck sessions unstick themselves.** A session could pin itself to a conversation that had already ended and then ignore every status update for the rest of its life — a frozen dot, and `r` reopening the wrong conversation. It now notices the old one has gone quiet and moves on.
--- a/internal/session/claude_name.go
+++ b/internal/session/claude_name.go
@@ -270,36 +270,36 @@ func sessionRotationVerdict(projectPath, ownerSessionID, newSessionID string) ro
 	return rotationNo
 }
 
-// deadOwnerSupersedeMargin is how far a rejected session's transcript must have
-// advanced past the owner's last entry before ownerConversationSuperseded accepts
-// that the owner conversation is dead.
+// deadOwnerSilenceWindow is how long the owner conversation must have been silent
+// before ownerConversationAbandoned accepts, on transcript evidence alone, that it
+// is gone.
 //
-// The margin has to clear the one benign way a foreign id outlives the owner's
-// last write: a nested child claude spawned from a tool call, during which the
-// parent writes nothing until the child returns. Fifteen minutes is far longer
-// than such a call runs, and far shorter than the "forever" a wrongly rejected
-// rotation otherwise lasts.
-const deadOwnerSupersedeMargin = 15 * time.Minute
+// Only the fallback path uses it — when the owner's pid is known, liveness answers
+// the question outright. It exists because the case the neg-cache guards against
+// (a nested agent that inherited FLEET_INSTANCE_ID) leaves the same trace as a
+// rotation: an owner that stopped writing. A parent blocked on a child inside a
+// tool call is silent for as long as the call runs, so the window has to clear
+// that, and fifteen minutes does — while staying far shorter than the "forever" a
+// wrongly rejected rotation otherwise lasts.
+const deadOwnerSilenceWindow = 15 * time.Minute
 
-// ownerConversationSuperseded reports whether newSessionID's transcript has
-// demonstrably taken over from ownerSessionID's: the owner stopped writing at
-// least deadOwnerSupersedeMargin before the new session's latest entry, and that
-// entry is itself recent — we only ever heal toward a conversation that is alive
-// now, never toward one that merely outlived the owner at some point in the past.
+// ownerConversationAbandoned reports whether the owner conversation looks dead from
+// its transcript alone: silent for at least deadOwnerSilenceWindow while the new
+// session has written since.
 //
-// This is the recovery path for a rotation sessionRotationVerdict could not prove
-// — a /clear whose handoff timestamps were unreadable, or a pair that stayed
-// undecidable past the retry cap. Those get neg-cached, and a neg-cached owner is
-// released by nothing: an in-process rotation fires no SessionEnd for the id it
-// leaves behind, so the session's status and its resume/fork id stay pinned to a
-// conversation nobody is in (issue #226).
+// This is the FALLBACK. The question it approximates — "has the conversation that
+// owns this session ended?" — is answered exactly by agentProcessAlive when the
+// owner's pid is known, and this is what remains when it isn't: a status file
+// written by a handler older than StatusFile.AgentPID. That is a one-launch
+// transient, since the next hook from the current handler records a pid.
 //
-// Deliberately a different question from sessionRotationVerdict's. That one asks
-// "is this the same conversation continued?" and answers from lineage. This one
-// asks only "is the owner still being written to?", because by the time it runs
-// the lineage answer has already come back no and been wrong. Liveness is the one
-// signal a dead owner cannot fake.
-func ownerConversationSuperseded(projectPath, ownerSessionID, newSessionID string) bool {
+// It asks how long the owner has been silent, NOT how far the new session has run
+// past it. The difference matters: a /clear followed by six minutes of work and
+// then a pause never advances a "new session ran N past the owner" measure beyond
+// six minutes, so the session would stay frozen for the rest of its life — and a
+// short piece of work after a /clear is the common shape, since /clear is usually
+// how new work starts.
+func ownerConversationAbandoned(projectPath, ownerSessionID, newSessionID string) bool {
 	if projectPath == "" || ownerSessionID == "" || newSessionID == "" {
 		return false
 	}
@@ -309,10 +309,12 @@ func ownerConversationSuperseded(projectPath, ownerSessionID, newSessionID strin
 	if ownerLast.IsZero() || newLast.IsZero() {
 		return false
 	}
-	if time.Since(newLast) > deadOwnerSupersedeMargin {
+	// The new conversation must have outlived the owner, not merely coexisted with
+	// it: a child that finished before the owner's last write is no successor.
+	if !newLast.After(ownerLast) {
 		return false
 	}
-	return newLast.Sub(ownerLast) >= deadOwnerSupersedeMargin
+	return time.Since(ownerLast) >= deadOwnerSilenceWindow
 }
 
 // transcriptParentLink returns the logicalParentUuid of the first
--- a/internal/session/claude_name.go
+++ b/internal/session/claude_name.go
@@ -270,6 +270,51 @@ func sessionRotationVerdict(projectPath, ownerSessionID, newSessionID string) ro
 	return rotationNo
 }
 
+// deadOwnerSupersedeMargin is how far a rejected session's transcript must have
+// advanced past the owner's last entry before ownerConversationSuperseded accepts
+// that the owner conversation is dead.
+//
+// The margin has to clear the one benign way a foreign id outlives the owner's
+// last write: a nested child claude spawned from a tool call, during which the
+// parent writes nothing until the child returns. Fifteen minutes is far longer
+// than such a call runs, and far shorter than the "forever" a wrongly rejected
+// rotation otherwise lasts.
+const deadOwnerSupersedeMargin = 15 * time.Minute
+
+// ownerConversationSuperseded reports whether newSessionID's transcript has
+// demonstrably taken over from ownerSessionID's: the owner stopped writing at
+// least deadOwnerSupersedeMargin before the new session's latest entry, and that
+// entry is itself recent — we only ever heal toward a conversation that is alive
+// now, never toward one that merely outlived the owner at some point in the past.
+//
+// This is the recovery path for a rotation sessionRotationVerdict could not prove
+// — a /clear whose handoff timestamps were unreadable, or a pair that stayed
+// undecidable past the retry cap. Those get neg-cached, and a neg-cached owner is
+// released by nothing: an in-process rotation fires no SessionEnd for the id it
+// leaves behind, so the session's status and its resume/fork id stay pinned to a
+// conversation nobody is in (issue #226).
+//
+// Deliberately a different question from sessionRotationVerdict's. That one asks
+// "is this the same conversation continued?" and answers from lineage. This one
+// asks only "is the owner still being written to?", because by the time it runs
+// the lineage answer has already come back no and been wrong. Liveness is the one
+// signal a dead owner cannot fake.
+func ownerConversationSuperseded(projectPath, ownerSessionID, newSessionID string) bool {
+	if projectPath == "" || ownerSessionID == "" || newSessionID == "" {
+		return false
+	}
+	ownerLast := lastTranscriptTimestamp(ClaudeTranscriptPath(ownerSessionID, projectPath))
+	newLast := lastTranscriptTimestamp(ClaudeTranscriptPath(newSessionID, projectPath))
+	// An unreadable transcript on either side is not evidence of anything.
+	if ownerLast.IsZero() || newLast.IsZero() {
+		return false
+	}
+	if time.Since(newLast) > deadOwnerSupersedeMargin {
+		return false
+	}
+	return newLast.Sub(ownerLast) >= deadOwnerSupersedeMargin
+}
+
 // transcriptParentLink returns the logicalParentUuid of the first
 // compact_boundary/summary entry in the transcript at path — the link a
 // continue/resume/fork rotation records to its parent session's tail uuid — or
--- a/internal/session/session.go
+++ b/internal/session/session.go
@@ -78,26 +78,21 @@ type Session struct {
 	hookUpdatedAt    time.Time
 	hookOverriddenAt time.Time // timestamp of hook that was overridden by pane; prevents re-evaluation of same stale hook
 	ownerSessionID   string    // Claude session_id that owns this fleet session; hooks from other (nested) Claudes are ignored
+	ownerPID         int       // agent process behind ownerSessionID (0 = unknown: pre-AgentPID status file)
 	forkParentID     string    // parent's claude session id while this fork hasn't diverged yet; lets us adopt the fork's own id deterministically (cleared on divergence)
 
-	// Negative cache for the rotation check: the last (owner, foreign) pair that
-	// sessionRotationVerdict rejected. A persistent foreign id (a nested-child eval
-	// harness) would otherwise rescan transcripts every worker cycle; once
-	// rejected we short-circuit until the owner or the foreign id changes.
-	rotRejectOwner   string
-	rotRejectForeign string
-	// When we last logged that this pair is still being rejected. The rejection
-	// itself is logged once, on the cycle it is cached — but a neg-cached pair then
-	// drops every subsequent hook silently, so a session whose hook layer is dead
-	// looks identical in the log to one that is simply quiet. Re-announcing it on a
-	// throttle is what makes "the hook is frozen because we rejected its id"
-	// greppable instead of a whole-pipeline trace.
-	rotRejectLoggedAt time.Time
-	// When we last paid the transcript I/O to ask whether a rejected pair has become
-	// recoverable (ownerConversationSuperseded). The neg-cache exists precisely to
-	// keep that scan off the ~500ms worker cycle, so the recovery check rides a
-	// slower clock of its own rather than reopening the cost the cache saved.
-	rotRejectCheckedAt time.Time
+	// Negative cache for the rotation check, keyed by the FOREIGN session id that
+	// was rejected. A persistent foreign id (a nested agent) would otherwise rescan
+	// transcripts every worker cycle; once rejected we short-circuit until the owner
+	// changes.
+	//
+	// Keyed rather than a single (owner, foreign) slot: an eval harness spawning
+	// sub-agents produces a stream of foreign ids, and one slot meant each new id
+	// evicted the last. Two ids alternating then paid a full sessionRotationVerdict
+	// — two transcript walks — on every ~500ms cycle, the exact cost this cache
+	// exists to avoid, and the recovery clock below was reset by every eviction so
+	// it never accumulated.
+	rotRejects map[string]*rotReject
 
 	// Bounded retry tracking for an undecidable (owner, foreign) pair — one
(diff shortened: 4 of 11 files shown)


internal/session/session.go
```diff
@@ -460,10 +465,22 @@ func (s *Session) UpdateHookStatus(hs *HookStatus, resolveRotation bool) bool {
...
-				// Already rejected this (owner, foreign) pair. Re-announce it on a
-				// throttle: this branch is the steady state of a session whose hook
-				// layer has gone dead, and dropping every hook in silence is what
-				// made that state cost a full pipeline trace to identify.
+				// Normally this pair stays rejected for the session's whole life — right
+				// for a nested child, wrong for a rotation we simply could not prove.
+				// Nothing else releases the owner (an in-process rotation fires no
+				// SessionEnd for the id it abandons), so recheck on a slow clock whether
+				// the owner conversation has gone silent while this one keeps writing.
+				if resolveRotation && s.dueForDeadOwnerRecheck() &&
```
hayke102: **The recovery can adopt a nested child's conversation id.**

`ownerConversationSuperseded` releases the neg-cache on *liveness alone*, and the case the neg-cache exists for can satisfy it. The margin's justification — "far longer than such a call runs" — holds for a synchronous tool call, but not for a backgrounded `claude -p` / eval harness that inherits `FLEET_INSTANCE_ID`: the lead finishes its turn and parks (writing nothing) while the child keeps writing for >15m. That's exactly `ownerLast` 15m behind a fresh `newLast`.

Once it fires, the adoption block sets `s.ownerSessionID`/`s.ClaudeSessionID` to the child, and it doesn't self-correct: the *parent's* next hook now arrives as foreign, `sessionRotationVerdict(child, parent)` returns `rotationNo` (huge timestamp gap, no lineage link), and it gets neg-cached and dropped. The dot follows the child, and `r` (→ `buildAgentCmd` → `claude --resume <child id>`) and `F` both open the throwaway sub-conversation.

The structural problem: transcript recency can't tell "the owner died" from "the owner is blocked on its child" — both write into the same project dir and both leave the owner silent. Whatever releases the neg-cache probably needs a signal the child can't produce, rather than a longer margin.

internal/session/session.go
```diff
@@ -628,6 +666,10 @@ func (s *Session) negCacheRotation(owner, foreign string) {
 	// after another from inheriting the previous pair's spent interval — that silence
 	// would land squarely in the window where a user hits `!` and files a report.
 	s.rotRejectLoggedAt = time.Now()
+	// Same for the recovery clock: a pair rejected this instant cannot yet show a
+	// superseded owner (the margin is minutes wide), so start its interval here
+	// rather than letting the very next cycle spend a transcript scan proving it.
+	s.rotRejectCheckedAt = time.Now()
```
hayke102: **Single-slot neg-cache: one third-party hook evicts the rotated pair and starves the fix.**

`rotRejectOwner`/`rotRejectForeign` hold exactly one pair, `negCacheRotation` overwrites unconditionally, and the short-circuit demands an exact pair match.

So: session frozen on `owner-aaa` after a `/clear` to `live-bbb`, pair cached. One hook from a nested child `ccc` isn't cached → takes the `default` branch → `rotationNo` → overwrites `rotRejectForeign` with `ccc`. From there:

- `live-bbb`'s hooks are no longer short-circuited, so they rescan both transcripts on every ~500ms cycle — the exact cost this cache exists to avoid;
- the two pairs evict each other on alternating hooks, so neither ever accumulates the recheck interval reliably;
- `ResolveForkParent` reads whichever pair happens to be resident, so `F` resolves the parent to `ccc` (a sub-agent scratch conversation) or to the stale `owner-aaa` depending on timing.

The nested-child case is the one the neg-cache was designed around, so it isn't hypothetical — it's the same session shape that motivates the 15m margin two hunks up. A small per-owner set, or keying the rejection by foreign id, would make the recovery deterministic.

internal/session/session.go
```diff
@@ -736,6 +778,36 @@ func (s *Session) GetClaudeSessionID() string {
...
+	stored, owner, rejected := s.ClaudeSessionID, s.ownerSessionID, s.rotRejectForeign
+	rejectedAgainst, projectPath := s.rotRejectOwner, s.ProjectPath
+	s.mu.RUnlock()
+
+	// Only a rejection standing against the *current* owner says anything about the
+	// id we are about to fork.
+	if rejected == "" || rejected == stored || owner == "" || rejectedAgainst != owner || stored != owner {
+		return stored, ""
+	}
+	if !ownerConversationSuperseded(projectPath, owner, rejected) {
```
hayke102: **The fork healing declines exactly the case the doc comment above cites.**

The comment says the worker recovery can't be relied on because "a session that rotated and then went quiet fires none" — but `ownerConversationSuperseded` carries a recency gate written for the hook path, where the new transcript is live by construction:

```go
// claude_name.go:312
if time.Since(newLast) > deadOwnerSupersedeMargin {
    return false
}
```

On the fork path that clause rejects the quiet session specifically. Repro: `/clear`, work five minutes, walk away an hour, press `F`. `time.Since(newLast)` ≈ 1h > 15m → `false` → this returns the frozen owner id and the fork resumes the pre-`/clear` conversation — the #226 symptom the changelog says is fixed.

Confirmed with a throwaway probe test: `negCachedSession(ownerLast=-3h, liveLast=-1h)` → `ResolveForkParent() == ("owner-aaa", "")`.

The two callers want different questions. The worker wants "is the new conversation alive *now*"; the fork wants "did the new conversation outlive the owner", full stop. Sharing one predicate is what makes the fork path answer the wrong one.

internal/session/claude_name.go
```diff
@@ -270,6 +270,51 @@ func sessionRotationVerdict(projectPath, ownerSessionID, newSessionID string) ro
...
+	ownerLast := lastTranscriptTimestamp(ClaudeTranscriptPath(ownerSessionID, projectPath))
+	newLast := lastTranscriptTimestamp(ClaudeTranscriptPath(newSessionID, projectPath))
+	// An unreadable transcript on either side is not evidence of anything.
+	if ownerLast.IsZero() || newLast.IsZero() {
+		return false
+	}
+	if time.Since(newLast) > deadOwnerSupersedeMargin {
+		return false
+	}
+	return newLast.Sub(ownerLast) >= deadOwnerSupersedeMargin
```
hayke102: **A rotation followed by under 15 minutes of work stays frozen forever.**

This asks the *new* transcript to have advanced ≥15m past the owner's last write — not that the owner has been silent ≥15m, which is what the doc comment above describes ("the owner stopped writing at least `deadOwnerSupersedeMargin` before the new session's latest entry" is the same clause, but in practice it's a floor on how much work the new conversation did, not on how long the owner has been dead).

`/clear` at 10:00 (verdict undecidable → neg-cached), two questions over six minutes, then stop. `newLast.Sub(ownerLast)` tops out at ~6m and never grows again, so this returns `false` for the rest of the session's life. Every hook from the live conversation keeps getting dropped: the dot stays pinned to the pre-`/clear` state (a stuck `waiting` keeps catching `Space`), and since `ResolveForkParent` gates on the same predicate, `F` still forks the abandoned conversation.

That's #226 unfixed for any short post-rotation session — and short sessions after a `/clear` are the common shape, since `/clear` is usually how a new piece of work starts.

`time.Since(ownerLast) >= deadOwnerSupersedeMargin` is the condition that matches the intent and doesn't depend on how much the user happened to type afterwards.

changelog/unreleased/frozen-conversation-id.md
```diff
@@ -0,0 +1,4 @@
+---
+type: fixed
+---
+**Stuck sessions unstick themselves.** A session could pin itself to a conversation that had already ended and then ignore every status update for the rest of its life — a frozen dot, and `r` reopening the wrong conversation. It now notices the old one has gone quiet and moves on.
```
hayke102: **The `r` half of this claim isn't delivered.**

Only the two fork entry points call `ResolveForkParent` (`app.go:3850`, `app.go:3901`). The restart path doesn't: `restartSession` (`app.go:3548`) → `Restart()` → `buildAgentCmd()`, which reads the raw `ClaudeSessionID`. Worse, `Restart()` calls `clearHookState()` first, which zeroes `rotRejectOwner`/`rotRejectForeign`/`rotRejectCheckedAt` — so by the time the command is built, the evidence a heal would need is already gone. `resumeSelected` goes through the same `Restart()`.

So `r` is healed only as a side effect of the worker recovery — which is precisely what's unavailable in the quiet case that the fork path got an explicit call for.

Either route the restart path through `ResolveForkParent` too, or drop `r` from the fragment and let it promise the frozen dot only.

internal/session/claude_name.go
```diff
@@ -270,6 +270,51 @@ func sessionRotationVerdict(projectPath, ownerSessionID, newSessionID string) ro
...
+// released by nothing: an in-process rotation fires no SessionEnd for the id it
+// leaves behind, so the session's status and its resume/fork id stay pinned to a
+// conversation nobody is in (issue #226).
+//
+// Deliberately a different question from sessionRotationVerdict's. That one asks
+// "is this the same conversation continued?" and answers from lineage. This one
+// asks only "is the owner still being written to?", because by the time it runs
+// the lineage answer has already come back no and been wrong. Liveness is the one
+// signal a dead owner cannot fake.
+func ownerConversationSuperseded(projectPath, ownerSessionID, newSessionID string) bool {
```
hayke102: **This reopens the scan the neg-cache existed to close** (lower confidence than the others — throttled, so it may never bite).

Two `lastTranscriptTimestamp` calls per invocation, each a full `scanTranscriptTimestamp` walk that `json.Unmarshal`s every line containing `"timestamp"` and keeps the last. Its own doc comment still says the scan "runs only on the rare session-id-change path" — `dueForDeadOwnerRecheck` now puts it on a once-a-minute clock for every neg-cached session, on the shared worker goroutine.

At 1/min that's cheap in the normal case. Where it stops being cheap: several sessions holding a permanent (owner, child) rejection — the nested-`claude` shape — against transcripts that are tens of MB on a long-lived session. That's the documented status-worker-stall signature (many sessions flushing together with varied hook ages), which is a genuinely annoying failure to diagnose.

An `os.Stat` mtime check before the walk would close it: the owner's transcript is by definition not being written, so the expensive side of the comparison is the one that's already cacheable.

internal/session/session.go
```diff
@@ -736,6 +778,36 @@ func (s *Session) GetClaudeSessionID() string {
...
+//
+// Normally the answer is just ClaudeSessionID. It is not when a foreign id is
+// neg-cached against the owner: the pane may be in that other conversation while
+// ClaudeSessionID still names the one we froze on, and a fork of a dead id silently
+// copies a conversation the user stopped working in an hour ago (issue #226). The
+// worker's own recovery can't be relied on here — it only reruns when a hook
+// arrives, and a session that rotated and then went quiet fires none.
+//
+// Does transcript I/O. Call it off the Bubble Tea Update loop.
+func (s *Session) ResolveForkParent() (parentID, healedFrom string) {
```
hayke102: **This justification is wrong — the worker does rerun without a new hook.**

> The worker's own recovery can't be relied on here — it only reruns when a hook arrives, and a session that rotated and then went quiet fires none.

`syncHookStatuses` (`app.go:5270`, `resolveRotation=true`) calls `hookWatcher.GetStatus`, which returns the *last cached* `HookStatus` (`hook_watcher.go:164`) and re-offers it on every worker cycle whether or not a new hook fired. That's the whole reason `rotRejectLogInterval` exists — its own comment says "The worker re-offers the same hook every fast cycle (~500ms)".

So the `negCached` recovery *does* re-evaluate once a minute for the life of a quiet session. By the time someone presses `f`/`F`, either the worker already healed `ClaudeSessionID`, or `ownerConversationSuperseded` is still false and this returns the same stale id anyway. On the stated reasoning the duplicate rule changes the outcome only inside the ≤60s throttle gap — while adding transcript I/O to the fork path and a second copy of the healing rule to keep in sync forever.

The path may still earn its keep for the *recency-gate* reason (see the thread below): the two callers genuinely want different predicates. But then the comment should say that, because as written it argues for something that isn't true and a future reader will delete the function on the strength of it.

internal/session/session.go
```diff
@@ -736,6 +778,36 @@ func (s *Session) GetClaudeSessionID() string {
...
+	s.mu.RLock()
+	stored, owner, rejected := s.ClaudeSessionID, s.ownerSessionID, s.rotRejectForeign
+	rejectedAgainst, projectPath := s.rotRejectOwner, s.ProjectPath
+	s.mu.RUnlock()
+
+	// Only a rejection standing against the *current* owner says anything about the
+	// id we are about to fork.
+	if rejected == "" || rejected == stored || owner == "" || rejectedAgainst != owner || stored != owner {
```
hayke102: **Two of these five clauses are unreachable.**

`negCacheRotation` is only ever called with `foreign != owner`, so `rotRejectForeign != rotRejectOwner` always holds. Given the surviving clauses require `rejectedAgainst == owner` and `stored == owner`:

- `rejected == stored` would imply `rejected == owner == rejectedAgainst` — impossible under that invariant.
- `owner == ""` is already covered by `stored != owner` (a non-empty `rotRejectForeign` implies a non-empty `rotRejectOwner`).

The condition that actually carries the meaning — "a rejection stands against the id we're about to fork" — is:

```go
if rejected == "" || stored != owner || rejectedAgainst != owner {
    return stored, ""
}
```

Not a bug, but the two dead clauses are load-bearing-looking: anyone changing the third has to first prove the invariants behind them.

changelog/unreleased/frozen-conversation-id.md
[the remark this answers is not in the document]
hayke102: Sentence count matches the shipped v2.23.2 entries (bold headline + what broke + what you get), so keeping that shape. The second-person point is valid and is fixed — the fragment is rewritten.

changelog/unreleased/fork-stale-conversation.md
```diff
@@ -0,0 +1,4 @@
+---
+type: fixed
+---
+**Forks resume the live conversation.** Forking a session that had rotated to a fresh conversation (a `/clear`, typically) could copy the one it left behind, so the fork opened work you finished an hour ago. It now follows the conversation the session is actually in.
```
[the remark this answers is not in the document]
hayke102: This structure matches the shipped v2.23.2 entries, so keeping it. The headline length flagged in the sibling thread was a fair catch and is fixed.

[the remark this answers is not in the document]
hayke102: Fixed — headline is now `**Forks follow the conversation.**` (4 words). You were right and I'd waved this off after counting sentences and never counting the headline.

changelog/unreleased/frozen-conversation-id.md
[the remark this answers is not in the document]
hayke102: Fixed — rewritten in second person, and the claim itself changed too (`r` now genuinely heals, see the thread on the restart path).

internal/ui/app.go
```diff
@@ -3856,10 +3866,10 @@ func (h *Home) forkSelected() tea.Cmd {
...
 type forkContext struct {
-	parentClaudeSessionID string
-	parentSessionID       string // fleet id of the session being forked (for diagnostics)
-	parentProjectPath     string
-	parentTitle           string
+	parentSession     *session.Session
+	parentSessionID   string // fleet id of the session being forked (for diagnostics)
+	parentProjectPath string
+	parentTitle       string
 }
```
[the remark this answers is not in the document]
hayke102: Not changing this. `ResolveLaunchID` (renamed) only reads fields under the session mutex, so a detached session is a live Go object — no nil deref, no crash. Reaching it also needs the fork source deleted while the worktree picker is open, and the picker is in `modalOpen()` (app.go:1787), so `d` never reaches the sidebar. Holding the pointer instead of a string snapshot is the deliberate part: the id has to be read at dispatch, not at picker-open.

internal/session/session.go
hayke102: Fixed, by taking the last line of this seriously — no margin makes liveness-by-transcript sound, so the neg-cache is now released by a signal the child can't produce.

A hook command runs as a direct child of the agent that fired it, so `os.Getppid()` in `fleet hook-handler` names the agent process. Verified before building on it:

```
hook pid=46219 ppid=45781
45781 45775 claude
45775  5809 /bin/zsh
```

That pid goes in the status file. Ownership then asks whether the owner's process is still running: gone → the conversation ended, a new id reporting is a rotation; alive → the reporter is a child. A rotation happens *inside* the owner process, a nested agent runs *alongside* it — which is the distinction every transcript signal was failing to reconstruct.

Fails closed: a recycled pid reads as "owner alive" and leaves the session frozen, i.e. today's behaviour, so a wrong answer costs nothing already lost — where the opposite bias hands status and the resume id to a scratch conversation.

One correction to the report: it does partially self-correct today. When the child exits it fires `dead` for the adopted id, which releases ownership at session.go:570, and the parent's next hook re-claims. Doesn't rescue the design — a backgrounded child never exits, and the exiting case still flashes `dead` (crash dump, error dot) with `r`/`F` pointing at the scratch conversation until the parent speaks — but 'doesn't self-correct' isn't quite right.

`TestUpdateHookStatusKeepsOwnerWhileProcessAlive` pins it with the owner silent for two hours; it fails if liveness is forced to report 'gone'.

internal/session/session.go
hayke102: Fixed — the cache is keyed by foreign id now (`rotRejects map[string]*rotReject`), with the log throttle and the recovery clock moved per entry so one noisy id can neither evict another's rejection nor reset its clocks. `TestNegCacheKeyedByForeignID` covers the eviction.

Agreed on the framing: the first two bullets were pre-existing and what this PR added was the third — a recovery that silently depended on which pair happened to be resident. `ResolveLaunchID` now looks up the rejection standing against the current owner rather than reading whatever reported last.

internal/session/session.go
hayke102: Fixed — the recency gate is gone entirely. With liveness deciding, there's no clause left that rejects a session because you walked away from it; the transcript fallback (no-pid case) has no recency gate either. Your repro — `/clear`, work five minutes, walk away an hour, press `F` — now heals.

You were right that the two callers wanted different questions, though the resolution went the other way: they want the *same* question, which is 'is the owner's process gone', and it was only the transcript proxy that forced them apart.

internal/session/claude_name.go
hayke102: Fixed — this was the one that made the PR not deliver what it claimed, and you had the right condition. It's `time.Since(ownerLast) >= deadOwnerSilenceWindow` now (renamed from …Margin, since it measures silence rather than a gap between two conversations).

Your `/clear` at 10:00 + six minutes + stop is now a test: `TestUpdateHookStatusRotationShortlyAfterClearRecovers`. Mutation-checked — restoring `newLast.Sub(ownerLast)` fails it.

The predicate is also demoted to the fallback for status files written before the pid field, so on the primary path the question doesn't depend on how much you happened to type after the rotation at all.

changelog/unreleased/frozen-conversation-id.md
hayke102: Fixed by routing the restart path through the resolver rather than by weakening the copy. `Restart()` and `RespawnClaude()` call `adoptResolvedLaunchID()` **before** `clearHookState()`, so the evidence is read while it still exists; `TestAdoptResolvedLaunchIDBeforeClear` pins that ordering. The fragment is rewritten anyway (second person, per the CodeRabbit thread), and its `r` claim is now true.

Worth noting your point about the worker in the sibling thread cuts both ways here: the side effect was real, since `syncHookStatuses` re-offers the cached hook every cycle. It just wasn't guaranteed at the moment the key is pressed, which is what an explicit call fixes.

internal/session/claude_name.go
hayke102: Good catch on the cost, and the pid signal takes it off the per-cycle path rather than needing the mtime guard: liveness is a `Signal(0)`, and the transcript walk survives only on the no-pid fallback — a status file written by a handler older than the field, which self-clears on that session's next hook. Adding a cache to a path that rare felt speculative.

If that fallback turns out stickier than expected the `os.Stat` guard is a few lines and `convTranscriptPath/Size/Mtime` is right there to model it on — happy to add it now if you'd rather not rely on that reasoning.

internal/session/session.go
hayke102: You're right, and I verified it: `GetStatus` returns `w.statuses[sessionID]` (hook_watcher.go:164) and the worker calls `syncHookStatuses(sessions, true)` every cycle (app.go:5270), so the recovery does re-evaluate for a quiet session. The comment argued from a false premise.

Rewritten, and it says neither of the two things I'd claimed. Both callers now share one predicate, so it isn't the predicate difference either. What the call buys is the gap — the recovery is on a one-minute clock and a key pressed inside it would launch the stale id — plus, for restart, ordering: `clearHookState()` wipes the state the heal derives from, so it has to run first. Renamed `ResolveLaunchID` now that fork, restart and resume all use it.

internal/session/session.go
hayke102: Fixed — down to the three that carry meaning:

```go
if rejected == "" || stored != owner {
    return stored, ""
}
```

(The third clause folded away too: the lookup now selects the rejection standing against the current owner, so `rejectedAgainst != owner` can't be reached.) The invariant is stated in the comment rather than encoded as dead clauses, which was the actual complaint.

hayke102:
Follow-up on the thread about what releases the neg-cache — I found a flaw in my own first attempt at it and pushed a correction (d434b9a), so the pid rule isn't quite what the resolved thread describes.

**Liveness alone gets the reported case exactly backwards.** `/clear` rotates the session id *inside* the running agent, so the process that owned the abandoned conversation is the same one now running the new conversation. It's alive. A rule asking "is the owner's process still running?" reads that as "owner still here, so the reporter is a child" and keeps the session frozen — the one shape #226 was actually reported in was the one shape the fix would have missed.

The question is identity, not vital signs, and the hook carries what answers it:

| owner pid vs reporter pid | owner process | verdict |
|---|---|---|
| same | (alive by definition) | the owner under a new name — **rotation** |
| different | gone | the reporter succeeded it — **rotation** |
| different | alive | two agents at once — **nested, reject** |

Liveness alone gets only the third row right. `conversationSucceeds` now takes both pids, and the rejected id's pid is kept on the neg-cache entry so `ResolveLaunchID` asks the same question rather than a weaker one.

Also fixed while there: with several ids rejected against one owner (an eval harness spawning sub-agents), `ResolveLaunchID` now prefers the one reported by the owner's own process — map iteration order is random, so the previous pick could vary between runs of the same state. That's the tail of the single-slot thread I'd called done.

Mutation-checked: deleting the same-pid branch fails `TestUpdateHookStatusRecoversInProcessRotation` and `TestResolveLaunchIDHealsInProcessRotation`.

PR #230, opened 2026-08-02
feat(cli): add `fleet worktree` and adopt externally-created sessions

### LINKED ISSUE #227: add an option to create a worktree from the CLI
### Problem
add an option to create a worktree from the CLI

### Environment
- **Version**: 2.23.1
- **OS**: macOS 26.2 (arm64)

hayke102:
Closes #227.

Creating a worktree meant opening the TUI and pressing `w`. Nothing in the CLI surface (`add` / `list` / `remove` / `hooks` / `update`) touched worktrees, so a terminal-first user couldn't script it or fire it from wherever they already were.

## `fleet worktree <branch>`

Alias `wt`. Mirrors the `w` key: create the worktree, copy `.claude/settings.local.json` + `copy_files` entries, start a session.

```bash
fleet worktree fix-login                      # worktree + session, default agent
fleet worktree fix-login --base develop       # base branch (default: repo's default branch)
fleet worktree fix-login --path ~/code/fleet  # target repo (default: cwd)
fleet worktree fix-login --agent codex        # agent (default: default_agent config)
cd "$(fleet worktree fix-login --no-session)" # worktree only, path on stdout
```

Three decisions worth review:

- **Flags parse on either side of the branch.** A single `flag.Parse` stops at the first positional, so `fleet worktree foo --no-session` — the order most people type — would fail. Caught during manual testing, not planning.
- **`--agent` is validated explicitly.** `agent.Parse` falls back to Claude for anything unrecognized, so `--agent codek` would otherwise silently launch the wrong agent.
- **Worktrees always derive from the main clone** (`git.GetMainWorktreePath`). Running it inside `myrepo-scratch-a` gives `myrepo-nested`, not `myrepo-scratch-a-nested` — matching what `w` does on an origin header.

It also calls `debuglog.Init()` first. The fallback `debuglog.Logger` writes to **stderr** and both the provider and `Session.Start` log at Info, so without it the command sprays slog lines over the terminal. `fleet add` / `fleet list` still do this — left alone as adjacent code, but it's a one-line fix if wanted.

## Adoption of externally-created sessions

The TUI reads sessions from SQLite exactly once, in `Init`. Without this, a session created by the CLI while fleet is open stays invisible until restart — the new command feeling broken for exactly the user most likely to run it.

`maybeAdoptExternalSessions` runs in the worker's heavy pass, self-throttled to 5s, diffs the table against the cycle's snapshot, and sends `adoptSessionsMsg`. The handler mirrors `handleSessionCreateResult` minus `SaveSession` and analytics.

- **Adoption only** — rows deleted by another process are never dropped. `fleet remove` from a second terminal still leaves a ghost until restart, unchanged from today.
- **The cursor must not move.** An adopted row arrives on a timer, not because the user asked for it, so it's captured via `targetForCursor()` and restored by identity after the rebuild — the hazard `CLAUDE.md` documents for the context menu.
- **No resurrection risk** — `deferDelete` removes the SQLite row immediately, ahead of the 5s undo window, so the sweep can't re-adopt a just-deleted session.
- Dedupe is authoritative only on the Update goroutine: the worker's snapshot is taken at cycle start, so it can offer a session the TUI created since.

## Verification

`make build`, `make lint` (0 issues), `make test` (`-race`) all pass. 15 new cases across `cmd/fleet/worktree_test.go` and `internal/ui/adopt_test.go`.

Manual, against a scratch repo under a throwaway `$HOME`: `--no-session` emits only the path, `--base`/`--path` work, all six error paths give clean messages, the full path starts tmux + writes the DB, and a worktree created from inside a worktree lands next to the main clone.

For adoption I ran the TUI in a detached tmux pane and created a session from a second shell — it appeared without a restart, and the breadcrumb confirmed the cursor stayed on its row even though the new one was inserted *above* it.

The cursor test carries a vacuity guard: it fails if the adopted row doesn't actually shift the parked row's index, so it can't silently stop testing anything.

Two changelog fragments (`added` + `improved`), split per the one-file-one-bullet rule. Neither is a highlight.

https://claude.ai/code/session_01Q1yXo3oP17DQwVbnagnWDY

## Summary by CodeRabbit

* **New Features**
  * Added the `fleet worktree` command for creating worktrees from the CLI.
  * Added options for base branch, path, agent selection, and creating a worktree without a session.
  * Fleet now detects externally created sessions while running and displays them automatically within approximately five seconds.

* **Documentation**
  * Added usage guidance and unreleased changelog entries for worktree creation and external-session adoption.

### DIFF
--- a/changelog/unreleased/adopt-external-sessions.md
+++ b/changelog/unreleased/adopt-external-sessions.md
@@ -1,0 +1,5 @@
+---
+type: improved
+---
+
+**No restart to see new sessions.** A session created from another terminal — `fleet worktree`, `fleet add` — now shows up in a running fleet within ~5 seconds, instead of waiting until the next launch.
--- a/changelog/unreleased/adopt-external-sessions.md
+++ b/changelog/unreleased/adopt-external-sessions.md
@@ -2,4 +2,4 @@
 type: improved
 ---
 
-**No restart to see new sessions.** A session created from another terminal — `fleet worktree`, `fleet add` — now shows up in a running fleet within ~5 seconds, instead of waiting until the next launch.
+**No restart needed.** A session created from another terminal — `fleet worktree`, `fleet add` — now shows up in a running fleet within ~5 seconds, instead of waiting until the next launch.
--- a/cmd/fleet/worktree.go
+++ b/cmd/fleet/worktree.go
@@ -1,0 +1,284 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/brizzai/fleet/internal/agent"
+	"github.com/brizzai/fleet/internal/config"
+	"github.com/brizzai/fleet/internal/debuglog"
+	"github.com/brizzai/fleet/internal/git"
+	"github.com/brizzai/fleet/internal/hooks"
+	"github.com/brizzai/fleet/internal/session"
+	"github.com/brizzai/fleet/internal/tmux"
+	"github.com/brizzai/fleet/internal/workspace"
+)
+
+const worktreeUsage = "Usage: fleet worktree <branch> [flags]"
+
+// errMissingBranch is returned when no branch was given. runWorktree prints the
+// usage line alongside it, so the message itself stays a plain error string.
+var errMissingBranch = errors.New("missing branch name")
+
+// worktreeOpts holds the parsed `fleet worktree` invocation. Base and agent are
+// left empty when unset; their defaults depend on the repo (default branch) and
+// the user's config (default agent), which parsing can't see.
+type worktreeOpts struct {
+	branch    string
+	base      string
+	repoPath  string
+	agentName string
+	noSession bool
+}
+
+// parseWorktreeArgs parses and validates the `fleet worktree` flags. Kept pure
+// (no git, no config, no filesystem) so the argument rules are testable.
+func parseWorktreeArgs(args []string) (worktreeOpts, error) {
+	fs := flag.NewFlagSet("fleet worktree", flag.ContinueOnError)
+	fs.Usage = func() {
+		fmt.Fprintln(fs.Output(), worktreeUsage)
+		fs.PrintDefaults()
+	}
+	var o worktreeOpts
+	fs.StringVar(&o.base, "base", "", "base branch to branch from (default: the repo's default branch)")
+	fs.StringVar(&o.repoPath, "path", "", "repo to create the worktree in (default: current directory)")
+	fs.StringVar(&o.agentName, "agent", "", "agent to run: claude, codex, or opencode (default: default_agent config)")
+	fs.BoolVar(&o.noSession, "no-session", false, "create the worktree only, print its path, and start no session")
+	// Parse in a loop, peeling off one positional at a time: Go's flag package
+	// stops at the first non-flag argument, so a single Parse would reject
+	// `fleet worktree my-branch --no-session` — the order most people type.
+	var positional []string
+	for rest := args; ; {
+		if err := fs.Parse(rest); err != nil {
+			return o, err
+		}
+		remaining := fs.Args()
+		if len(remaining) == 0 {
+			break
+		}
+		positional = append(positional, remaining[0])
+		rest = remaining[1:]
+	}
+
+	if len(positional) == 0 {
+		return o, errMissingBranch
+	}
+	if len(positional) > 1 {
+		return o, fmt.Errorf("unexpected argument %q — expected a single branch name", positional[1])
+	}
+	o.branch = strings.TrimSpace(positional[0])
+
+	if msg := workspace.ValidateBranchName(o.branch); msg != "" {
+		return o, fmt.Errorf("%s", msg)
+	}
+	// agent.Parse falls back to Claude for anything it doesn't recognize, so a
+	// typo would silently launch the wrong agent. Reject it here instead.
+	if o.agentName != "" {
+		switch agent.Type(o.agentName) {
+		case agent.Claude, agent.Codex, agent.OpenCode:
+		default:
+			return o, fmt.Errorf("unknown agent %q — expected claude, codex, or opencode", o.agentName)
+		}
+		if o.noSession {
+			return o, fmt.Errorf("--agent has no effect with --no-session (no session is started)")
+		}
+	}
+	return o, nil
+}
+
+func runWorktree(args []string) {
+	opts, err := parseWorktreeArgs(args)
+	if err != nil {
+		if errors.Is(err, flag.ErrHelp) {
+			return // flag package already printed usage
+		}
+		fmt.Fprintln(os.Stderr, err)
+		if errors.Is(err, errMissingBranch) {
+			fmt.Fprintln(os.Stderr, worktreeUsage)
+		}
+		os.Exit(1)
+	}
+
+	// Both the workspace provider and Session.Start log at Info level, and
+	// debuglog's fallback logger writes to stderr — without Init those lines
+	// land on the user's terminal instead of ~/.config/fleet/debug.log.
+	debuglog.Init()
+	defer debuglog.Close()
+
+	if err := tmux.IsTmuxAvailable(); err != nil && !opts.noSession {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+
+	repoPath, err := resolveWorktreeRepo(opts.repoPath)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+
+	base := opts.base
+	if base == "" {
+		base = git.GetDefaultBranch(repoPath)
+	}
+
+	cfg := config.Load()
+	ag := agent.Parse(cfg.GetDefaultAgent())
+	if opts.agentName != "" {
+		ag = agent.Parse(opts.agentName)
+	}
+	if !opts.noSession {
+		if _, err := exec.LookPath(ag.Binary()); err != nil {
+			fmt.Fprintf(os.Stderr, "%s CLI not found — install %s to create sessions\n", ag.Binary(), ag.DisplayName())
+			os.Exit(1)
+		}
+	}
+
+	name := workspace.SanitizeBranchName(opts.branch)
+	provider := workspace.ResolveProvider(repoPath)
+	if !provider.CanCreate() {
+		fmt.Fprintf(os.Stderr, "This repo's workspace provider can't create worktrees (no create command in .fleet.json)\n")
+		os.Exit(1)
+	}
+
+	info, err := provider.Create(repoPath, name, opts.branch, base)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to create worktree: %v\n", err)
+		os.Exit(1)
+	}
+	if info == nil || info.Path == "" {
+		fmt.Fprintln(os.Stderr, "Worktree provider returned no path")
+		os.Exit(1)
+	}
+
+	if cfg.IsCopyClaudeSettingsEnabled() && !provider.IsCustom() {
+		copyClaudeSettings(repoPath, info.Path)
+	}
+	workspace.CopyConfiguredFiles(repoPath, info.Path)
+
+	// --no-session prints the path and nothing else, so the command composes:
+	// cd "$(fleet worktree my-branch --no-session)"
+	if opts.noSession {
+		fmt.Println(info.Path)
+		return
+	}
+
+	// A CLI-created session needs the agent's hooks installed for status
+	// detection, which normally only happens on TUI launch. Only the chosen
+	// agent's hooks are touched — never create a config dir for an agent that
+	// isn't being launched.
+	installAgentHooks(ag, info.Path)
+
+	storage, err := session.Open(session.DefaultDBPath())
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to open database: %v\n", err)
+		os.Exit(1)
+	}
+	defer storage.Close()
+
+	s := session.NewSession(name, info.Path)
+	s.Agent = ag
+	s.WorkspaceName = name
+	if err := s.Start(); err != nil {
+		fmt.Fprintf(os.Stderr, "Created worktree %s, but failed to start session: %v\n", info.Path, err)
+		os.Exit(1)
+	}
+	if err := storage.SaveSession(s.ToRow()); err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to save session: %v\n", err)
+		os.Exit(1)
+	}
+	// Pin the new checkout so it shows in the sidebar even before it has a
+	// running session, mirroring what the TUI does on session create.
+	if err := storage.PinRepo(session.GetRepoRoot(info.Path)); err != nil {
+		debuglog.Logger.Error("failed to pin repo", "repo", info.Path, "err", err)
+	}
+
+	fmt.Printf("Created worktree %s (branch %s)\n", info.Path, opts.branch)
+	fmt.Printf("Started %s session '%s' (%s)\n", ag.DisplayName(), name, s.ID)
+}
+
+// resolveWorktreeRepo resolves the repo to create the worktree in: the given
+// path (or the current directory), walked up to its git root and then to the
+// main worktree. Basing on the main clone keeps derived worktree paths siblings
+// of it — running this from inside worktree "repo-foo" otherwise yields
+// "repo-foo-bar" — and matches pressing `w` on an origin header in the TUI.
+func resolveWorktreeRepo(path string) (string, error) {
+	if path == "" {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return "", fmt.Errorf("cannot determine current directory: %w", err)
+		}
+		path = cwd
+	}
+	path = expandPath(path)
+	if info, err := os.Stat(path); err != nil || !info.IsDir() {
+		return "", fmt.Errorf("invalid directory: %s", path)
+	}
+	root := repoRootOf(path)
+	if root == "" {
+		return "", fmt.Errorf("not a git repository: %s", path)
+	}
+	return git.GetMainWorktreePath(root), nil
+}
+
+// repoRootOf returns path's git repo root, or "" when path isn't inside a repo.
+// session.GetRepoRoot can't be used here: it returns the input unchanged when
+// git fails, which is indistinguishable from "path is already the root" — so a
+// non-repo would silently proceed into `git worktree add`.
+func repoRootOf(path string) string {
+	ctx, cancel := context.WithTimeout(context.Background(), 8*time.Second)
+	defer cancel()
+	out, err := exec.CommandContext(ctx, "git", "-C", path, "rev-parse", "--show-toplevel").Output()
+	if err != nil {
+		return ""
+	}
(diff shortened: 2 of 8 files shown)


internal/ui/app.go
```diff
@@ -3003,6 +3014,59 @@ func (h *Home) handleSessionCreateResult(msg sessionCreateResultMsg) (tea.Model,
...
+		}
+	}
+	if len(fresh) == 0 {
+		return h, nil
+	}
+
+	// Unlike a session the user just created, an adopted row arrives on a timer.
+	// Inserting it above the cursor would silently move the selection out from
+	// under whoever is mid-keystroke, so re-find the row by identity afterwards.
+	target := h.targetForCursor()
```
hayke102, line 3050: **The cursor can still drift — when it's parked on a "Creating…" phantom.**

`targetForCursor()` knows three row kinds: origin header, checkout header, session (app.go:6926-6933). A `PendingWorkspace` row is none of them — `Session == nil`, no header flags (sidebar.go:294-301). So it returns the zero `contextMenuTarget`, `find` matches no case and returns `-1`, and `h.cursor` is left on its stale numeric index.

That isn't a corner case here: `handleWorkspaceCreate` **auto-selects the phantom** (app.go:1263-1270), so for the entire duration of a TUI-initiated worktree create the cursor sits on exactly the row that can't be re-found — and a shell `fleet worktree` landing in that window is the scenario this PR ships. Adopt a session that sorts above it (new origin group, or a repo earlier in the sort) and the selection silently slides, which is the thing the comment two lines up promises won't happen.

Smallest fix is a fourth case: add `pendingID` to `contextMenuTarget` and teach `targetForCursor`/`find` about `item.Pending`. The context menu never sets it, so nothing else moves.

(Also: `syncViewport()` only runs inside the `idx >= 0` branch, so in this case the viewport isn't re-synced either.)

cmd/fleet/worktree.go
```diff
@@ -0,0 +1,284 @@
...
+	defer storage.Close()
+
+	s := session.NewSession(name, info.Path)
+	s.Agent = ag
+	s.WorkspaceName = name
+	if err := s.Start(); err != nil {
+		fmt.Fprintf(os.Stderr, "Created worktree %s, but failed to start session: %v\n", info.Path, err)
+		os.Exit(1)
+	}
+	if err := storage.SaveSession(s.ToRow()); err != nil {
```
hayke102, line 244: **A failure here leaks a live tmux session that nothing points at.**

`s.Start()` has already created `fleet_<id>`. If `SaveSession` fails we `os.Exit(1)` with the agent running and no DB row — so the TUI will never adopt it, never list it, never offer to delete it. The user has to go find it with `tmux ls`. `busy_timeout` is 5s (storage.go:95), and a TUI writing at the same moment is the realistic trigger, so this isn't purely theoretical.

Two cheap improvements, complementary:
- Kill the tmux session before exiting so the failure is clean (or at least print the tmux session name so it's recoverable).
- Move `session.Open` above `provider.Create` (line 150). Then a DB that can't be opened at all fails *before* a worktree exists on disk, instead of after.

cmd/fleet/worktree.go
```diff
@@ -0,0 +1,284 @@
...
+		fmt.Fprintf(os.Stderr, "Failed to save session: %v\n", err)
+		os.Exit(1)
+	}
+	// Pin the new checkout so it shows in the sidebar even before it has a
+	// running session, mirroring what the TUI does on session create.
+	if err := storage.PinRepo(session.GetRepoRoot(info.Path)); err != nil {
+		debuglog.Logger.Error("failed to pin repo", "repo", info.Path, "err", err)
+	}
+
+	fmt.Printf("Created worktree %s (branch %s)\n", info.Path, opts.branch)
```
hayke102: **`--base` is silently dropped when the branch already exists.**

`GitWorktreeProvider.Create` retries `git worktree add <path> <branch>` without `-b` when the `-b` form fails (provider.go:92-97). So `fleet worktree existing-branch --base develop` checks out the existing branch and the `--base` you typed does nothing — while this line prints the same `Created worktree … (branch existing-branch)` as the fresh-branch case.

In the TUI you'd have seen that branch sitting in the existing-worktrees list before you committed. From a shell there's no signal at all, and "Created … (branch X)" actively reads as "I made branch X off your base".

Worth having `Create` report which path it took, so this can print `reusing existing branch %s` — and ideally note that `--base` was ignored.

internal/ui/app.go
```diff
@@ -3003,6 +3014,59 @@ func (h *Home) handleSessionCreateResult(msg sessionCreateResultMsg) (tea.Model,
...
+
+	h.workerMu.Lock()
+	h.sessions = append(h.sessions, fresh...)
+	h.rebuildSessionMap()
+	h.workerMu.Unlock()
+
+	cmds := make([]tea.Cmd, 0, len(fresh))
+	for _, s := range fresh {
+		repo := session.GetRepoRoot(s.ProjectPath)
+		h.setExpanded(repo, true)
```
hayke102: **Auto-expanding on a timer isn't the same call as auto-expanding on a keypress.**

`setExpanded` persists (`SetGroupCollapsed`, app.go:4094), so an adopted session pops open a group the user deliberately collapsed *and writes that back*. Copying `handleSessionCreateResult` is right when the user just pressed `a` and the row auto-selects; here it fires off a 5s sweep with no input.

The case that actually bites: snoozing a group collapses it (`setExpanded(key, false)`), and that collapse is the visual half of the snooze. Adopting a session into a snoozed group re-expands it without waking it — you end up with an expanded group still carrying a `☾` countdown, i.e. the two halves of snooze disagreeing.

Suggest gating the expand: skip it when the group is snoozed, or when `repoExpanded[key]` is already explicitly `false`.

cmd/fleet/worktree.go
```diff
@@ -0,0 +1,284 @@
...
+	fmt.Printf("Started %s session '%s' (%s)\n", ag.DisplayName(), name, s.ID)
```
hayke102, line 264: **Sessions created here are invisible to telemetry — flagging as a decision, not a defect.**

No `analytics.Track(EventSessionCreated)` on this path, and `handleAdoptSessions` deliberately skips it too. `fleet add` has the same gap today so this is consistent rather than a regression — but `fleet worktree` is pitched as a first-class way to start sessions, so `session_created` (and `onboarding_first_session_created`) will drift low exactly as the feature succeeds.

Firing from the CLI means `analytics.Init` + a `Shutdown` flush, which is real weight for a one-shot command. The cheaper alternative is tracking at adoption time with a `source: "cli"` property — you'd lose the no-TUI-running case, but keep the trend.

hayke102 (commented):
Read the whole diff; `go build ./...` and `go test ./cmd/fleet/... ./internal/ui/...` both pass. Shape is good — the adopt sweep mirrors the suspend sweep's cadence/send-guard exactly, and `unknownSessions` being pure + the demo-prefix parity test is the right call. Five threads below: one real hole in the cursor invariant this PR sets out to hold, one error-path leak in the CLI, and three judgment calls.

internal/ui/app.go
```diff
@@ -5400,6 +5529,11 @@ drainPriority:
 	// which is why it lives here on the worker, not the Update loop.
 	h.maybeSuspendIdleSessions(sessions)
 
+	// 5c. Adopt sessions another fleet process created (e.g. `fleet worktree`
+	// from a shell) since the last sweep. Self-throttled; reads SQLite, which is
+	// why it lives here rather than on the Update loop.
+	h.maybeAdoptExternalSessions(sessions)
```
hayke102, line 5607: **Adoption never runs on an empty fleet — the exact case the feature is for.**

`statusWorkerCycle` returns early ~200 lines above this, at app.go:5330:

```go
if len(sessions) == 0 && len(shells) == 0 {
	return
}
```

This is the only call site of `maybeAdoptExternalSessions` (grep confirms: app.go:5535, nothing else), so with zero sessions *and* zero drawer shells the sweep is unreachable. Every cycle bails before it.

That's not an exotic state — it's fresh install, or "I just deleted my last session". Open `fleet`, run `fleet worktree feature-x` in another pane, and nothing appears until you restart the TUI. The user most likely to drive from the shell is exactly the user with an empty sidebar, and the empty state is where a session appearing would be most visible.

One drawer shell open masks it entirely, which is probably why it survived manual testing.

Fix: hoist the adopt sweep above the early return. It doesn't read `sessions` for anything but the dedupe snapshot, and an empty snapshot is a perfectly valid input to `unknownSessions` — the existing `TestUnknownSessionsRespectsDemoPrefix` already passes `nil` for `known`. Worth a regression test that drives a cycle with zero sessions.

cmd/fleet/worktree.go
```diff
@@ -0,0 +1,284 @@
...
+	return o, nil
+}
+
+func runWorktree(args []string) {
+	opts, err := parseWorktreeArgs(args)
+	if err != nil {
+		if errors.Is(err, flag.ErrHelp) {
+			return // flag package already printed usage
+		}
+		fmt.Fprintln(os.Stderr, err)
```
hayke102, line 124: **Flag errors print twice.**

`flag.ContinueOnError` doesn't suppress output — `Parse` calls `failf`, which writes the error to `fs.Output()` *and* then prints usage. This line writes it a second time. Reproduced against a local build:

```
$ fleet worktree foo --bogus
flag provided but not defined: -bogus
Usage: fleet worktree <branch> [flags]
  -agent string
    	agent to run: claude, codex, or opencode (default: default_agent config)
  ...
flag provided but not defined: -bogus
```

The duplicate lands *after* the usage block, so it reads like a second, separate failure.

`errMissingBranch` and the `ValidateBranchName` / unknown-agent errors are fine — those are returned after `Parse` succeeds, so nothing has printed them yet. So the fix is to skip the re-print only for errors that came out of `Parse`: either wrap them in a sentinel before returning, or set `fs.SetOutput(io.Discard)` and own all the printing here.

cmd/fleet/worktree.go
```diff
@@ -0,0 +1,284 @@
...
+		if errors.Is(err, errMissingBranch) {
+			fmt.Fprintln(os.Stderr, worktreeUsage)
+		}
+		os.Exit(1)
+	}
+
+	// Both the workspace provider and Session.Start log at Info level, and
+	// debuglog's fallback logger writes to stderr — without Init those lines
+	// land on the user's terminal instead of ~/.config/fleet/debug.log.
+	debuglog.Init()
```
hayke102, line 145: **No `migration.Run()` — this command can permanently strand a legacy `brizz-code` config dir.**

`migration.Run()` is called from `runTUI` only (main.go:87-90), with the comment "before debuglog.Init creates `~/.config/fleet/`". This path does `debuglog.Init()` here and `session.Open(session.DefaultDBPath())` further down, both of which create `~/.config/fleet/` — including `state.db`.

Once that exists, `migrateConfigDir` takes the bail-out branch (migrate.go:178-184): *"both legacy and new config dirs have state.db; skipping move"* — and it still writes the marker, so the migration never retries. A user on the old `brizz-code` config whose first-ever `fleet` invocation is `fleet worktree` silently loses their sessions, pins, slot bindings and config to the TUI, with only a `Warn` in a debug log they have no reason to read.

Pre-existing for `fleet add`, so not a regression — but `fleet worktree` is a much more attractive front door, and it's a one-line fix. Cleanest is to hoist `migration.Run()` out of `runTUI` into `main()` next to the existing `migration.AliasLegacyEnv()` (main.go:33), so every subcommand is covered rather than playing whack-a-mole per entry point.

hayke102 (commented):
Second pass — three more, all reproduced locally. The first one is the headline: I missed it on the first read and it defeats the feature in its most likely scenario.

changelog/unreleased/adopt-external-sessions.md
```diff
@@ -0,0 +1,5 @@
+---
+type: improved
+---
+
+**No restart to see new sessions.** A session created from another terminal — `fleet worktree`, `fleet add` — now shows up in a running fleet within ~5 seconds, instead of waiting until the next launch.
```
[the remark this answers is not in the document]
hayke102: Fixed in c01cb28 — headline is now `**No restart needed.**` (3 words). On `highlight`: confirmed `no`, decided before the fragment was written; the file has no `highlight` key and stays out of the What's New reel.

cmd/fleet/worktree.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		if errors.Is(err, flag.ErrHelp) {
+			return // flag package already printed usage
+		}
+		fmt.Fprintln(os.Stderr, err)
+		if errors.Is(err, errMissingBranch) {
+			fmt.Fprintln(os.Stderr, worktreeUsage)
+		}
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
hayke102: Fixed in c01cb28 — reproduced first, then fixed by owning all the printing: `fs.SetOutput(io.Discard)` plus a no-op `fs.Usage`, with the flag registration factored into `worktreeFlagSet` so `printWorktreeUsage` can render the defaults. `runWorktree` now prints error → usage, once each. Note only *parse* errors were doubled — `errMissingBranch`, `ValidateBranchName` and the unknown-agent error return after Parse succeeds, so those kept their single print. Bad values now print the message alone, without the flag dump burying it.

internal/ui/app.go
```diff
@@ -3003,6 +3014,59 @@ func (h *Home) handleSessionCreateResult(msg sessionCreateResultMsg) (tea.Model,
...
+				debuglog.Logger.Error("failed to pin repo", "repo", repo, "err", err)
+			}
+		}
+		// The creating process already spun up the tmux server, so the probe
+		// runs against a live one. Self-guards: nil unless the path sits under
+		// an unprobed macOS-protected folder.
+		if cmd := h.probeTCCCmd(s.ProjectPath); cmd != nil {
+			cmds = append(cmds, cmd)
+		}
+	}
```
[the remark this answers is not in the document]
hayke102: Fixed in c01cb28 — roots are resolved in `maybeAdoptExternalSessions` (already on the worker) and carried in `adoptSessionsMsg.repoRoots`, with a fallback to `ProjectPath`. Your cache-miss reasoning is what makes this bite: `handleSessionCreateResult` calls `GetRepoRoot` on Update too, but there the path is usually warm from the worker's git refresh — for an adopted session the miss is guaranteed by construction, since the path was created by a different process.

internal/ui/app.go
```diff
@@ -5400,6 +5529,11 @@ drainPriority:
 	// which is why it lives here on the worker, not the Update loop.
 	h.maybeSuspendIdleSessions(sessions)
 
+	// 5c. Adopt sessions another fleet process created (e.g. `fleet worktree`
+	// from a shell) since the last sweep. Self-throttled; reads SQLite, which is
+	// why it lives here rather than on the Update loop.
+	h.maybeAdoptExternalSessions(sessions)
+
```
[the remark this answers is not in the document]
hayke102: Fixed in c01cb28 — correct, and this was the most serious finding. The sweep now runs above the early return (gated on `heavy`; the existing call stays for the non-empty path, and the throttle makes the double call safe). Added `TestStatusWorkerCycleAdoptsOnEmptyFleet`, verified failing with the fix reverted and passing with it. My manual verification missed this because the test TUI had one existing session, which put the cycle past the guard.

internal/ui/app.go
hayke102: Fixed in c01cb28 — same as the CodeRabbit thread on this line. You're right that one drawer shell masks it entirely, which is why it survived manual testing. The sweep now runs above the early return, and `TestStatusWorkerCycleAdoptsOnEmptyFleet` drives a zero-session cycle (confirmed failing without the fix). Re-verified live: booted the TUI against a wiped DB so the sidebar showed "No Sessions Yet", ran `fleet worktree` from another pane, and the session appeared without a restart.

cmd/fleet/worktree.go
hayke102: Fixed in c01cb28 — reproduced exactly as you described, duplicate landing after the usage block. Took your framing: only errors out of `Parse` were doubled, so the fix silences the flag package (`SetOutput(io.Discard)` + no-op `Usage`) and `runWorktree` owns all printing. Bad values now print the message alone — the flag dump would bury it — while a bad flag or missing branch still shows the flag list. `-h` prints usage to stdout and exits 0.

internal/ui/app.go
hayke102: Fixed in c01cb28 — took the suggested shape: `pendingID` is a fourth case on `contextMenuTarget`, taught to `targetForCursor`/`find`. The context menu never sets it, so nothing else moved. Also moved `syncViewport()` out of the `idx >= 0` branch — adoption only ever adds rows, so the index stays in range, but the rows under it shifted either way. `TestHandleAdoptSessionsKeepsCursorOnPendingRow` parks the cursor on a phantom and adopts above it; it carries a vacuity guard that fails if the phantom's index doesn't actually shift.

cmd/fleet/worktree.go
hayke102: Fixed in c01cb28 — both, they're independent. `session.Open` moved above `provider.Create`, so an unopenable DB fails before a worktree exists on disk. And a `SaveSession` failure now kills the tmux session before exiting, naming it if the kill also fails so it's recoverable via `tmux ls`. Your point that it's worse than a plain leak is the one that decided it: with no row, the adoption sweep this PR adds can't see it either, so there was no recovery path at all.

internal/ui/app.go
hayke102: Fixed in c01cb28 — the auto-expand is gone, behind a new `canAutoExpand`: skipped when `repoExpanded[key]` is explicitly `false`, and when the checkout or its origin is snoozed. The snooze case is what settled it — an expanded group still carrying a `☾` countdown is the two halves of snooze contradicting each other. Your framing that `setExpanded` *persists*, so this wasn't just a display slip, is the part I'd missed when copying `handleSessionCreateResult`. Covered by `TestCanAutoExpandRespectsCollapseAndSnooze`.

cmd/fleet/worktree.go
hayke102: Fixed in c01cb28, CLI-side rather than through the provider — kept the shared `Provider` (and the TUI path) untouched. Before `Create`, a `git show-ref --verify` on the branch; when it exists the command says so and states that `--base` is being ignored, and the success line reads `(existing branch X)` instead of `(branch X)`. Warnings go to stderr, so `cd "$(fleet worktree … --no-session)"` still composes. Skipped for a custom `ShellProvider`, which defines its own branch semantics.

cmd/fleet/worktree.go
hayke102: Fixed in c01cb28 — `migration.Run()` at the top of `runWorktree`, before `debuglog.Init()`, mirroring `runTUI`'s ordering and its reasoning. Went with the per-command placement rather than hoisting into `main()`: `main()` also dispatches `hook-handler`, which agents spawn on every hook event, so that path would take an `os.Stat` per invocation — and on a machine with no marker yet, the full run, which shells out to tmux. `fleet add`'s identical gap is left for a follow-up as you noted.

cmd/fleet/worktree.go
hayke102: Taken — went with your cheaper option in c01cb28. `handleAdoptSessions` now fires `EventSessionCreated` with `source: "cli"`, plus the `onboarding_first_session_created` milestone you named. One addition: the existing TUI call is now tagged `source: "tui"` — otherwise the property is null on half the events and can't be split on. Accepted trade-off: sessions created while no TUI is running are still missed, so the count is a floor rather than exact.

Issue #220, opened 2026-07-28
Wrong status: showed running, expected idle — codex shows running for no reason


1jonahg1:
## Wrong Status Detected

**fleet showed `running` — should have been `idle`**

### What happened
codex shows running for no reason

### Status Signals
| Signal | Value |
|--------|-------|
| agent | `codex` |
| captured at | 2026-07-28T13:25:30.946781+03:00 |
| hook status | `running` |
| hook event | `Stop` |
| hook age | 18h27m31.947s |
| pane detected | `` |
| TUI showed | `running` |
| **mismatch** | **false** |
| worker stalled | false |
| worker last cycle | 7ms ago |

<details><summary>Screen (last 40 lines)</summary>

```
  1. “Zero behavior change for unrestricted tenants” is true only if the middleware exits immediately when the flag is disabled. Missing edge authentication must not globally reject requests.
  2. “Five token-mint sites” appears to undercount MCP OAuth.

  The production issuance paths I found are:

  - Browser OAuth callback
  - CLI/non-browser OAuth token exchange
  - OAuth refresh
  - API-key-to-JWT exchange
  - SAML ACS
  - MCP OAuth authorization-code exchange
  - MCP OAuth refresh-token exchange

  There is also the development OAuth bypass, which should use the same check for test parity.

  Additionally, apps/backend/internal/routes/mcp/bearer_auth.go must enforce before it mints its loopback token or performs its initial loopback tenant lookup. Checking only afterward would be too late.

  So my refined recommendation is:

  Worker:
    X-Brizz-Edge-Auth
    X-Brizz-Client-IP

  Backend:
    ClientIPResolver
        ↓
    /api/v1 middleware after authentication
        ↓
    Explicit checks at every external token/session issuance path
        ↓
    MCP outer-boundary check before any loopback work

  That is the smallest correct design and has substantially less blast radius than changing Cloud Run ingress or introducing a GCP load balancer.

───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────


› Use /skills to list available skills

  gpt-5.6-sol max fast · ~/brizzai-BRZ-2664-IP-allowlist
```
</details>

### Diagnostics
- **Version**: 2.23.1
- **OS**: macOS 26.5.2 (arm64)
- **tmux**: tmux 3.7b
- **Claude CLI**: 2.1.220 (Claude Code)
- **Codex CLI**: codex-cli 0.145.0
[...]

PR #235, opened 2026-08-02
fix(status): hand the agent pid to the ownership gate

hayke102:
## What

`syncHookStatuses` is the **only** production caller of `UpdateHookStatus`, and it rebuilds the `HookStatus` field by field. `AgentPID` was left out.

So both sides of `conversationSucceeds(ownerPID, hs.AgentPID)` were always `0` → it returns `known=false` → the neg-cached rotation is never released → a session pinned to a conversation that already ended stays pinned for its whole life. `ownerConversationAbandoned` is the only fallback, and it returns `false` on unreadable transcripts.

That is the **v2.24.0 "Stuck sessions recover" fix, shipped inert** (#229).

## Why it wasn't caught

Every test in `ownership_test.go` injects `AgentPID` straight into `UpdateHookStatus`:

```go
s.UpdateHookStatus(&HookStatus{Status: "waiting", SessionID: "live-bbb", AgentPID: livePID}, true)
```

The gate's logic is well covered. The **seam** carrying data into it was not, so the whole feature stayed green while production ran blind.

`TestSyncHookStatusesPropagatesAgentPID` closes that: it writes a real status file, lets the real watcher load it, calls `syncHookStatuses`, and asserts on the reconstructed struct — so a dropped field fails *there*, not in the gate. Verified it fails on the unfixed code with `ownerPID = 0, want 82501`.

## Also: make wrong-status reports self-diagnosing

Found while investigating #220, which arrived showing:

| hook status | hook event |
|---|---|
| `running` | `Stop` |

Those are two different sources under labels implying one — `hook status` is in-memory, `hook event` is read fresh off disk. `mapEventToStatus` maps `Stop`→`finished` deterministically, so no single file can hold that pair: fleet was 18h27m behind the file and not applying it. But the report couldn't say **why** — the conversation ids that decide it are captured at `!` and then dropped before the issue body is built.

The signals table now carries:

- `hook status (applied)` vs `hook status (on disk)`, and a computed **`applied matches disk`**
- `hook session (on disk)` vs `owner session (latched)` — shortened to 8 chars, enough to answer "same or different" without pasting full ids into a public issue
- `owner pid`, and the hook file's own age alongside the applied age

Governing rule is unchanged: allowlisted field reads only, never a marshal of the snapshot map (`hook.file_contents.user_prompt` is the reporter's verbatim prompt). `TestBuildStatusReportBody_NeverLeaksUserPrompt` still passes; the new rows add a full-id leak check of their own.

## Not fixed here

#220's root cause is **still unconfirmed** — it narrows to (a) the ownership gate rejecting the later hooks, or (b) a lost watcher entry, and the deciding field was discarded before the report was written. This PR may well fix it outright if (a); the new rows settle it if it recurs.

**Correction (was wrong in the original description).** I first wrote that a Codex session can never be adopted and stays rejected for its life. That was true *until this PR*. `conversationSucceeds` is pure pid math with no agent gate, and `UpdateHookStatus` has no agent check — so once `AgentPID` flows, a neg-cached Codex pair recovers on pids. `sessionRotationVerdict` still can't resolve a non-Claude rotation (it reads Claude transcripts), so such a pair always neg-caches first and recovery rests entirely on the pid path. Thanks @hayke102 for catching it.

## Test plan

- `go test -race ./...` green
- `TestSyncHookStatusesPropagatesAgentPID` fails on unfixed code, passes here
- `make build` clean

Refs #220

https://claude.ai/code/session_01UJTDzXNbPxHRyzkrEP27Pj

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -5222,9 +5222,18 @@ func (h *Home) syncHookStatuses(sessions []*session.Session, resolveRotation boo
 			oldFirstPrompt := s.FirstPrompt
 			oldPromptCount := s.PromptCount
 			if s.UpdateHookStatus(&session.HookStatus{
-				Status:      hs.Status,
-				SessionID:   hs.SessionID,
-				UpdatedAt:   hs.UpdatedAt,
+				Status:    hs.Status,
+				SessionID: hs.SessionID,
+				UpdatedAt: hs.UpdatedAt,
+				// AgentPID is what lets ownership ask "is the conversation that owns
+				// this session still running?" — the only question that releases a
+				// neg-cached rotation. Dropping it here (this is the sole production
+				// caller of UpdateHookStatus) leaves both sides of that comparison at
+				// 0, so conversationSucceeds answers "unknown" forever and the
+				// process-based recovery never fires. The ownership tests inject
+				// AgentPID straight into UpdateHookStatus, so they stay green while
+				// production runs blind — hence TestSyncHookStatusesPropagatesAgentPID.
+				AgentPID:    hs.AgentPID,
 				UserPrompt:  hs.UserPrompt,
 				PromptCount: hs.PromptCount,
 			}, resolveRotation) {
--- a/internal/ui/statusreport.go
+++ b/internal/ui/statusreport.go
@@ -3,6 +3,7 @@ package ui
 import (
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"charm.land/lipgloss/v2"
@@ -371,7 +372,22 @@ func shortSessionID(id string) string {
 	return id
 }
 
-// metaString/metaBool/metaSub read named fields out of the snapshot.json map.
+// shortPID renders a recorded agent pid, or "-" when there is none.
+//
+// 0 is not a pid. It is what a status file written before StatusFile.AgentPID
+// existed leaves behind — every session already running across an upgrade — and
+// it is exactly the state in which conversationSucceeds returns known=false, so
+// the pid recovery cannot run at all. Printing "0" spells "the mechanism was
+// unavailable" as "here is the pid", which is the opposite of what this row is
+// for. "-" matches how shortSessionID spells an absent id.
+func shortPID(pid int) string {
+	if pid <= 0 {
+		return "-"
+	}
+	return strconv.Itoa(pid)
+}
+
+// metaString/metaInt/metaSub read named fields out of the snapshot.json map.
 //
 // Everything filed goes through these. The map is NOT marshalled wholesale —
 // it embeds hook.file_contents.user_prompt, the reporter's verbatim prompt
@@ -384,6 +400,23 @@ func metaSub(m map[string]any, key string) map[string]any {
 	}
 	sub, _ := m[key].(map[string]any)
 	return sub
+}
+
+// metaInt reads a numeric field. The map is built in-process (int), but a
+// snapshot.json round-tripped through encoding/json yields float64 — accept both
+// so a reader of a persisted capture gets the same answer as the live dialog.
+func metaInt(m map[string]any, key string) int {
+	if m == nil {
+		return 0
+	}
+	switch v := m[key].(type) {
+	case int:
+		return v
+	case float64:
+		return int(v)
+	default:
+		return 0
+	}
 }
 
 func metaString(m map[string]any, key string) string {
@@ -453,21 +486,37 @@ func buildStatusReportBody(desc string, expected session.Status, f *statusReport
 	if ev := metaString(metaSub(hook, "file_contents"), "event"); ev != "" {
 		fmt.Fprintf(&b, "| hook event (on disk) | `%s` |\n", ev)
 	}
+	// Only a divergence too large to be lag gets the bold. Applied lags disk by
+	// design, so a capture landing mid-write reads false benignly — bolding that
+	// trains people to quote the row without reading the ages that qualify it.
 	if am := metaString(hook, "applied_matches_file"); am != "" {
-		fmt.Fprintf(&b, "| **applied matches disk** | **%s** |\n", am)
+		switch {
+		case am != "false":
+			fmt.Fprintf(&b, "| applied matches disk | %s |\n", am)
+		case metaString(hook, "divergence_significant") == "true":
+			fmt.Fprintf(&b, "| **applied matches disk** | **false** (applied hook is %s old) |\n", metaString(hook, "age"))
+		case metaString(hook, "file_newer_by") != "":
+			fmt.Fprintf(&b, "| applied matches disk | false (file %s newer — in flight) |\n", metaString(hook, "file_newer_by"))
+		default:
+			fmt.Fprintf(&b, "| applied matches disk | false |\n")
+		}
 	}
 	fmt.Fprintf(&b, "| hook age (applied) | %s |\n", metaString(hook, "age"))
 	if fa := metaString(hook, "file_age"); fa != "" {
 		fmt.Fprintf(&b, "| hook file age (on disk) | %s |\n", fa)
 	}
-	// Ownership: a hook whose session_id is not the owner is dropped, and for a
-	// non-Claude agent it can never be adopted (the rotation check reads Claude
-	// transcripts), so it is dropped for the session's life. Short ids — enough to
-	// answer "same or different", which is the whole question.
+	// Ownership: a hook whose session_id is not the owner is dropped. For a
+	// non-Claude agent the transcript check that decides "rotation or nested child?"
+	// can never resolve — it reads Claude transcripts — so the pair is neg-cached
+	// and recovery falls entirely to comparing the two agent pids
+	// (conversationSucceeds, which has no agent gate). That is why the pid sits here
+	// beside the ids: without it there is no telling a gate that could not decide
+	// from one that decided against. Short ids — enough to answer "same or
+	// different", which is the whole question.
 	if hookSess, ownerSess := metaString(hook, "file_session_id"), metaString(hook, "owner_session_id"); hookSess != "" || ownerSess != "" {
 		fmt.Fprintf(&b, "| hook session (on disk) | `%s` |\n", shortSessionID(hookSess))
 		fmt.Fprintf(&b, "| owner session (latched) | `%s` |\n", shortSessionID(ownerSess))
-		fmt.Fprintf(&b, "| owner pid | %s |\n", metaString(hook, "owner_pid"))
+		fmt.Fprintf(&b, "| owner pid | %s |\n", shortPID(metaInt(hook, "owner_pid")))
 	}
 	if ov := metaString(hook, "overridden_at"); ov != "" {
 		fmt.Fprintf(&b, "| hook overridden | %s ago |\n", ov)
--- a/internal/ui/statusreport.go
+++ b/internal/ui/statusreport.go
@@ -356,6 +356,21 @@ func sanitizeHome(s string) string {
 	return strings.ReplaceAll(s, home, "~")
 }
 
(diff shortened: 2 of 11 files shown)


hayke102 (commented):
Reviewed the wiring and the report changes. **The core fix is right** — `syncHookStatuses` is the only production constructor of `session.HookStatus` (grep-verified), `AgentPID` was the sole field being dropped, and the new guard test earns its place: removing the line makes it fail with `ownerPID = 0, want <pid>`, so it fails on a dropped field rather than on gate logic. `go build ./...` + `go test ./internal/...` pass.

Four notes below, none blocking. The first is the one I'd actually change.

internal/ui/statusreport.go
```diff
@@ -426,11 +441,34 @@ func buildStatusReportBody(desc string, expected session.Status, f *statusReport
...
+		fmt.Fprintf(&b, "| hook file age (on disk) | %s |\n", fa)
+	}
+	// Ownership: a hook whose session_id is not the owner is dropped, and for a
+	// non-Claude agent it can never be adopted (the rotation check reads Claude
+	// transcripts), so it is dropped for the session's life. Short ids — enough to
+	// answer "same or different", which is the whole question.
+	if hookSess, ownerSess := metaString(hook, "file_session_id"), metaString(hook, "owner_session_id"); hookSess != "" || ownerSess != "" {
+		fmt.Fprintf(&b, "| hook session (on disk) | `%s` |\n", shortSessionID(hookSess))
+		fmt.Fprintf(&b, "| owner session (latched) | `%s` |\n", shortSessionID(ownerSess))
+		fmt.Fprintf(&b, "| owner pid | %s |\n", metaString(hook, "owner_pid"))
```
hayke102: **`owner pid | 0` reads as a real pid — and 0 is the reading that matters most.**

`metaString` renders `int` 0 as `"0"`, so an unknown owner prints as a plausible pid. Two ways to land there:

- The block is entered on `file_session_id` alone, so a session where fleet never claimed ownership renders `owner session (latched) | -` directly above `owner pid | 0` — the two adjacent rows disagreeing about how to spell "absent".
- More common: `session.go:81` documents 0 as *"unknown: pre-AgentPID status file"* and records it as-is. That's every session already running across an upgrade — including, right now, every hook file on my machine (all `agent_pid` absent, because the installed handler predates the field).

Why it's worth more than cosmetics: `conversationSucceeds` returns `known=false` when `ownerPID <= 0`, so the pid-based recovery **cannot fire at all** in that state — it falls to `ownerConversationAbandoned`. This row exists to tell a maintainer whether the gate could have released; printing `0` disguises "the mechanism was unavailable" as "here is the pid."

`shortSessionID` already solved this for ids. Suggest the same for the pid — render `-` (or `unknown`) when it's 0. `TestBuildStatusReportBody_OmitsDivergenceRowsWithoutHookFile` already encodes that rule for the neighbouring rows.

internal/ui/statusreport.go
```diff
@@ -426,11 +441,34 @@ func buildStatusReportBody(desc string, expected session.Status, f *statusReport
...
+	}
+	if am := metaString(hook, "applied_matches_file"); am != "" {
+		fmt.Fprintf(&b, "| **applied matches disk** | **%s** |\n", am)
+	}
+	fmt.Fprintf(&b, "| hook age (applied) | %s |\n", metaString(hook, "age"))
+	if fa := metaString(hook, "file_age"); fa != "" {
+		fmt.Fprintf(&b, "| hook file age (on disk) | %s |\n", fa)
+	}
+	// Ownership: a hook whose session_id is not the owner is dropped, and for a
+	// non-Claude agent it can never be adopted (the rotation check reads Claude
```
hayke102: **This claim is falsified by this very commit.**

> for a non-Claude agent it can never be adopted (the rotation check reads Claude transcripts), so it is dropped for the session's life

True of the transcript path, but that's no longer the only path out. `conversationSucceeds` (`agent_process.go`) is pure pid comparison with no agent gate, and `UpdateHookStatus`'s ownership block doesn't gate on agent either — the only `s.Agent ==` checks in `session.go` are at 175/977/989, all elsewhere.

So for Codex/OpenCode: the transcript verdict can't resolve → neg-cached after the retry cap → the *next* hook lands in the `negCached` branch (`session.go:487`), which now reaches `conversationSucceeds` with two real pids and can adopt. The AgentPID this PR wires up is exactly what makes that reachable for non-Claude agents.

The cost of leaving it: a future debugger reading this row's rationale rules out the one recovery path the commit just switched on.

internal/ui/statusreport.go
```diff
@@ -426,11 +441,34 @@ func buildStatusReportBody(desc string, expected session.Status, f *statusReport
...
+	fmt.Fprintf(&b, "| hook status (applied) | `%s` |\n", metaString(hook, "status"))
+	if fs := metaString(hook, "file_status"); fs != "" {
+		fmt.Fprintf(&b, "| hook status (on disk) | `%s` |\n", fs)
+	}
 	if ev := metaString(metaSub(hook, "file_contents"), "event"); ev != "" {
-		fmt.Fprintf(&b, "| hook event | `%s` |\n", ev)
+		fmt.Fprintf(&b, "| hook event (on disk) | `%s` |\n", ev)
+	}
+	if am := metaString(hook, "applied_matches_file"); am != "" {
+		fmt.Fprintf(&b, "| **applied matches disk** | **%s** |\n", am)
```
hayke102: **Bolded verdict, no staleness qualifier.**

`applied_matches_file` compares two sources read at the same instant, and the applied side lags by design — fsnotify delivery plus a worker cycle. A hook written by the agent in that window produces a benign `false`, identical to the value the pathological dropped-hook case produces. Since this is the row rendered in bold, it's the one people will quote out of the table.

`hook age (applied)` / `hook file age (on disk)` do disambiguate it, but only for a reader who consults them — which is the habit the bold discourages. Worth either qualifying the row when the applied hook is fresh (sub-second / under a cycle), or dropping the bold so it reads as one signal among several rather than the verdict.

internal/ui/app.go
```diff
@@ -5222,9 +5222,18 @@ func (h *Home) syncHookStatuses(sessions []*session.Session, resolveRotation boo
...
+				UpdatedAt: hs.UpdatedAt,
+				// AgentPID is what lets ownership ask "is the conversation that owns
+				// this session still running?" — the only question that releases a
+				// neg-cached rotation. Dropping it here (this is the sole production
+				// caller of UpdateHookStatus) leaves both sides of that comparison at
+				// 0, so conversationSucceeds answers "unknown" forever and the
+				// process-based recovery never fires. The ownership tests inject
+				// AgentPID straight into UpdateHookStatus, so they stay green while
+				// production runs blind — hence TestSyncHookStatusesPropagatesAgentPID.
+				AgentPID:    hs.AgentPID,
```
hayke102, line 5236: Not a problem with this line — flagging it *here* because this line is what makes the pid path execute in production for the first time, so #229's assumption gets its first real test on merge.

`hook_handler.go:183` takes the agent pid as `os.Getppid()`. I checked all three install paths and they're fine:

- Claude + Codex share `hookCommand()` (`claude_hooks.go:137`) → a simple `<quoted-path> hook-handler --fleet-hook`. Verified on this box that `sh -c '<bin>'` **execs** rather than forks (ppid came back as the invoking shell, no intermediate process), so `getppid()` is the agent.
- OpenCode uses `spawnSync(FLEET_BIN, [...])` — no shell in the picture at all.

The one way it inverts: any *compound* command (a `;`, `&&`, `|`, or redirection) makes `sh` fork instead of exec. Then `ownerPID` is a shell that exits milliseconds later, `agentProcessAlive` is false forever, and `conversationSucceeds` returns `(true, true)` for **every** foreign session id — adopting precisely the nested-child conversation the gate exists to reject. Only reachable via a hand-edited `settings.json` today, so not urgent; but given the failure mode is "gate silently means the opposite," a cheap sanity check on the recorded pid (or a note in `conversationSucceeds` that it trusts the command to be exec-shaped) seems worth the line.

internal/ui/statusreport.go
hayke102: Fixed — added `shortPID()` alongside `shortSessionID`, rendering `-` for anything `<= 0`.

Confirmed your point empirically before fixing: I sampled the live hook files on this machine and **20 of 20 have no `agent_pid`**, so on any upgrade-in-place every row would have read `0`. Agreed the semantic argument is the stronger one — `0` is precisely the state where `conversationSucceeds` returns `known=false` and the pid recovery can't run, so rendering it as a pid inverted the row's meaning.

`TestBuildStatusReportBody_UnknownOwnerPIDRendersAsAbsent` pins `0 → -` (and asserts `| owner pid | 0 |` never appears); `TestShortPID` covers 0/-1/4242.

internal/ui/statusreport.go
hayke102: Fixed — you're right, and it's worse than a stale comment: I asserted the same thing in the PR description and on issue #220. Correcting both.

Verified your reasoning: `conversationSucceeds` (`agent_process.go:30`) takes only two pids, and `UpdateHookStatus` has no `s.Agent ==` check anywhere in its body — the ones in `session.go` are at 175/977/989, all in `initialRunStatus`/`UpdateStatus`. So for Codex the transcript verdict can't resolve → `rotationUnknown` ×10 → neg-cached → the next hook enters the `negCached` branch and adopts on two real pids.

The comment was true pre-PR (both pids were always 0, so `known=false` and it fell to `ownerConversationAbandoned`, which returns false on unreadable transcripts). Wiring AgentPID is exactly what falsifies it. Rewritten to say the transcript check can't resolve a non-Claude rotation, so recovery falls to the pid comparison — which is why the pid is in that table at all.

internal/ui/statusreport.go
hayke102: Fixed — qualified by skew rather than dropping the bold, so a real divergence keeps announcing itself.

`buildSnapshotJSON` now computes `file_newer_by` (file mtime minus the applied hook's timestamp) and a `divergence_significant` flag. The row renders three ways:

```
| **applied matches disk** | **false** (applied hook is 18h27m old) |   ← significant
| applied matches disk | false (file 412ms newer — in flight) |        ← benign lag
| applied matches disk | true |
```

Grace is 3s, and the sizing is less obvious than the pipeline cost suggests: the applied timestamp comes from the status file's `ts`, which is Unix **seconds**, while mtime is sub-second — so re-reading the very same file yields skew in [0, 1s) with nothing wrong. 3s clears that plus the ~600ms debounce+cycle, and the case this catches is off by hours.

Two tests: `TestBuildStatusReportBody_BoldsOnlySignificantDivergence` for both renderings, and `TestBuildSnapshotJSON_DivergenceSignificance` for the producer — including the second-granularity artifact explicitly. That second one exists because of this PR's own lesson: the renderer tests drive off literal maps, so nothing would have pinned how the flag is actually computed.

internal/ui/app.go
hayke102: Documented the assumption rather than adding a runtime check — agreed it's non-urgent given only a hand-edited `settings.json` reaches it, but the failure shape deserves naming: the gate *inverts* rather than degrades, adopting every nested child, which reads as correct behavior from the outside.

Two comments:

- `StatusFile.AgentPID` now states that `getppid()` names the agent only because the installed command is a single simple command, cites both install paths (`GetHookCommand()`'s `'<path>' hook-handler --fleet-hook` through a shell, which execs; OpenCode's `spawnSync` with no shell), and spells out that a compound command records the shell instead.
- `conversationSucceeds` now says it trusts both pids to be agent processes per that contract, and describes how it fails if broken — a short-lived pid is never alive, so case 3 collapses into case 2.

Confirmed your read of the install paths independently before writing it. A real `comm`-based check felt like its own PR; happy to open one if you'd rather have it than the note.

PR #237, opened 2026-08-02
fix(github): stop a superseded run's failure from reddening the badge

hayke102:
## Problem

A red PR badge is supposed to mean *go look at this*. On a repo whose heavy tests are gated behind a `/e2e` comment, it didn't.

The gate job publishes its verdict on push — before the suites it gates have been triggered — so it goes red. Commenting `/e2e` starts a second run of the same workflow, which re-publishes the verdict green at the end. In between, `latestRunPerCheck` can't collapse the two: they're different job names (`heavy_gate` vs `E2E / E2E Tests`), so the stale red survived and the badge stayed FAILURE for the entire E2E run — pointing at a failure the workflow was already recomputing.

## Fix

Skip a failure when a **newer check of the same workflow is still running**. The badge reports PENDING (wait) instead of FAILURE (go look).

Observed shape:

| check | workflow | started | state |
|---|---|---|---|
| `heavy_gate` | CI - Heavy Tests | 15:45:07 | FAILURE |
| `E2E / E2E Tests` | CI - Heavy Tests | 15:49:12 | IN_PROGRESS |

Scoping is per workflow, so:
- a failure from a workflow that isn't re-running still goes red immediately
- another workflow running never excuses this one's failure
- status contexts (empty `workflowName`) are excluded — they'd share one key and suppress each other unrelated
- an ignored (`pr_checks.ignore`) in-flight check can't suppress anything, since the filter runs first

Requires `workflowName` from the rollup, which `gh pr view` already returns.

## Test

`TestDeriveCIStatusSupersededFailure` covers all six cases above. `go test -race ./internal/github/` and `make build` pass.

https://claude.ai/code/session_01CpQT3wMAS2wRGt4RoE83Ss

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented outdated failed checks from keeping a pull request blocked when a newer run for the same workflow is in progress.
  * Preserved accurate reporting for active failures, pending checks, ignored checks, and unrelated workflow failures.

* **Documentation**
  * Added an unreleased changelog entry describing the stale CI failure fix.

### DIFF
--- a/changelog/unreleased/stale-ci-failure.md
+++ b/changelog/unreleased/stale-ci-failure.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**Red PR badge means act now** — a check that failed in an earlier run no longer shows red while the same workflow is running again, so a gate that goes red before its suites are triggered reads as pending until the real verdict lands.
+**Stale CI failures** — A PR badge no longer goes red for a failure a later run of the same workflow is already recomputing; you see pending until the new verdict lands.
--- a/changelog/unreleased/stale-ci-failure.md
+++ b/changelog/unreleased/stale-ci-failure.md
@@ -1,0 +1,5 @@
+---
+type: fixed
+---
+
+**Red PR badge means act now** — a check that failed in an earlier run no longer shows red while the same workflow is running again, so a gate that goes red before its suites are triggered reads as pending until the real verdict lands.
--- a/internal/github/pr.go
+++ b/internal/github/pr.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"os/exec"
 	"path"
+	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -106,6 +108,7 @@ type statusCheckEntry struct {
 	Conclusion   string `json:"conclusion"`
 	StartedAt    string `json:"startedAt"`    // RFC3339; used to pick the latest run when a check has several
 	WorkflowName string `json:"workflowName"` // empty for status contexts; called workflows report their caller
+	DetailsURL   string `json:"detailsUrl"`   // .../actions/runs/<run>/job/<job> for check runs; carries run identity
 }
 
 // GetPRForBranch returns the PR associated with the current branch, or nil if none.
@@ -280,17 +283,42 @@ func isInFlightCheck(c statusCheckEntry) bool {
 	return c.Conclusion == ""
 }
 
-// newestInFlightPerWorkflow returns, per workflow, the StartedAt of the most
-// recently started check still running. Status contexts (no workflow) are
-// excluded: they'd all share the "" key and suppress each other unrelated.
-func newestInFlightPerWorkflow(checks []statusCheckEntry) map[string]string {
-	newest := make(map[string]string)
+var runIDPattern = regexp.MustCompile(`/actions/runs/(\d+)`)
+
+// runID extracts the workflow run id from a check run's details URL
+// (`.../actions/runs/<run_id>/job/<job_id>`), or 0 when there is none —
+// status contexts point wherever their publisher likes.
+//
+// Run identity has to come from here because the rollup's own timestamps are
+// per *job*: jobs of one run start seconds to minutes apart (`needs:` chains,
+// runner acquisition), so comparing StartedAt cannot tell a later run from a
+// later-starting sibling of the same run.
+func runID(c statusCheckEntry) int64 {
+	m := runIDPattern.FindStringSubmatch(c.DetailsURL)
+	if m == nil {
+		return 0
+	}
+	id, err := strconv.ParseInt(m[1], 10, 64)
+	if err != nil {
+		return 0 // absurdly long digit string; treat as unidentified
+	}
+	return id
+}
+
+// newestInFlightRunPerWorkflow returns, per workflow, the highest run id that
+// still has a check running. Run ids ascend over time within a repo, so a
+// greater id is a later run.
+//
+// Status contexts (no workflow) are excluded: they'd all share the "" key and
+// suppress each other unrelated.
+func newestInFlightRunPerWorkflow(checks []statusCheckEntry) map[string]int64 {
+	newest := make(map[string]int64)
 	for _, c := range checks {
 		if c.WorkflowName == "" || !isInFlightCheck(c) {
 			continue
 		}
-		if c.StartedAt > newest[c.WorkflowName] {
-			newest[c.WorkflowName] = c.StartedAt
+		if id := runID(c); id > newest[c.WorkflowName] {
+			newest[c.WorkflowName] = id
 		}
 	}
 	return newest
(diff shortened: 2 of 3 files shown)


internal/github/pr.go
```diff
@@ -263,26 +264,72 @@ func latestRunPerCheck(checks []statusCheckEntry) []statusCheckEntry {
...
+	inFlight := newestInFlightPerWorkflow(considered)
 
-		if conclusion == "FAILURE" || conclusion == "ERROR" || conclusion == "TIMED_OUT" {
+	hasFailure := false
+	hasPending := false
+
+	for _, check := range considered {
+		switch {
+		case isFailedCheck(check):
+			if since, ok := inFlight[check.WorkflowName]; ok && check.StartedAt < since {
```
hayke102: **The suppression key can't distinguish "a newer run of this workflow" from "a sibling job of the same run."**

`inFlight` is keyed on `WorkflowName` and compared by `StartedAt`, and neither carries run identity — `startedAt` on a rollup entry is the **check run's (job's)** start time, not the run's. So any in-flight job of the same workflow that merely started later suppresses a failure, including a job of the very same run.

Concrete: workflow `CI` starts one run with `lint` (starts 10:00:05, FAILURE at 10:00:30) and `e2e` (starts 10:00:07, 10-minute suite). `inFlight["CI"] = 10:00:07`, `lint.StartedAt = 10:00:05 < 10:00:07`, so the real failure is dropped and `deriveCIStatus` returns `PENDING`. The badge stays yellow for the rest of the run even though the verdict is final and actionable — the inverse of what the fragment promises ("Red PR badge means act now").

This fires for essentially any multi-job workflow: `needs:`-chained jobs start minutes apart, and even parallel jobs stagger by seconds on runner acquisition. A failure escapes suppression only if it happened to be the last-started job in its workflow. The single-run case is much more common than the re-run case this targets.

Two sub-cases of the same root cause:
- Per the new field comment, called/reusable workflows report their **caller** as `workflowName`, so two unrelated reusable workflows invoked from one caller collapse to one key — an in-flight job in reusable workflow A suppresses a real failure in reusable workflow B.
- A failed entry with a missing/empty `startedAt` compares as `"" < since` and is suppressed by any in-flight sibling.

**Suggested fix:** key the suppression on actual run identity instead of workflow + timestamp. `detailsUrl` on a CheckRun is `.../actions/runs/<run_id>/job/<job_id>`, so requesting that field and extracting `<run_id>` lets you suppress only failures whose run id differs from a run that currently has in-flight jobs. The same-run case then can't trigger, and both sub-cases above go away with it.

---
The rest of the change reads correctly: `continue` inside the `switch` targets the `for` loop; a suppressed failure can never yield `SUCCESS` (the in-flight sibling that caused the suppression is in `considered`, so `hasPending` is always set); the ignore filter runs before `newestInFlightPerWorkflow`; and the extracted `isFailedCheck`/`isInFlightCheck` preserve the original predicates and their precedence.

internal/github/pr_test.go
```diff
@@ -206,3 +206,93 @@ func TestDeriveCIStatus(t *testing.T) {
...
+// workflow is already recomputing it.
+func TestDeriveCIStatusSupersededFailure(t *testing.T) {
+	tests := []struct {
+		name   string
+		checks []statusCheckEntry
+		ignore []string
+		want   string
+	}{
+		{
+			"stale gate failure while its workflow re-runs",
```
hayke102: This case is structurally identical to the same-run scenario in the thread on `pr.go` — a gate job that started earlier plus a later-starting in-flight job, same workflow. Nothing in the fixture says the two came from different runs, because nothing in `statusCheckEntry` can. So this test passes whether the suppression is run-aware or not, and it won't catch the same-run regression.

Once suppression keys on run id, this fixture wants two distinct run ids to keep asserting what its name says.

internal/github/pr.go
hayke102: Fixed in e1cad62 — you're right, and it's the common case, not the edge one. Verified against live data: every job in one `CI - Merge Gatekeeper` run shares run id `30754985868` but starts 40s+ apart, so any early failure was suppressed by a later-starting sibling of its own run.

Suppression now keys on run identity parsed from `detailsUrl` (`.../actions/runs/<id>/job/<id>`) and fires only on a strictly greater run id. Same-run siblings compare equal so they can't suppress, and a failure whose run id doesn't parse is never suppressed — an unrecognized URL shape costs a false red, not a missed one. The reusable-workflow sub-case goes with it: called workflows report the caller's name but share its run id.

internal/github/pr_test.go
hayke102: Fixed — fixtures now carry `DetailsURL` run ids, so this case actually encodes "a different run". Added the regression it couldn't express: one run id, failure started earlier than an in-flight sibling, must stay FAILURE (it fails against the old timestamp logic). Also added an unparseable-run-id case. Timestamps across the table are arranged so a StartedAt compare answers wrong in both directions.

changelog/unreleased/stale-ci-failure.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Red PR badge means act now** — a check that failed in an earlier run no longer shows red while the same workflow is running again, so a gate that goes red before its suites are triggered reads as pending until the real verdict lands.
```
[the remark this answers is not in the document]
hayke102: Rewrote the fragment with a 3-word headline in second person:

> **Stale CI failures** — A PR badge no longer goes red for a failure a later run of the same workflow is already recomputing; you see pending until the new verdict lands.

Not taking the suggested wording verbatim: "when 1 newer run … is active" invents a number to satisfy the rule. The guideline asks for numbers *over adjectives* (`~150ms instead of 2s`) — a count of 1 measures nothing. On `highlight`: no, already decided with the author.

internal/github/pr.go
```diff
@@ -101,10 +101,11 @@ type ghPRResponse struct {
...
 type statusCheckEntry struct {
-	Name       string `json:"name"`
-	Status     string `json:"status"`
-	Conclusion string `json:"conclusion"`
-	StartedAt  string `json:"startedAt"` // RFC3339; used to pick the latest run when a check has several
+	Name         string `json:"name"`
+	Status       string `json:"status"`
+	Conclusion   string `json:"conclusion"`
+	StartedAt    string `json:"startedAt"`    // RFC3339; used to pick the latest run when a check has several
+	WorkflowName string `json:"workflowName"` // empty for status contexts; called workflows report their caller
```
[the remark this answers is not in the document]
hayke102: Not doing this one, but it's narrower now than when you wrote it. Keying on run id (see the pr.go:328 thread) closes both halves you'd hit in practice: same-run siblings and reusable workflows, which report the caller's display name but share its run id.

What's left needs two distinct workflow *files* declaring the same `name:`, where a newer run of one masks an older failure of the other — and the symptom is PENDING instead of FAILURE until that run ends, not a wrong green.

Closing it means dropping `gh pr view --json statusCheckRollup` (fixed shape: name/status/conclusion/startedAt/completedAt/detailsUrl/workflowName — no workflow path) for a hand-written GraphQL query on `checkSuite.workflowRun.workflow.path`, i.e. a new failure surface on a call that runs every 60s per repo. Not worth it for duplicate workflow names.

PR #240, opened 2026-08-04
feat(accounts): run sessions under multiple Claude subscriptions

hayke102:
Adds a first-class **account** concept: fleet knows about several Claude subscriptions, assigns each session one based on remaining quota, and shows what's left of each.

Closes the problem where a spent 5-hour window stalled every Claude session at once while a second subscription sat idle.

## How it works

One environment variable per tmux session:

```
tmux new-session -e FLEET_INSTANCE_ID=… -e CLAUDE_CODE_OAUTH_TOKEN=sk-ant-oat01-…
```

Set in `Session.sessionEnv()` — the single chokepoint already feeding `Start`, `Restart` and `RespawnClaude`. It overrides **auth only**, so `~/.claude` stays shared and hooks, `projects/` (auto-naming + `--resume`), skills, plugins and local MCP servers are untouched.

`CLAUDE_CONFIG_DIR` was rejected precisely because it forks all of that.

## Using it

`Ctrl+K` → **Manage Claude Accounts** → `a`. Fleet runs `claude setup-token` in a real pane, watches for the token, and pulls you back automatically once it appears — no `Ctrl+Q`, no copy-paste. New sessions then go to whichever account has the most 5-hour headroom.

Strategy is configurable in Settings → Behavior: `least_used` (default), `waterfall`, or `manual`.

## Things measured rather than assumed

This took several wrong turns, all from guessing at API behaviour. The findings are in the commit messages and code comments so nobody repeats them:

- **`claude setup-token` mints an inference-only credential.** Anthropic's wording: "limited to inference-only for security reasons". `/api/oauth/usage` and `/api/oauth/profile` both answer 403 — permanently, by design. Not a bug to work around.
- **`claude auth status --json` cannot identify such a token.** It answers `{"loggedIn":true,"authMethod":"oauth_token"}` and nothing else — and returns `loggedIn:true` for a string of 95 `A`s, so it verifies nothing either.
- **401 vs 403 is the real discriminator.** A bogus token gets 401; a valid-but-scope-limited one gets 403. Treating 403 as rejection refused every good token.
- **Quota is in the response headers.** `anthropic-ratelimit-unified-5h-utilization` et al. arrive on any Messages response, needing no special scope. That's where quota now comes from — a `max_tokens=1` haiku probe, roughly nine tokens per account per 180s.
- **Those headers are `0..1`; the usage endpoint is `0..100`.** Two other projects document being bitten by this. Tested both ways.
- **`capture-pane` without `-J` truncates at every pane width** — the prompt eats columns before the value starts. Measured 38/58/98 characters of a 108-character token at 60/80/120 columns. Regression test drives a real tmux pane.

## Two judgement calls worth reviewing

**Fleet identifies as `fleet/<version>`, not `claude-code/<version>`.** Measured: the endpoint serves both identically, and serves a request with no User-Agent at all. There was nothing to gain by impersonating the CLI, so it doesn't. (Community tools split on this; claude-swap, the most used, also sends its own name.)

**Auto-naming matches the token's org against `~/.claude.json`.** Claude Code caches the local login's profile there. If a token's `anthropic-organization-id` matches, the email is already in the user's own file and fleet uses it. The reasoning for why this is fair — and where the line is — is in `identity.go`: the scope limit stops a *leaked* token identifying its owner, and this match only works for someone already reading the user's home directory. Every other account is named by hand.

## Security

Tokens are year-long credentials, so they live in `~/.config/fleet/accounts.json` (0600), **deliberately not `config.json`**, which the bug-report flow publishes into public GitHub issues.

`sanitizeHome` is renamed `sanitizeForIssue` and now redacts `sk-ant-*` at that chokepoint, so no call site can forget. Raw pane captures use a looser pattern, since a token cut off mid-print falls under the strict one's length floor. Guarded by tests in both packages — the add-account pane genuinely puts a credential on screen.

`GuardConflictingAuth` refuses to launch when `ANTHROPIC_API_KEY` or `ANTHROPIC_AUTH_TOKEN` is set: those outrank our token, so the session would silently bill elsewhere.

## Also fixes

`dispatchForkToWorktree` never set `agent` at all — `F` fork-to-worktree fell back to the config default instead of inheriting from the parent, masked only by its Claude-only guard.

## Not included

- **Origin allowlist UI.** `allowed_accounts` is plumbed and tested (keyed `origin:<key>`, reusing the `collapsed_groups` keyspace), but config-file only. With two accounts it has one meaningful state.
- **Regenerating a token creates a second entry** rather than updating — the fingerprint changes. Delete-then-add is the right order; bites once a year at most.

## Verification

`make build`, `go test -race ./...`, `golangci-lint` (0 issues). Exercised live against two real subscriptions: capture → auto-name → store → poll quota → assign on headroom → confirmed the token reached the tmux pane by hash, and that `/status` inside reports `authMethod: oauth_token`.

https://claude.ai/code/session_01SExXRoA82UQVRpJ9HBmYAe

## Summary by CodeRabbit

- **New Features**
  - Added multi-account Claude management, including login, renaming, reordering, removal, and default selection.
  - Sessions automatically select accounts with available quota, inherit accounts when forked, and support manual account moves.
  - Added five-hour and weekly usage displays, reset countdowns, selection strategies, per-origin allowlists, and `fleet worktree --account`.
  - Added account indicators to session previews and an option to hide usage displays.

- **Bug Fixes**
  - Sessions recover from logged-out or exhausted accounts when possible.
  - Improved wrapped terminal output capture and credential redaction in reports.


internal/ui/accounts_header.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+const accountLabelMax = 10
+
+func truncateLabel(s string) string {
+	if lipgloss.Width(s) <= accountLabelMax {
+		return s
+	}
+	r := []rune(s)
+	return string(r[:accountLabelMax-1]) + "…"
```
hayke102: `truncateLabel` can panic here. `lipgloss.Width` counts *display* width (double-width runes count as 2), but the slice below is indexed by *rune count*.

A label of 6 double-width characters (e.g. CJK/emoji) has `Width == 12` (> `accountLabelMax`, so truncation triggers) but `len(r) == 6`. `r[:accountLabelMax-1]` is `r[:9]` on a 6-element slice — out-of-range panic, crashing the whole render loop for every session.

Reachable via the account rename flow (`accountsRename` state) any time a user sets a label using wide characters.

internal/ui/bugreport.go
```diff
@@ -316,7 +316,7 @@ func (d *BugReportDialog) openGitHubIssue(description string) tea.Cmd {
 		// action log, and debug log are reproduction context for a defect; on an
 		// idea they are noise the maintainer has to scroll past.
 		label = "enhancement"
-		body = "## Feature Request\n\n### Problem\n" + sanitizeHome(description) + "\n\n" +
+		body = "## Feature Request\n\n### Problem\n" + sanitizeForIssue(description) + "\n\n" +
 			fmt.Sprintf("### Environment\n- **Version**: %s\n- **OS**: %s (%s)\n",
 				d.report.Version, d.report.OSSummary(), d.report.Arch)
 	} else {
```
hayke102, line 322: The `else` branch a couple of lines below this (line 324, `body = d.report.FormatMarkdownWithDesc(description)`) is the one report path that never runs `description` through `sanitizeForIssue`. The `kindFeature` branch right above got the `sanitizeHome` → `sanitizeForIssue` rename+redact treatment, but the plain "Bug" path was missed.

If a user pastes a Claude setup-token (`sk-ant-...`) into the bug description while explaining a failed `claude setup-token` add, it goes straight into a public GitHub issue unredacted.

internal/session/session.go
```diff
@@ -178,12 +179,80 @@ func (s *Session) initialRunStatus() Status {
...
+	default:
+		tok := accountToken(s.Account)
+		if tok == "" {
+			// The store no longer knows this account: removed, or the resolver
+			// was never installed. Falls back rather than failing to launch.
+			debuglog.Logger.Warn("session env: account has no token, falling back to ambient login",
+				"id", s.ID, "account", s.Account)
+			break
+		}
+		env = append(env, "CLAUDE_CODE_OAUTH_TOKEN="+tok)
```
hayke102: `sessionEnv()` injects `CLAUDE_CODE_OAUTH_TOKEN` here, but `GuardConflictingAuth()` is only called at the 3 new-session-creation sites (`worktree.go`, `main.go`, `app.go` `handleSessionCreate`). `Restart()` and `RespawnClaude()` both route through this same `sessionEnv()` without ever calling the guard.

A session created cleanly can later have `ANTHROPIC_API_KEY`/`ANTHROPIC_AUTH_TOKEN` show up in the shell env (or just get restarted/resumed from suspend) and silently start billing a different account, with no warning — the exact failure this guard exists to catch.

internal/claudeaccount/client.go
```diff
@@ -0,0 +1,260 @@
...
+
+// GuardConflictingAuth reports the name of an ambient credential that would
+// outrank a per-session token, or "" when the coast is clear.
+//
+// Without this the feature fails silently and expensively: fleet sets
+// CLAUDE_CODE_OAUTH_TOKEN, Claude ignores it in favour of the higher-priority
+// credential, and every session bills to the wrong account with no error
+// anywhere. A wrong-billing failure must be loud.
+func GuardConflictingAuth() string {
+	for _, name := range []string{"ANTHROPIC_AUTH_TOKEN", "ANTHROPIC_API_KEY"} {
```
hayke102: This only checks `ANTHROPIC_AUTH_TOKEN`/`ANTHROPIC_API_KEY`. CLAUDE.md's own auth-precedence table lists `apiKeyHelper` (rank 4) as also outranking `CLAUDE_CODE_OAUTH_TOKEN` (rank 5), but it's never checked here — a repo-wide grep for `apiKeyHelper` turns up no Go code reading it.

A user with `apiKeyHelper` configured in `~/.claude/settings.json` gets a clean pass from this guard, but Claude actually authenticates through the helper — silent misbilling with no warning.

internal/claudeaccount/client.go
```diff
@@ -0,0 +1,260 @@
...
+// The endpoint is the only source: an account fleet cannot poll reads as
+// unknown rather than spent, so least-used degrades to waterfall order instead
+// of wrongly excluding a healthy subscription.
+func (u Usage) Exhausted(now time.Time) bool {
+	// A stale reading is not evidence of exhaustion: an account whose reset has
+	// passed is available again whether or not we have managed to re-poll it.
+	if !u.FiveHourReset.IsZero() && !u.FiveHourReset.After(now) {
+		return false
+	}
+	return u.FiveHourPct >= ExhaustedPct
```
hayke102: `Exhausted` only ever looks at `FiveHourPct`/`FiveHourReset` — `SevenDayPct` is never checked anywhere in selection. An account whose weekly quota is fully spent but whose 5-hour window just reset reports `Exhausted() == false`, so `least_used`/`waterfall` keep routing new sessions to it and the first real API call fails with a weekly rate-limit rejection.

internal/ui/app.go
```diff
@@ -2959,6 +3149,7 @@ func (h *Home) startSessionCmd(msg sessionCreateMsg) tea.Cmd {
 	s := session.NewSession(msg.title, msg.path)
 	s.WorkspaceName = msg.workspaceName
 	s.Agent = msg.agent
+	s.Account = msg.account
```
hayke102, line 3526: `launchLaunchpadSet` (line 3169, unchanged by this diff) calls `startSessionCmd` directly for every batch-launched session, so `msg.account` here is always `""` — it bypasses `resolveAccount`/`GuardConflictingAuth` entirely.

With 2+ accounts and `account_strategy=least_used`, launching several repos at once from the first-run launchpad puts every one of them on the ambient login regardless of configured strategy.

cmd/fleet/worktree.go
```diff
@@ -173,6 +186,34 @@ func runWorktree(args []string) {
...
+	session.SetAccountTokenFunc(accounts.TokenFor)
+	account := ""
+	if !opts.noSession && ag == agent.Claude {
+		if opts.account != "" {
+			if _, ok := accounts.Get(opts.account); !ok {
+				fmt.Fprintf(os.Stderr, "unknown account %q — configure it in fleet first\n", opts.account)
+				os.Exit(1)
+			}
+			account = opts.account
+		} else if acct, ok := claudeaccount.Select(claudeaccount.SelectOpts{
```
hayke102, line 254: This `Select()` call doesn't pass `SelectOpts.Allowed`, unlike `app.go`'s `resolveAccount` — so the per-origin `allowed_accounts` allowlist is silently unenforced for `fleet worktree`/`fleet add`. The explicit `--account` path one branch up (line 201) skips the allowlist check entirely too — an unlisted account can be forced in with no warning.

internal/ui/app.go
```diff
@@ -828,6 +934,16 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 		s.ForkFromID = msg.parentClaudeSessionID
 		s.Agent = ag
+		// Inherit rather than re-pick: the fork resumes the parent's
+		// conversation, so it has to authenticate as the account that
+		// conversation belongs to. A fork whose account was chosen afresh would
+		// resume against a different subscription and lose the prompt cache.
+		// (msg.account is empty only on paths that predate this field, where
+		// falling back to the strategy is still better than nothing.)
+		s.Account = msg.account
+		if s.Account == "" {
```
hayke102: The comment above assumes `msg.account == ""` only happens on paths that predate this field — but `session.go:55` documents `Account == ""` as the live, intentional "ambient /login" state, not just a migration artifact.

Forking a session that's legitimately running on ambient auth hits this fallback and gets reassigned to whatever `resolveAccount` picks — the fork can end up authenticating as a different account than the one that owns the resumed conversation.

internal/ui/accounts.go
```diff
@@ -0,0 +1,931 @@
...
+		// The probe is the only route, deliberately.
+		//
+		// /api/oauth/usage is free but needs the `user:profile` scope that
+		// `claude setup-token` withholds, so for fleet's accounts it answers
+		// 403 — and when rate-limited it answers 429 instead, which is
+		// indistinguishable from a transient failure and left accounts with no
+		// quota at all. It also only tolerates callers claiming to be
+		// claude-code. One mechanism that works for every token type, under
+		// fleet's own name, is worth roughly nine tokens a poll.
+		u, _, err := claudeaccount.ProbeUsage(context.Background(), a.Token)
```
hayke102: This loop calls `ProbeUsage` (blocking HTTP, ~10s client timeout) sequentially per account, inline on `maybePollAccountUsage` — which per the doc comment above runs on the same worker goroutine that drives session status detection, hook processing, and pane refresh for the whole fleet.

If several accounts come due in the same cycle and the endpoint is slow, this can stall fleet-wide status updates for up to N×10s — the same class of bug the `gitWorker` split was built to avoid for git/PR calls.

cmd/fleet/main.go
```diff
@@ -237,6 +238,24 @@ func runAdd(path string) {
 	title := session.TitleFromPath(path)
 	s := session.NewSession(title, path)
 
+	// `fleet add` leaves Agent empty, which resolves to Claude at launch — so
+	// the session is account-eligible and needs one picked, or it silently
+	// ignores a configured multi-account setup.
+	accounts := claudeaccount.Load()
+	session.SetAccountTokenFunc(accounts.TokenFor)
+	cfg := config.Load()
+	if acct, ok := claudeaccount.Select(claudeaccount.SelectOpts{
```
hayke102, line 255: `SelectOpts.Usage` is never populated here (quota is only cached in-memory by a running TUI) — same gap as `worktree.go`. Without it, `pctOf` degrades every account to `unknownPct` and `least_used` silently ties, resolving by configured `Order` instead. Running `fleet worktree foo` from a shell can land the new session on a nearly-exhausted account with no indication the strategy degraded.

internal/ui/accounts.go
```diff
@@ -0,0 +1,931 @@
...
+	// Validation already read this account's quota; keep it rather than paying
+	// for a second probe seconds later.
+	if u := msg.account.InitialUsage(); u.Known() {
+		current := h.accountUsageSnapshot()
+		next := make(map[string]claudeaccount.Usage, len(current)+1)
+		for k, v := range current {
+			next[k] = v
+		}
+		next[msg.account.Email] = u
+		h.accountUsage.Store(&next)
```
hayke102: `h.accountUsage` (an `atomic.Pointer[map]`) is updated with a plain Load-copy-mutate-Store here, and the same pattern repeats in `maybePollAccountUsage`'s `Store` (line ~876) — no `CompareAndSwap` retry, unlike the existing `gitInfoCache`/`writeGitInfo` pattern for the identical cross-goroutine copy-on-write shape.

If a poll cycle's `Store` lands after a concurrent account-add's `Store`, the newer state (with the just-added account's usage) gets silently overwritten until the next poll cycle.

internal/ui/accounts.go
```diff
@@ -0,0 +1,931 @@
...
+			// Anthropic won't identify a setup-token credential, so the row is
+			// keyed by a hash. Say so, or the box reads as a pointless chore.
+			for _, l := range wrapTo(inner, "Added. Anthropic won't say which account a token belongs to, so give it a name you'll recognise:") {
+				b.WriteString(dimStyle.Render(l) + "\n")
+			}
+		} else {
+			b.WriteString(dimStyle.Render(ansi.Truncate("Name for "+d.selectedName(), inner, "…")) + "\n")
+		}
+		b.WriteString("\n" + d.input.View() + "\n")
+		if picks := d.nameSuggestions(); len(picks) > 0 {
```
hayke102, line 330: `nameSuggestions()` calls `claudeaccount.LocalIdentity()`, which does an uncached `os.ReadFile` + `json.Unmarshal` of `~/.claude.json`. This is in `View()`'s rename-state branch, so it re-reads and re-parses that file on every render frame (and again on every keystroke via the matching path in `Update()`) while the rename dialog is open — the blocking-I/O-on-the-render-path pattern this project's architecture otherwise avoids everywhere else.

internal/claudeaccount/probe.go
```diff
@@ -0,0 +1,151 @@
...
+	switch strings.ToLower(h.Get("anthropic-ratelimit-unified-status")) {
+	case "rejected", "exhausted":
+		u.FiveHourPct = 100
+	}
+	return u, h.Get(orgIDHeader), true
+}
+
+// headerPct reads a 0..1 fraction and returns whole percent.
+func headerPct(h http.Header, key string) (int, bool) {
+	raw := h.Get(key)
```
hayke102: The generic `anthropic-ratelimit-unified-status` header (no `5h`/`7d` infix) forces `u.FiveHourPct = 100` here with no equivalent `SevenDayPct` handling — but this header can just as well reflect the 7-day window being the one that's exhausted.

An account whose weekly quota is spent while its 5-hour window still has headroom gets misreported as 5-hour-exhausted, and the header/dialog show a short 5-hour reset countdown instead of the real multi-day wait.

cmd/fleet/worktree.go
```diff
@@ -111,6 +114,16 @@ func parseWorktreeArgs(args []string) (worktreeOpts, error) {
...
+	if o.account != "" {
+		if o.noSession {
+			return o, fmt.Errorf("--account has no effect with --no-session (no session is started)")
+		}
+		// CLAUDE_CODE_OAUTH_TOKEN is a claude.ai credential; the other agents
+		// never read it, so accepting the flag here would silently do nothing.
+		if o.agentName != "" && agent.Type(o.agentName) != agent.Claude {
+			return o, fmt.Errorf("--account only applies to claude sessions (got --agent %s)", o.agentName)
+		}
+	}
```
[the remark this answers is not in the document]
hayke102: Fixed — the check now runs again after `ag` is resolved from `default_agent`, so `--account` is rejected for a Codex/OpenCode session even when `--agent` was never passed.

internal/claudeaccount/client.go
```diff
@@ -0,0 +1,248 @@
...
```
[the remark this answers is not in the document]
hayke102: Fixed — `Exhausted` now treats either window as spent, and `resetOf` returns the reset of the window actually blocking the account (the later one when both are spent, since it isn't usable until the slower returns). Tests added for 7d-only exhaustion, reset-clears-exhaustion, and mixed reset ordering.

internal/ui/accounts_header.go
```diff
@@ -0,0 +1,178 @@
...
+	if len(chips) == 0 {
+		return ""
+	}
+
+	dim := lipgloss.NewStyle().Foreground(ColorTextDim)
+	sep := lipgloss.NewStyle().Foreground(ColorBorder).Render(" │ ")
+	// One label for the whole readout rather than one per chip: both chips show
+	// the same window, so repeating it would be noise that costs real columns.
+	return strings.Join(chips, sep) + dim.Render(" "+win.label())
+}
```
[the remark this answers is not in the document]
hayke102: Not needed any more. The readout was rewritten to render each density widest-to-narrowest and return the first whose `lipgloss.Width` fits the budget, returning "" when none do — it can no longer exceed the width it's handed, at any account count. An omitted-account indicator is a reasonable idea but a feature rather than a defect, so leaving it out for now.

internal/ui/app.go
```diff
@@ -564,9 +573,55 @@ func NewHome(storage *session.StateDB, cfg *config.Config, version string, ident
...
+func (h *Home) allowedAccountsFor(repoPath string) []string {
+	if len(h.cfg.AllowedAccounts) == 0 {
+		return nil
+	}
+	info, ok := h.gitInfo()[repoPath]
+	if !ok || info == nil || info.OriginKey == "" {
+		return nil
+	}
+	return h.cfg.GetAllowedAccounts(OriginExpandKey(info.OriginKey))
+}
```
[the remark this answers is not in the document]
hayke102: Fixed — with an allowlist configured, `allowedAccountsFor` now resolves the origin directly (`git.GetOriginKey`) when the cache hasn't populated it yet, instead of returning nil and being read as unrestricted. Same hole in the CLI fixed alongside.

internal/ui/app.go
```diff
@@ -828,6 +934,16 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+		// Inherit rather than re-pick: the fork resumes the parent's
+		// conversation, so it has to authenticate as the account that
+		// conversation belongs to. A fork whose account was chosen afresh would
+		// resume against a different subscription and lose the prompt cache.
+		// (msg.account is empty only on paths that predate this field, where
+		// falling back to the strategy is still better than nothing.)
+		s.Account = msg.account
+		if s.Account == "" {
+			s.Account = h.resolveAccount(ag, msg.path)
+		}
```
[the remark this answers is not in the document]
hayke102: Fixed — `forkSessionMsg` gained an explicit `accountSet` flag, so "" stops doubling as the unset sentinel. Both fork producers set it, and a fork of an ambient-login session now stays on the ambient login.

internal/ui/accounts_header.go
hayke102: Fixed — good catch, that was a crash. Now truncates with `ansi.Truncate` (display width, like every other truncation in the package) instead of a rune slice. Test added covering CJK, emoji, ASCII and mixed labels.

internal/ui/bugreport.go
hayke102: Fixed, and you're right that it's a real leak — `FormatMarkdownWithDesc`'s sanitizer only rewrites the home directory, so `sk-ant-*` went into public issues verbatim on the Bug path. The description now goes through `sanitizeForIssue` there too, with a guard test covering both kinds.

internal/session/session.go
hayke102: Fixed, and this got worse rather than better with the CLAUDE_CONFIG_DIR rewrite: a config dir's login sits at the *bottom* of the precedence order, so one ambient key now overrides every account at once. The guard moved into `sessionEnv` itself, which every launch path funnels through — create, restart, respawn and the idle-suspend wake.

internal/claudeaccount/client.go
hayke102: Fixed — `GuardConflictingAuth` now reads `apiKeyHelper` from `~/.claude/settings.json` and names it in the warning. It's the worst of the three to miss: it isn't an env var, so it survives every scrub and every fresh shell.

internal/claudeaccount/client.go
hayke102: Fixed — same defect @coderabbitai raised independently. `Exhausted` now treats either window as spent, and reset selection follows whichever window is actually blocking.

internal/ui/app.go
hayke102: Fixed — `launchLaunchpadSet` now resolves an account per item. It was the worst place to miss: first run, several sessions at once, all silently on the ambient login.

cmd/fleet/worktree.go
hayke102: Fixed both halves — the CLI now passes `Allowed` (resolving the origin key itself), and an explicit `--account` is validated against the allowlist rather than bypassing it. An allowlist that held in the TUI and not the CLI was worse than none.

internal/ui/app.go
hayke102: Fixed — see #948. `accountSet` now distinguishes "parent runs on ambient" from "no account carried", and the misleading comment is gone.

internal/ui/accounts.go
hayke102: Fixed — the poll moved to its own `accountWorker` goroutine, mirroring `gitWorker` for exactly the reason you name. Worth noting it had got slightly worse: post-rewrite it also shells out to `security` per account before the HTTP call.

cmd/fleet/main.go
hayke102: Addressed by logging rather than fetching. Populating `Usage` from the CLI means an HTTP call per account on every `fleet worktree`, which is a real latency cost on a command people script. It now logs plainly that no quota was available and configured order decided, so the degradation is visible instead of silent.

internal/ui/accounts.go
hayke102: Fixed — both stores now go through a `writeAccountUsage` helper that CompareAndSwaps in a retry loop, matching `writeGitInfo`. The poll also builds a delta and merges it, rather than storing a wholesale replacement map that could clobber a concurrent add.

internal/ui/accounts.go
hayke102: No longer applies — `LocalIdentity` was deleted in the CLAUDE_CONFIG_DIR rewrite. Identity now comes from `claude auth status` at add time, and `nameSuggestions` returns two static strings with no filesystem access.

internal/claudeaccount/probe.go
hayke102: No longer applies — `probe.go` was deleted. Quota comes from `/api/oauth/usage`, which reports each window separately with no shared status header. The underlying concern was real though, and is now fixed properly: `Exhausted` treats either window as spent.

internal/claudeaccount/account.go
```diff
@@ -0,0 +1,449 @@
...
+
+// Upsert adds an account or updates the one it refers to, preserving that
+// account's key, Order and Label so re-adding a token doesn't silently reorder
+// the rotation, discard a rename, or mint a second identity for a subscription
+// fleet already knows.
+//
+// QuotaUnavailable is deliberately NOT carried over: it is a verdict about a
+// specific token, and a replacement deserves its own. Inheriting it would
+// silently disable quota forever if a future token did carry the scope, where
+// clearing it costs one HTTP call that re-marks it immediately.
```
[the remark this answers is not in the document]
hayke102: No longer applies — both the field and the comment were removed with the probe; quota no longer has a scope-forbidden state to persist.

internal/claudeaccount/account.go
```diff
@@ -0,0 +1,449 @@
...
+	defer s.mu.Unlock()
+	for i := range s.accounts {
+		if s.accounts[i].Email != email {
+			continue
+		}
+		if s.accounts[i].OrgUUID != "" {
+			return false
+		}
+		s.accounts[i].OrgUUID = org
+		return true
```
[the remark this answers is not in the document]
hayke102: No longer applies — `SetOrgUUID` and the org backfill are gone. An account is now created only by a completed login, which reports its real email and organization together, so the blank-org window this depends on can't occur. `Upsert` still matches on organization first.

internal/ui/account_move_test.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	h.sessions = append(h.sessions, beta)
+	h.flatItems = append(h.flatItems, sessionRow(beta))
+
+	h.openAccountPicker() // opened on alpha
+	h.cursor = 1          // an async rebuild lands the cursor on beta
+
+	if _, cmd := h.Update(accountPickedMsg{email: "second@x.com"}); cmd != nil {
+		cmd()
+	}
```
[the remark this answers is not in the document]
hayke102: Fixed — dropped the `cmd()` call. You're right that it was leaving a real tmux session behind; the move and its persistence both happen synchronously before the command is returned, so the assertions cover it.

changelog/unreleased/account-reset-times.md
```diff
@@ -0,0 +1,5 @@
+---
+type: improved
+---
+
+**Quota now comes with a horizon.** The account readout shows when each window resets — `weekly │ yuval ▰▰▱▱ 40%·2h │ hayke102 ▰▰▱▱ 54%·5d` — because "54% of a week" tells you nothing until you know whether it refills tomorrow or in five days. The strip renders the most detail your terminal width allows and tightens as it shrinks, rather than switching between two fixed layouts.
```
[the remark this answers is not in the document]
hayke102: Fixed — retitled to "Quota with a deadline" (3 words) and cut the closing clause about the layout algorithm.

internal/claudeaccount/account.go
```diff
@@ -0,0 +1,408 @@
...
+	kept := s.accounts[:0]
+	for _, a := range s.accounts {
+		if a.ConfigDir == "" {
+			debuglog.Logger.Error("account has no config dir and cannot be used; log it in again with Ctrl+K → Manage Claude Accounts",
+				"email", a.Email, "label", a.Label)
+			continue
+		}
+		kept = append(kept, a)
+	}
+	s.accounts = kept
```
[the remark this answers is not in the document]
hayke102: Fixed — `Load` now rewrites `accounts.json` immediately when it drops legacy records, and logs loudly if that save fails. You're right that filtering in memory left the migration looking done while a year-long credential stayed on disk.

internal/claudeaccount/identity.go
```diff
@@ -0,0 +1,212 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return Identity{}, ctx.Err()
+		case <-t.C:
+			id, err := Identify(ctx, dir)
+			if err == nil {
+				debuglog.Logger.Info("account login completed",
+					"email", id.Email, "plan", id.Plan, "org", id.OrgUUID)
+				return id, nil
```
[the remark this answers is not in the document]
hayke102: Keeping this one. The email is the account key rather than incidental PII, and "which account did this session launch as" is the first question every multi-account report has to answer — the line exists for that. Credentials are redacted by `sanitizeForIssue` at the chokepoint; the account name is the diagnostic.

internal/session/account_test.go
```diff
@@ -0,0 +1,304 @@
...
+	out := buf.String()
+	if strings.Contains(out, "sk-ant-") {
+		t.Fatalf("a credential reached the debug log:\n%s", out)
+	}
+	if !strings.Contains(out, "work@x.com") {
+		t.Errorf("launch log should name the account:\n%s", out)
+	}
+	if !strings.Contains(out, dirs["work@x.com"]) {
+		t.Errorf("launch log should name the config dir:\n%s", out)
+	}
```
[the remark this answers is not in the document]
hayke102: Keeping the log field. `sanitizeForIssue` already rewrites the home directory to `~` before anything reaches an issue, so what publishes is `~/.config/fleet/accounts/8c11c968` — no username. The path is the most useful single field for diagnosing which account a session launched as, so the assertion is deliberate.

internal/ui/accounts_dialog_test.go
```diff
@@ -0,0 +1,185 @@
...
+			d.Show([]claudeaccount.Account{long, fp}, nil, long.Email, true)
+		}},
+		{"waiting for login", func(d *AccountsDialog) {
+			d.Refresh([]claudeaccount.Account{long}, nil, "")
+			d.SetBusy("Waiting for the login…")
+		}},
+		{"rename", func(d *AccountsDialog) {
+			d.Refresh([]claudeaccount.Account{fp}, nil, "")
+			d.Added(fp.Email)
+		}},
```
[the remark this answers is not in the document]
hayke102: Fixed — that was my regression when `Added` stopped opening the rename box. The case now enters `accountsRename` explicitly, so the prompt line, input and suggestion row are wrap-tested again at every width.

internal/ui/accounts.go
```diff
@@ -0,0 +1,890 @@
...
+			var row []string
+			for i, p := range picks {
+				row = append(row, fmt.Sprintf("%d %s", i+1, p))
+			}
+			b.WriteString(dimStyle.Render(ansi.Truncate(strings.Join(row, "   "), inner, "…")) + "\n")
+		}
+
+	case accountsWaitingLogin:
+		b.WriteString(d.notice + "\n")
+		b.WriteString(dimStyle.Render("Checking the token with Anthropic…") + "\n")
```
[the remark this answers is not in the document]
hayke102: Fixed — reworded to "Waiting for the browser login to finish…", and the dialog test's assertion updated with it.

internal/ui/accounts.go
```diff
@@ -0,0 +1,890 @@
...
+		return func() tea.Msg {
+			return accountLoggedInMsg{err: fmt.Errorf("could not start `claude`: %w", err)}
+		}
+	}
+
+	// Stays on screen for the whole flow, unlike anything printed into the pane.
+	// On the pane border, where it survives whatever the login flow paints.
+	ts.ApplyStatusBar(tmux.StatusBarOpts{
+		DisplayName: "Signing in to another Claude account · Ctrl+Q returns to fleet",
+	})
```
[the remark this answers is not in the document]
hayke102: Valid, and deferring deliberately. Unlike the quota poll (which I did move to its own goroutine in this batch), this runs once per account-add on a keypress the user is already waiting on, so the stall is bounded and visible rather than recurring. Tracking as a follow-up rather than growing this PR further.

internal/ui/app.go
```diff
@@ -1516,6 +1676,11 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			// branch/dirty/PR would stay blank for the whole process lifetime.
 			// No bootstrap probe runs here, so the repoLastHotAt reasoning holds.
 			go h.gitWorker()
+			// Its own goroutine for the same reason gitWorker has one: the poll
+			// shells out to `security` and then makes an HTTP call, per account,
+			// sequentially. On the status worker several accounts falling due in
+			// one cycle could stall status detection for the whole fleet.
+			go h.accountWorker()
```
hayke102: **This is the only `go h.accountWorker()` in the tree — and it's on the empty-fleet path only.**

`discoveryMsg` has a single emitter, `scanDiscovery()`, dispatched only when `len(h.bootstrapRepoSet()) == 0`. Every other launch resolves through `bootstrapGitInfo` → `bootstrapDoneMsg` (~:1644), which starts `statusWorker` and `gitWorker` but not this one.

So for anyone with at least one pinned repo or session — exactly the population that configures a second account — the quota poll never runs for the process lifetime. `accountUsage` stays empty, so `Usage.Known()` is false everywhere:

- the header readout renders `""` forever (`renderAccountUsageHeader` skips every account)
- `least_used` scores every account at `unknownPct` and silently degenerates to configured order
- `ErrNotLoggedIn` is never observed, so `dropLoggedOut` never fires
- `healAccountBeforeRelaunch` can never see a spent or logged-out pin, so `r` never heals

The whole quota half of the feature is dead on the normal boot path. Needs `go h.accountWorker()` in the `bootstrapDoneMsg` branch too (or hoisted somewhere both paths reach).

internal/session/session.go
```diff
@@ -189,12 +191,109 @@ func (s *Session) initialRunStatus() Status {
...
+// SetAccountConfigDirFunc installs the account resolver. Passing nil disables
+// per-account auth, which is the state fleet runs in when no accounts are
+// configured.
+func SetAccountConfigDirFunc(fn func(string) string) {
+	accountDirMu.Lock()
+	defer accountDirMu.Unlock()
+	accountDirFn = fn
+}
+
+func accountConfigDir(email string) string {
```
hayke102, line 224: **Silent nil default + a CLI entry point that never installs the resolver = sessions billed to the wrong subscription.**

`accountConfigDir` returns `""` when `accountDirFn == nil`, and `sessionEnv`'s default branch then logs a warning and emits no `CLAUDE_CONFIG_DIR` — the session runs on whatever login is ambient.

`SetAccountConfigDirFunc` is installed in exactly four places: `ui/app.go`, `ui/accounts.go`, `cmd/fleet/main.go`, `cmd/fleet/worktree.go`. **`cmd/fleet/send.go` is not one of them** — yet its `deliver()` calls `s.Restart()` to wake a suspended session, which goes straight through here.

So `fleet send my-session "carry on"` on a session pinned to the work subscription relaunches it under the personal login, then persists the new tmux name and status so the row looks healthy. The sidebar keeps naming the pinned account in the preview footer while the live process bills the other one — exactly the lie `moveSelectedToAccount` and `healAccountBeforeRelaunch` document as unacceptable, discovered only from an invoice.

Worth considering whether a nil resolver should be a hard failure for a session that names an account, rather than a warn-and-continue — the package-global-with-silent-default is what makes this easy to miss in the next entry point too.

internal/ui/app.go
```diff
@@ -2875,9 +3071,209 @@ func (h *Home) handleSessionCreate(msg sessionCreateMsg) (tea.Model, tea.Cmd) {
...
+// account that is merely busy, or onto a specific one, which is a judgement call
+// fleet has no business making on its own.
+func (h *Home) openAccountPicker() {
+	s := h.selectedSession()
+	if s == nil {
+		return
+	}
+	usage := h.accountUsageSnapshot()
+	rows := make([]accountPickerRow, 0, h.accounts.Len())
+	for _, a := range h.accounts.List() {
```
hayke102, line 3353: **The picker bypasses `allowed_accounts`.** It builds rows straight off `h.accounts.List()`; every other assignment path goes through `claudeaccount.Select` with `Allowed: h.allowedAccountsFor(path)`.

The `worktree.go` change in this same branch adds that filter to the CLI with the rationale that a policy holding in one surface and not another is worse than not having one — including for an explicit `--account`. This is the surface that missed it: rows are disabled only for `current` and `logged out`.

A user who restricted `origin:github.com/acme/*` to their work subscription opens `.` → "Move to Account…" on an acme session, picks their personal account, and fleet persists the pin and relaunches — billing client work to the personal subscription with no warning.

`cmd/fleet/main.go` `runAdd` has the same gap (constructs `SelectOpts` with three fields, no `Allowed`).

internal/ui/statusreport.go
```diff
@@ -344,12 +345,20 @@ func (d *BugReportDialog) viewStatusForm() string {
...
+//
+// Two things are removed:
+//
+//   - The user's home directory, rewritten to "~".
+//   - Anthropic credentials. The add-account flow puts a year-long
+//     account token on a real tmux pane, and pane excerpts are
+//     published here — so the redaction lives at the chokepoint rather than at
+//     each call site, where a future caller could forget it.
+func sanitizeForIssue(s string) string {
+	s = claudeaccount.Redact(s)
```
hayke102: **`Redact` is the strict pattern; pane excerpts need `RedactCaptured`, which now has zero production callers.**

`buildStatusReportBody` pushes the 40-line pane excerpt through `sanitize` → this function → `claudeaccount.Redact`, whose regex is `sk-ant-[a-z0-9]{3,8}-[A-Za-z0-9_-]{16,}`.

`account.go` states the rule explicitly: *"Call this [`RedactCaptured`] — never `Redact` — when the text is raw pane content, where a truncated or mid-print token would slip under `Redact`'s length floor."* Grep says `RedactCaptured` has no non-test caller left.

Pane lines are wrapped and clipped, so a credential split across the pane boundary leaves fragments shorter than the 16-char floor — `Redact` matches nothing and the fragments are filed verbatim in a public issue. `TestBuildStatusReportBody_NeverLeaksAccountToken` feeds only a full-length token, so it passes while the exact case the looser pattern exists for goes uncovered.

internal/ui/accounts.go
```diff
@@ -0,0 +1,898 @@
...
+		// quota at all. It also only tolerates callers claiming to be
+		// claude-code. One mechanism that works for every token type, under
+		// fleet's own name, is worth roughly nine tokens a poll.
+		u, err := claudeaccount.FetchUsage(context.Background(), a.ConfigDir)
+		if err != nil {
+			// A missing login is the one failure that carries information.
+			// Everything else here means "fleet could not ask", which says
+			// nothing about the account; this says it cannot run a session at
+			// all, and selection must act on it rather than score it as unknown.
+			loggedOut := errors.Is(err, claudeaccount.ErrNotLoggedIn)
```
hayke102, line 832: **A genuinely logged-out account fails as `ErrNoCredential`, not `ErrNotLoggedIn`, so `LoggedOut` never gets set.**

`/logout` (or `claude auth logout`) inside a session deletes that config dir's Keychain item. The next poll's `accessToken()` fails `security find-generic-password` and returns `ErrNoCredential` (`usage.go:76/85/88`) — `ErrNotLoggedIn` only comes from a 401/403 on the HTTP call, which never happens because there's no bearer to send.

So `prev.LoggedOut` stays false, `Usable()` stays true, `dropLoggedOut` keeps it as a candidate, and `pctOf` scores it `unknownPct` (50) because `Known()` is false. 50 beats a healthy account at 59% — every subsequent new session is assigned to the account that cannot run, launches, and sits at a login prompt, with no error surfaced and the header still showing it as fine.

This is the precise scenario `dropLoggedOut` was written for. Either treat `ErrNoCredential` as evidence of logged-out here, or have `accessToken` distinguish "item missing" (evidence) from "keychain unreadable" (not evidence).

internal/claudeaccount/identity.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	// apiKeyHelper is not an environment variable, so it survives every scrub and
+	// every fresh shell — and it outranks a config dir's login just the same. A
+	// user who configured one once, months ago, gets a clean pass from the env
+	// checks above while Claude authenticates through the helper.
+	if helperConfigured() {
+		debuglog.Logger.Warn("apiKeyHelper is configured and outranks every account login",
+			"file", "~/.claude/settings.json",
+			"effect", "sessions would authenticate through the helper, not the chosen account")
+		return "apiKeyHelper (in ~/.claude/settings.json)"
```
hayke102: **`apiKeyHelper` was added to a return channel whose callers all treat non-empty as fatal — this now hard-blocks all Claude session creation.**

The doc comment moved from *"Refusing loudly is the only honest answer"* to *"Saying so is the only honest answer"*, but no caller changed: `handleSessionCreate` does `setError` + `return h, nil`; `runAdd` and `runWorktree` both `os.Exit(1)`.

A user with `apiKeyHelper` in `~/.claude/settings.json` — an enterprise/Bedrock/key-rotation setup configured months ago — plus at least one fleet account can no longer create *any* Claude session. `a`, `n`, `w` all flash the error; `fleet add` and `fleet worktree` exit 1.

And the remedy is unfollowable: *"unset it"* is not something you can do to a JSON key from a shell. The only escape is removing every configured account.

If the intent was a warning rather than a refusal, this needs a separate return channel (or a severity) so callers can distinguish `apiKeyHelper` from the env-var cases.

internal/ui/app.go
```diff
@@ -564,9 +576,84 @@ func NewHome(storage *session.StateDB, cfg *config.Config, version string, ident
...
+	// launch it has no OriginKey for a repo. Returning nil there would read as
+	// "no restriction" — Select treats an empty allowlist as unrestricted — and
+	// a session created in that window could land on a disallowed account. With
+	// an allowlist configured, resolve it directly rather than guess.
+	origin := ""
+	if info, ok := h.gitInfo()[repoPath]; ok && info != nil {
+		origin = info.OriginKey
+	}
+	if origin == "" {
+		origin = git.GetOriginKey(repoPath)
```
hayke102: **This shells out to git on the Update goroutine — three lines below the doc comment forbidding exactly that.**

`allowedAccountsFor`'s own comment says it resolves from cached git info *"rather than shelling out, since this runs on the Update goroutine."* `git.GetOriginKey` runs `git -C <repo> config --get remote.origin.url`.

Reached from `resolveAccount` → `handleSessionCreate`, `launchLaunchpadSet`, `healAccountBeforeRelaunch` — all inside `Update()`. In the exact window this fallback targets (first seconds after launch, before the bootstrap fan-out resolves), pressing `a` on a repo on a slow or networked FS blocks the whole TUI, and `launchLaunchpadSet` multiplies it by one call per selected repo.

Separately: `GetOriginKey` **never returns `""`** — it falls back to `"local:"+filepath.Base(repoPath)`. So the `if origin == ""` guard below is unreachable, and a repo with no remote is now keyed `origin:local:<dir>` instead of being treated as unrestricted.

hayke102 (commented):
Findings from a multi-agent review of this branch (74 candidates → 15 verified). Most cluster around three roots: the quota poll never starting on the normal boot path, the widened `Usage.Exhausted` contract not reaching its consumers, and residue from the abandoned `ANTHROPIC_AUTH_TOKEN` design.

internal/ui/accounts.go
```diff
@@ -0,0 +1,898 @@
...
+
+// renderQuotaCell renders the 5-hour window as "42% · resets 15:04", or the
+// reset time alone once the window is spent — at that point the percentage
+// tells you nothing you can act on and the clock is the whole answer.
+func renderQuotaCell(u claudeaccount.Usage) string {
+	dimStyle := lipgloss.NewStyle().Foreground(ColorTextDim)
+	if u.Exhausted(time.Now()) {
+		s := "spent"
+		if !u.FiveHourReset.IsZero() {
+			s += " · back " + u.FiveHourReset.Local().Format("15:04")
```
hayke102: **Branches on the widened `Exhausted`, formats the 5-hour clock.**

`Usage.Exhausted` is now `windowSpent(FiveHour…) || windowSpent(SevenDay…)`, but this cell always prints `FiveHourReset`. An account at 99% weekly (refills in five days) whose 5-hour bucket just reset renders `spent · back 16:20`.

That string is reused by `AccountPickerDialog.renderRow`, where only `current` / `logged out` rows are disabled — a spent row is fully pickable. So: session stalls, user opens "Move to Account…", sees `spent · back 16:20`, decides 20 minutes is acceptable, picks it. `moveSelectedToAccount` persists the pin and calls `Restart()` — a real relaunch that discards the prompt cache — onto an account blocked for five days.

Same mismatch in `accounts_header.go:177` (hardcoded `windowFiveHour`) and `app.go:3146` (`"its 5-hour window is spent"`). Widening `Exhausted` needs a companion "which window" accessor so all three can name the right one.

internal/ui/bugreport.go
```diff
@@ -316,12 +316,18 @@ func (d *BugReportDialog) openGitHubIssue(description string) tea.Cmd {
...
 		// Inject user description and render stats into the report.
-		body = d.report.FormatMarkdownWithDesc(description)
+		//
+		// sanitizeForIssue, not the raw description: the report formatter only
+		// rewrites the home directory, so a credential pasted into a bug
+		// description reached a public issue verbatim. The feature branch above
+		// already sanitizes; this path was missed. Someone describing a failed
+		// account add is exactly the person likely to paste one.
+		body = d.report.FormatMarkdownWithDesc(sanitizeForIssue(description))
```
hayke102: **The description is sanitized here, but the two blocks appended after it are not — and this branch just filled both with account emails and org UUIDs.**

`FormatMarkdownWithDesc` → `FormatEnvironmentMarkdown(true)`, whose `sanitize` closure is `strings.ReplaceAll(s, home, "~")` and nothing else — it never reaches `sanitizeForIssue`.

So the **Config** block publishes `"default_account": "yuval@company.com"` and `"allowed_accounts": {"origin:github.com/acme/x": ["work@acme.com"]}` verbatim, and the **Debug Log** block publishes the lines this branch added on every launch and poll: `account loaded email=… plan=max org=<uuid>`, `session env: launching as claude account account=…`, `account usage polled account=…`.

`sanitizeForIssue` was extended for `sk-ant-*`, but emails and org UUIDs are new PII the chokepoint doesn't cover — and the reporter is shown neither block before it's filed.

internal/ui/app.go
```diff
@@ -2875,9 +3071,209 @@ func (h *Home) handleSessionCreate(msg sessionCreateMsg) (tea.Model, tea.Cmd) {
...
+	reason := h.accountUnusableReason(s.Account)
+	if reason == "" {
+		return false
+	}
+
+	// Select already refuses to hand back a rejected account, but it will return
+	// a spent one when every candidate is spent — so this can land on the account
+	// it started from. Moving to itself would cost the cache for nothing.
+	next := h.resolveAccount(agent.Claude, s.ProjectPath)
+	if next == s.Account {
```
hayke102: **The `next == s.Account` guard doesn't cover the case the comment above it describes.**

When every account is spent, `Select` returns `soonestReset` — which is often a *different* spent account. Three accounts all at 100%, session pinned to A (resets in 3h), C resets in 30m: `next` = C ≠ A, so the session is persisted onto C, forced down the full `Restart` path (no `RespawnClaude`), and toasted "Moved to C" — while C can't run it either. Cache discarded for nothing.

`TestRestartStaysPutWhenEveryAccountIsSpent` only passes by accident: both fixtures come from `knownUsage(100)`, whose `FiveHourReset` is `time.Now().Add(time.Hour)` evaluated twice, so "second" is later than "first" by the nanoseconds between the two calls and `soonestReset` happens to return the already-pinned account. Any fixture ordering where the other reset is genuinely earlier fails the test and exposes the move.

The guard probably wants to be "stay put unless `next` is actually *usable*", not "unless `next` is me".

internal/claudeaccount/configdir.go
```diff
@@ -0,0 +1,280 @@
...
+		// Replace only a symlink we own. A real directory here means something
+		// wrote into the account dir directly, and deleting it would destroy
+		// data we didn't create.
+		if fi, err := os.Lstat(dst); err == nil {
+			if fi.Mode()&os.ModeSymlink == 0 {
+				debuglog.Logger.Warn("account dir holds a real directory where a shared link belongs; leaving it",
+					"dir", dir, "entry", name)
+				continue
+			}
+			_ = os.Remove(dst)
```
hayke102, line 122: **Remove-then-create with no locking, and this `return err` skips `mirrorSettings`/`mirrorClaudeJSON`.**

Two sessions on the same account restarting concurrently (`reloadAll` fans out per session) both run `Provision`. A removes `<dir>/projects` and recreates it; B — which `Lstat`'d before A's remove — removes A's fresh symlink and calls `os.Symlink`; one of them gets `EEXIST`.

Two consequences:
1. `Provision` returns here, so `settings.json` (carrying fleet's hooks) and `.claude.json` are never refreshed for that launch — and `sessionEnv` logs the error and launches anyway.
2. A `claude --resume` starting inside the window where `projects` doesn't exist finds no transcript and opens an empty conversation.

A package mutex around `Provision` (like the one `GitWorktreeProvider` uses for concurrent removes) would close both.

internal/ui/accounts.go
```diff
@@ -0,0 +1,898 @@
...
+	h.attachStartedAt.Store(time.Now().UnixNano())
+	h.actionLog.Add("add claude account", "", true)
+
+	// Watch for the login completing and pull the user back the moment it does,
+	// so finishing the browser flow is the last thing they have to do. Runs
+	// alongside tea.Exec, which blocks until the attached client goes away;
+	// detaching from here is what makes it return.
+	identified := make(chan claudeaccount.Identity, 1)
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), accountLoginWindow)
```
hayke102: **No cancellation path: the watcher outlives the pane by up to 10 minutes and can resurrect the dir the cleanup just deleted.**

The context derives from `Background()` and `cancel` never escapes the closure, so nothing in the `tea.Exec` completion callback can stop it.

User presses Ctrl+Q to abandon the login → the callback runs `Identify`, errors, calls `RemoveConfigDir(dir)`. The watcher is still ticking: it keeps invoking `claude auth status` with `CLAUDE_CONFIG_DIR=<the deleted dir>` every ~1.5s for the remaining window — roughly 400 node process spawns — and **Claude Code creates its config dir when missing**, so the directory the cleanup just removed reappears. That leaves exactly the orphaned, login-less account dir `Provision` will keep refreshing on every launch.

Pressing `a` again to retry stacks a second watcher on top of the first.

Wants the `cancel` stored on `Home` (or a per-attempt generation) so the completion callback can stop it.

internal/ui/app.go
```diff
@@ -813,6 +905,52 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	case sessionCreateMsg:
 		return h.handleSessionCreate(msg)
 
+	case accountLoginMsg:
+		return h, h.runAccountLogin()
+
+	case accountLoggedInMsg:
+		return h.handleAccountLoggedIn(msg)
+
+	case accountRemoveMsg:
```
hayke102, line 959: **Removing an account leaks its config dir and its Keychain login, permanently.**

This handler calls `Store.Remove` + `persistAccounts` and stops. `~/.config/fleet/accounts/<8hex>/` stays on disk (holding the mirrored `.claude.json` with the user's MCP server config) and the Keychain item `Claude Code-credentials-<hash>` stays logged in to that subscription.

`RemoveConfigDir` exists and has exactly one production caller — the abandoned-login cleanup in `runAccountLogin` — so **a deliberate remove leaks what an abandoned add cleans up.**

And it compounds: re-adding the same subscription mints a fresh random dir (`Upsert` matches on `OrgUUID` and overwrites `ConfigDir`), orphaning another one each time. A user cycling logins accumulates dirs and stale Keychain items with no way to clear them from fleet.

If leaving the login is deliberate (so a remove is reversible), the confirm dialog should say so — right now nothing tells the user a credential survives.

internal/ui/app.go
```diff
@@ -1730,9 +1895,22 @@ func (h *Home) View() tea.View {
...
+	rightEdge := h.width - 1
 	if h.hasUnseenWhatsNew && !h.modalOpen() {
 		badge := renderWhatsNewBadge(h.whatsNewFrame)
-		base = overlayAt(badge, base, h.width-lipgloss.Width(badge)-1, 0)
+		badgeW := lipgloss.Width(badge)
+		base = overlayAt(badge, base, rightEdge-badgeW, 0)
+		rightEdge -= badgeW + 2
+	}
+	if ShowAccountUsage && !h.modalOpen() {
+		readout := renderAccountUsageHeader(h.accounts.List(), h.accountUsageSnapshot(), rightEdge, time.Now())
```
hayke102: **`rightEdge` is an x-coordinate, not a width budget — so the strip is fitted against the whole screen and drawn over the breadcrumb.**

On a 100-col terminal with two accounts and no What's New badge, `rightEdge` is 99, so `renderAccountUsageHeader` accepts the widest density that fits *99 columns* (~74 for two labelled chips with gauges and countdowns), and this overlays it at x=25.

`renderHeader` has already drawn `❯_ fleet › brizzai/fleet › feat-multi-account › <title up to 40 chars>`, which extends well past column 25 — so its tail is overprinted mid-word with no separator.

The What's New badge uses the same pattern and gets away with it because it's ~14 columns; this strip can be 70+. Nothing measures the breadcrumb's width or truncates it to make room, and there's no clipping on either side.

internal/ui/accounts.go
```diff
@@ -0,0 +1,898 @@
...
+		return lipgloss.NewStyle().Foreground(ColorRed)
+	case pct >= 60:
+		return lipgloss.NewStyle().Foreground(ColorYellow)
+	default:
+		return lipgloss.NewStyle().Foreground(ColorGreen)
+	}
+}
+
+// accountSetupPrefix names the throwaway tmux session that runs
+// `claude setup-token`.
```
hayke102: Doc rot from the mid-branch pivot — this says the pane runs `claude setup-token`, but `accountLoginCmd` twelve lines below is `claude auth login --claudeai`. This is the single most misleading leftover: a maintainer grepping for where `setup-token` is invoked lands here and concludes the token flow is still live.

Two more in the same file:
- **:487** — *"a `claude setup-token` credential simply isn't entitled to the usage endpoint"*, describing a mechanism this branch removed.
- **:865** — `accountLabel`'s comment reasons about `TokenFor`, which no longer exists anywhere in the tree.

(The `setup-token` mentions in `claudeaccount/*.go` read fine — those are the documented *why-not*, not claims about current behavior.)

internal/ui/app.go
hayke102: Fixed — this was the biggest one in the round. Extracted `startAccountWorker`, guarded by a `sync.Once` on Home, and called it from the `bootstrapDoneMsg` branch as well as `discoveryMsg`. The Once rather than a bool because the two branches aren't provably exclusive and two pollers would double the `security`/HTTP traffic and race on `writeAccountUsage`.

Added `TestBothRevealPathsStartTheAccountWorker`, which parses app.go and fails if either reveal branch stops calling it — asserted structurally because there's nothing to observe at runtime: a goroutine that was never started looks exactly like a fleet whose accounts happen to be unpolled.

internal/session/session.go
hayke102: Fixed — `runSend` now installs the resolver alongside `debuglog.Init()`.

On your second point, I went with a louder log rather than a hard failure. The two causes of an empty dir are now separated: no resolver at all is a wiring bug in whichever binary is running (this session names an account, so accounts exist, so somebody forgot the call) and logs at **Error** naming the fix; a genuinely removed account stays a Warn. A hard failure would refuse to deliver a message for a user with no accounts configured at all, which is worse than the thing it prevents.

internal/ui/app.go
hayke102: Fixed on both surfaces. `openAccountPicker` now resolves `allowedAccountsFor(s.ProjectPath)` and disables disallowed rows with the note `not allowed here`; `runAdd` passes `Allowed` (shelling out to git is fine there — one-shot CLI, not the Update goroutine).

Dimmed rather than hidden: a row that vanishes reads as a missing account, where a disabled one carrying its reason teaches the policy. `TestPickerRefusesAnAccountTheOriginDisallows` covers both halves.

internal/ui/statusreport.go
hayke102: Fixed — `sanitizeForIssue` now calls `RedactCaptured`.

Widened wholesale rather than adding a second path: `sk-ant-\S*` only matches strings that already begin with the literal prefix, so there's no over-matching risk on ordinary prose, and one chokepoint is the entire reason the function exists. `TestTruncatedCredentialIsRedactedToo` covers the fragment cases (clipped mid-token, clipped right after the prefix, clipped by the renderer's ellipsis) plus a clean-prose control.

internal/ui/accounts.go
hayke102: Fixed, taking your second option — `accessToken` now distinguishes evidence from ignorance.

`security` exits 44 (`errSecItemNotFound`) when the item is absent, so that maps to `ErrNotLoggedIn`; an item that exists but holds no `accessToken` counts too, since no session can start from it. An unreadable or unparseable item stays `ErrNoCredential` — a locked Keychain says nothing about the login, and blanket-mapping would report the whole fleet logged out the moment `security` had a bad day. Four cases pinned in `usage_test.go` against a stub `security` on PATH.

internal/claudeaccount/identity.go
hayke102: Fixed — you're right that it was a regression, and the remedy text was the tell.

`GuardConflictingAuth` now returns an `AuthConflict{Name, Fatal, remedy}`. The env vars stay fatal (one `unset` fixes them, so refusing costs nothing), `apiKeyHelper` is not: there's no command that clears it for a single launch, so refusing left the user with no way to run any Claude session until they edited a file fleet never named. It now says its piece and the session launches.

`Message(account)` builds the whole sentence including the right remedy, which also removes the three call sites that each hardcoded "unset it".

internal/ui/app.go
hayke102: Both fixed.

The shell-out is memoized per repo path (`originKeyFor`, mutex-guarded map on Home) — at most one exec per repo per process instead of one per keypress. Kept rather than dropped, since removing it reopens the permissive window the comment exists to close, and with an allowlist configured that window misbills.

Dead guard deleted, and the comment now says what the code does. On your second point: `local:<dir>` keying is correct rather than a bug — the sidebar groups a remoteless repo the same way, so `allowed_accounts` keyed `origin:local:<dir>` works. Only the unreachable guard was misleading.

internal/ui/accounts.go
hayke102: Fixed — added `Usage.SpentWindow(now) (Window, time.Time, bool)` as the companion accessor you described, and corrected all three call sites.

The `Window` type and its `Name()` live in `claudeaccount` rather than the UI, because these three surfaces have to agree and two copies of "5-hour"/"weekly" is how they stop agreeing — `accounts_header.go` now aliases it instead of defining its own.

Left the picker row selectable: with the right window named, choosing a spent account is a legitimate informed call (it may be the right one to queue work on). The bug was that it named the wrong one. Covered by `TestSpentWeeklyWindowIsTimedByTheWeeklyClock` and by `SpentWindow` tests for both-spent and neither-spent.

internal/ui/bugreport.go
hayke102: Fixed in three parts.

`FormatEnvironmentMarkdown` and `FormatMarkdownWithDesc` now take a `diagnostics.Scrubber`, and the UI passes `sanitizeForIssue` — so the Config and Debug Log blocks reach the same chokepoint every other published string does. A required parameter rather than a package default precisely because forgetting it is how this went wrong; `nil` (home-rewrite only) is now visible at the call site.

`sanitizeForIssue` gained an email pattern, redacting local part **and** domain — on a work subscription the domain is the client's name.

Org UUIDs are handled at the source instead: `account loaded` logs an 8-character prefix, which still answers the only question asked of it (are these two accounts the same subscription). Redacting every UUID here would also mask the Claude/Codex session ids that wrong-status reports exist to compare. `TestAccountEmailsNeverReachAnIssue` covers both blocks.

internal/ui/app.go
hayke102: Fixed — the guard is now "stay put unless \`next\` is actually usable", via `accountUnusableReason(next)`. An empty `next` is excluded explicitly: that's the ambient login, which is the documented fallback and not an unusable account.

You were right about the test too. Added `TestRestartStaysPutWhenTheSoonerAccountIsAlsoSpent` with explicit reset times (pinned account 3h, other 30m) — verified it fails against the old `next == s.Account` guard and passes against the new one. Left the original in place with a note that it passes against either, so the distinction is on the record.

internal/claudeaccount/configdir.go
hayke102: Fixed, both consequences.

Package mutex around `Provision`, same as `GitWorktreeProvider` uses for concurrent removes. And the symlink error no longer returns early: the first error is recorded, the loop continues, both mirrors run, and the error comes back at the end — a partial provision that still refreshes `settings.json` (and so fleet's hooks) beats one that doesn't.

`TestProvisionSurvivesConcurrentCallers` runs 8 concurrent provisions; verified it fails 5/5 with the mutex removed and passes with it.

internal/ui/accounts.go
hayke102: Fixed — the cancel now lives on Home behind a mutex (`setAccountLoginCancel` / `stopAccountLoginWatch`).

The `tea.Exec` completion callback stops the watcher as its **first** action, ahead of the `Identify` fallback and the `RemoveConfigDir` cleanup, so nothing is still polling a path that's about to be deleted. And starting a new attempt cancels the previous watcher, so pressing `a` again can't stack them.

internal/ui/app.go
hayke102: Fixed — the handler reads `ConfigDirFor` before `Store.Remove` drops the record, then calls `RemoveConfigDir`. Failure is logged rather than raised: the account is already gone from fleet's model, and refusing the remove over an undeletable directory is the wrong trade.

The Keychain item is deliberately left, and — to your last point — the confirm dialog now says so: *"Its Claude login stays in your Keychain; only fleet's copy of the settings is deleted."* Deleting it would log that subscription out of Claude Code entirely, which is a bigger action than "remove from fleet" implies and isn't fleet's credential to revoke. Covered by `TestRemovingAnAccountDeletesItsConfigDir`.

internal/ui/app.go
hayke102: Fixed — the budget is now `rightEdge - lipgloss.Width(h.renderHeader()) - accountReadoutGap`, measured from what was actually drawn, so a long session title shrinks the strip instead of colliding with it. A budget that goes non-positive yields `""` from the existing density loop, so no new branch was needed.

Two tests: `TestReadoutBudgetLeavesRoomForTheHeader` (with an assertion that the fixture genuinely exercises the squeeze, so it can't quietly stop testing anything) and `TestReadoutYieldsNothingWhenThereIsNoRoom`.

internal/ui/accounts.go
hayke102: Fixed all three, plus one more you didn't flag: **accounts.go:790**, `maybePollAccountUsage`'s comment still argued "the probe is the only route" and priced it at "roughly nine tokens a poll" — describing the `/v1/messages` probe this branch deleted, and inverting the actual reasoning (the endpoint is now free and available *because* each account is a real login).

Agreed on the carve-out: the `setup-token` mentions in `claudeaccount/*.go` are the documented why-not and stay.

cmd/fleet/worktree.go
```diff
@@ -206,6 +220,64 @@ func runWorktree(args []string) {
...
+			account = acct.Email
+			// Quota lives only in a running TUI's memory, so SelectOpts.Usage is
+			// empty here and least_used degrades to configured order. Said out
+			// loud: a scripted `fleet worktree` can otherwise land on a nearly
+			// spent account with nothing to indicate the strategy didn't apply.
+			if cfg.GetAccountStrategy() == claudeaccount.StrategyLeastUsed && accounts.Len() > 1 {
+				debuglog.Logger.Info("account select: no quota available from the CLI, configured order decided",
+					"chosen", account, "strategy", cfg.GetAccountStrategy())
+			}
+		}
```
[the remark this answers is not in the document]
hayke102: Fixed, though not quite as suggested — the diff collapses two cases that need opposite answers.

`Select` declines for three reasons: nothing configured (ambient is correct), an allowlist naming no configured account (the real bug you found), and every allowed account being logged out. The suggested `else if accounts.Len() > 0 { os.Exit(1) }` catches the last two together, which reverses `dropLoggedOut`'s documented decision — *"the login the user is already using probably works, and an account nobody is logged into certainly doesn't"*. Taken as written, a fleet whose logins had all expired could no longer create sessions at all.

So the two are separated by `claudeaccount.AllowedConfigured`, a question about *configuration* rather than state:

- **Allowlist names nothing configured** → refuse. `fleet worktree` and `fleet add` exit 1 naming the accounts (so a typo is distinguishable from a removed account); the TUI returns a refusal from `resolveAccount` and `handleSessionCreate` declines. Verified end to end: exit 1, message printed, no worktree created.
- **Everything logged out** → keep the fallback, but say so: stderr in the CLIs, an on-screen notice in the TUI.

Two more surfaces had the same hole and are fixed with it: `runAdd` (where I added `Allowed` in the previous round) and the TUI's `resolveAccount`, which logged a Warn and fell through. `resolveAccount` now returns the refusal rather than logging it, so no creation path can reach a session without having seen it.

The stderr note for the `least_used` degradation is in — you're right that a comment claiming "said out loud" while only writing to debug.log wasn't.

`TestAllLoggedOutIsSatisfiablePolicyNotAPolicyFailure` exists specifically to fail if someone later merges the two branches back together.

internal/diagnostics/diagnostics.go
```diff
@@ -187,14 +212,8 @@ func (r *Report) formatMarkdown(description string) string {
...
-func (r *Report) FormatEnvironmentMarkdown(includeLogs bool) string {
-	home, _ := os.UserHomeDir()
-	sanitize := func(s string) string {
-		if home != "" {
-			return strings.ReplaceAll(s, home, "~")
-		}
-		return s
-	}
+func (r *Report) FormatEnvironmentMarkdown(includeLogs bool, scrub Scrubber) string {
+	sanitize := sanitizerFor(scrub)
```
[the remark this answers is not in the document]
hayke102: Fixed — every dynamic value in `FormatEnvironmentMarkdown` now goes through `sanitize`, including `Version`, `Arch`, `OSSummary`, `KernelVersion`, `TmuxVersion`, `GhVersion` and the whole terminal block.

Worth naming the real defect: it wasn't that `LANG` is likely to hold a path — it's `en_US.UTF-8`. It's that `ClaudeVersion` and `CodexVersion` *were* sanitized and their neighbours weren't, so someone had already decided a version string can carry a path and new fields inherited whichever line they were pasted next to. That matters more now that `sanitize` also strips credentials and emails.

`TestEveryEnvironmentFieldIsScrubbed` is table-driven per field, so a regression names which one — and each case also asserts the field actually reached the output, so a test that stops exercising anything fails rather than passing vacuously. Added `TestNilScrubberStillRewritesHome` alongside it, so the new parameter can't be read as making sanitizing optional.

internal/ui/account_move_test.go
```diff
@@ -0,0 +1,350 @@
...
+	h.accounts.Upsert(claudeaccount.Account{Email: "third@x.com", ConfigDir: dir})
+
+	h.Update(accountRemoveMsg{email: "third@x.com"})
+
+	if _, err := os.Stat(dir); !os.IsNotExist(err) {
+		t.Errorf("config dir survived the remove: %v", err)
+	}
+	if h.accounts.ConfigDirFor("third@x.com") != "" {
+		t.Error("account still in the store after removal")
+	}
```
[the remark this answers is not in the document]
hayke102: Fixed, via `tea.Cmd` rather than the account worker. That worker is a poll loop with its own cadence, so handing it one-shot work means a channel and a result message for something Bubble Tea already runs off-thread; `removeConfigDirCmd` keeps the ordering plain — store update and persist stay synchronous, only the filesystem call defers.

Test updated to drain the returned batch before asserting, and it now checks both halves separately: the store must be updated synchronously (the row cannot outlive the delete on screen) while the directory must not be. Confirmed it fails against the synchronous version.

Two corrections to the finding. `RemoveAll` doesn't follow symlinks, so it unlinks the six shared links rather than walking into `~/.claude` — the dir is a handful of small files, not "large". And `persistAccounts` on the next line still writes `accounts.json` synchronously, so this is a partial fix by construction; I left that deliberately, since it's a sub-kilobyte atomic temp+rename and the store has to be consistent before the next render reads it.

PR #243, opened 2026-08-06
feat(cli): start a worktree with a prompt, message a running session

### LINKED ISSUE #242: It would be cool to support either starting a worktree from…
### Problem
It would be cool to support either starting a worktree from CLI with an initial prompt (meaning to already provide the first message as a parameter to the CLI call) or to have another fleet CLI command to send a message to a session

### Environment
- **Version**: 2.24.1
- **OS**: macOS 26.2 (arm64)

hayke102:
Closes #242 — both halves it asked for.

## What

```bash
fleet wt fix-242 -p "fix the flaky worker test"   # opens already working
gh issue view 242 | fleet wt fix-242 -p -         # -p - reads stdin
fleet send fix-242 "now run the tests and push"   # type into a live session
```

`send` names a session by `@0`-`@9` slot, exact id, title, workspace name, git branch, id prefix, then title substring. A tier matching several sessions lists them rather than guessing.

## Two decisions worth reviewing

**The prompt never enters the launch command.** `Session.Start` *types* the command into the pane's shell, so an embedded prompt would be parsed by that shell — `$(...)` and backticks would execute, and a newline would end the command early. It rides `tmux new-session -e FLEET_INITIAL_PROMPT=<raw>` instead (no shell involved), leaving only `"$FLEET_INITIAL_PROMPT"` in the command: one argv element whatever it contains. Verified the child receives `argc=1` with `$(echo pwned)` intact and unexecuted; pinned by `TestBuildLaunchCmdNeverEmbedsPromptText` and `TestInitialPromptTravelsInEnvNotCommand`.

It is **one-shot** — cleared by a successful `Start()` alongside `ForkFromID` — so pressing `r` or waking from idle-suspend can't silently re-submit the original task, and it is never persisted.

**`send`'s guards, because a pane is a typing surface, not an API.**

| state | behaviour |
|---|---|
| suspended | woken with the message as its resume prompt |
| no tmux / `pane_dead` | refused |
| pane back at a shell prompt | refused, **not** `-force`-able — the message would run as a shell command |
| agent waiting on a prompt | refused unless `-force` — typed text is discarded and the Enter confirms the highlighted option |

The waiting check reads the **pane, not the hook**: granting a permission fires no hook, so the stored hook status stays `waiting` long after the user answered — stale in exactly the direction that would block legitimate messages.

## Verified against real sessions

- multi-line piped prompt arrived intact in the pane's env and ran
- a message containing `$(echo INJECTED)` landed as **one** prompt, literal and unexecuted
- the waiting guard refused against a real `/model` menu
- the shell guard refused `echo PWNED` after `/exit`; pane stayed clean
- suspended wake produced `⏺ walnut` and persisted the new tmux name + status

`go test -race ./...` passes and `golangci-lint` reports 0 issues. The suite stays hermetic — no test spawns tmux or an agent.

## Known limits, stated rather than hidden

- `PaneIndicatesWaiting` returns false for OpenCode: its status is plugin-driven and fleet has no structural detector for its UI, so there is nothing to read. Better a caller that knows the check is unavailable than one trusting a guess.
- Waking a suspended session prints a note that an already-open TUI still shows it suspended — the TUI reads SQLite once at startup and only *adopts* rows it has never seen.
- CLI-only. The `w` dialog has no prompt field yet.


internal/agent/agent.go
```diff
@@ -88,35 +106,54 @@ type LaunchOpts struct {
...
 	}
 
-	// Claude (default).
-	cmd := "claude"
-	if o.ForkID != "" {
-		cmd += fmt.Sprintf(" --resume %s --fork-session", o.ForkID)
-	} else if o.ResumeID != "" {
-		cmd += fmt.Sprintf(" --resume %s", o.ResumeID)
+	if o.Prompt != "" {
+		cmd += " " + t.promptFlag() + promptRef
```
hayke102: The env var keeps the prompt away from the **shell**, but the **agent's own argv parser** still sees it — and here it arrives as a bare positional (Claude/Codex) or as a separate `--prompt` value (OpenCode). If that one argv element starts with `-`, Commander / clap / yargs read it as an option, not as text.

This collides with a rule the PR states elsewhere: `parseSendArgs` requires flags *before* the selector precisely so "a message is free-form user prose that can perfectly well begin with a dash" — and `send_test.go` pins that. But the suspended-wake path routes exactly that message through `InitialPrompt` → argv:

```
fleet send <suspended> "--force isn't needed here"
→ claude "$FLEET_INITIAL_PROMPT"  →  argv[1] = "--force isn't needed here"  →  unknown option
```

The worktree/tmux session is already created by then, so the failure mode is a live session parked at a shell prompt with no agent and the message gone. Same for `gh issue view N | fleet wt br -p -` when the piped text opens with a markdown bullet.

Suggest `-- "$FLEET_INITIAL_PROMPT"` for the positional agents and `--prompt="$FLEET_INITIAL_PROMPT"` for OpenCode (worth confirming `--` is accepted per agent before landing).

cmd/fleet/send.go
```diff
@@ -0,0 +1,340 @@
...
+			ag.DisplayName(), row.Title)
+	}
+	// A live tmux session is not proof an agent is listening: the pane runs a
+	// shell and the agent is a process inside it, so `/exit` (or a crash) drops
+	// the pane back to a prompt with tmux none the wiser. Text pasted there is a
+	// command line and the Enter runs it — "delete the old build dir and rerun
+	// make" would be executed by the shell. Deliberately not bypassable with
+	// -force: there is no version of this the user wanted.
+	tmux.RefreshSessionCache() // cache is populated by the TUI's tick, which isn't running here
+	if paneCmd := tmux.PaneCurrentCommand(row.TmuxSession); shell.IsShellCommand(paneCmd) {
```
hayke102: This guard has no notion of *startup*, so it misfires on the scripted flow the PR advertises.

`tmux.Session.Start` creates the pane, `send-keys` the launch command into the pane's shell, and returns; `fleet worktree` prints and exits immediately. For the few hundred ms the shell spends sourcing its rc before `exec`ing the agent, `pane_current_command` is `zsh`/`bash`. So:

```
gh issue view 242 | fleet wt fix-242 -p -
fleet send fix-242 "also update the tests"
```

fails with "the claude process in session … has exited — its pane is back at a zsh prompt", which is flatly untrue. And since this branch is deliberately not `-force`-able (rightly so), there's no way past it but to sleep and retry.

Treating `StatusStarting` — or a session whose tmux was created seconds ago — as "not yet at a prompt", or briefly retrying the pane read, would close the window without weakening the guard.

cmd/fleet/send.go
```diff
@@ -0,0 +1,340 @@
...
+		return fmt.Errorf("the %s process in session %q has exited — its pane is back at a %s prompt, "+
+			"where the message would run as a shell command.\nRestart it with r in the TUI",
+			ag.DisplayName(), row.Title, paneCmd)
+	}
+
+	if !force {
+		if pane, err := ts.CapturePane(); err == nil && session.PaneIndicatesWaiting(ag, pane) {
```
hayke102, line 228: `PaneIndicatesWaiting` returns `false` unconditionally for OpenCode (`session.go:2318`), so for that agent this guard is a silent no-op — and silence is the problem, since the failure it prevents is the destructive one.

An OpenCode session sitting on a permission prompt takes the paste (discarded by the menu) and then the trailing `Enter` **confirms whichever option is highlighted** — granting a permission the user never saw. Refusing isn't obviously right when there's no detector, but proceeding without saying so isn't either. At minimum print "cannot check whether this session is waiting" for OpenCode so `-force` is a choice rather than the invisible default.

internal/tmux/tmux.go
```diff
@@ -679,6 +679,44 @@ func (s *Session) SendLiteralKeys(text string) error {
...
+// markers are only emitted when the pane's application asked for bracketed
+// paste, so a pane that didn't is left with plain text rather than escape
+// sequences on screen.
+//
+// Enter is a separate call on purpose — it must land after the paste's closing
+// marker for the agent to read it as "submit" rather than as pasted content.
+func (s *Session) PasteAndSubmit(text string) error {
+	// Buffer name is per-session so two concurrent sends can't overwrite each
+	// other's text between load and paste.
+	buf := "fleet-send-" + s.Name
```
hayke102: The comment says per-session naming stops concurrent sends from overwriting each other, but the name is keyed on the **target session** — which is exactly what two concurrent sends to the same session share.

`fleet send x A & fleet send x B` can interleave as:

1. A `load-buffer` → buffer holds A
2. B `load-buffer` → buffer now holds B
3. A `paste-buffer -d` → pastes **B's** message, deletes the buffer
4. B `paste-buffer` → fails, "no buffer"

A's text is silently replaced by B's, A reports success, and B — whose message actually landed — reports an error. The error-path `delete-buffer` below can likewise wipe a concurrent send's freshly loaded buffer.

A pid/counter nonce in the name (`fleet-send-<pid>-<n>`) makes the comment true.

cmd/fleet/send.go
```diff
@@ -0,0 +1,340 @@
...
+	// which is the same one-shot path `fleet worktree -p` uses.
+	if session.Status(row.Status) == session.StatusSuspended {
+		s.InitialPrompt = message
+		if err := s.Restart(); err != nil {
+			return fmt.Errorf("failed to wake suspended session %q: %w", row.Title, err)
+		}
+		// Restart built a new tmux session; without persisting the new name the
+		// row points at a session that no longer exists and the next reader
+		// (TUI or CLI) reads the session as dead.
+		if err := storage.UpdateTmuxSession(row.ID, s.TmuxSessionName); err != nil {
```
hayke102, line 177: The wake persists `tmux_session` and `status`, but not `claude_session_id` — and `Restart()` may have just healed it.

`Restart()` calls `adoptResolvedLaunchID` (`session.go:744`), which can promote a resolved id over a stale/dead one (the #226 path). In the TUI the hook worker re-persists it a moment later, but `fleet send` is a one-shot process that exits right after `deliver` returns, so the healed id dies with it and the next wake resumes the same dead conversation again.

internal/session/session.go
```diff
@@ -211,6 +229,12 @@ func (s *Session) Start() error {
 	// transcript heuristic (which can't see a fork). Cleared on divergence.
 	s.forkParentID = s.ForkFromID
 	s.ForkFromID = "" // Clear after first start so restarts use session's own ClaudeSessionID.
+	// Same one-shot rule as the fork id, and for a louder reason: a prompt left
+	// in place would be re-submitted by every Restart — pressing `r` on a
+	// finished session, or waking it from idle-suspend, would silently ask the
+	// agent to redo the original task. Cleared only on success, so a failed
+	// launch can be retried with the prompt intact.
+	s.InitialPrompt = ""
```
hayke102: The comment claims the one-shot rule covers restarts ("a prompt left in place would be re-submitted by every Restart — pressing `r` … or waking it from idle-suspend"), but the clear only happens here in `Start()`. `Restart()` (session.go:739) never touches `InitialPrompt`.

It's safe today only by accident: `deliver()` sets `InitialPrompt` and calls `Restart()`, then the CLI process exits. Any caller that keeps the `*Session` and restarts it twice re-submits the original task — the exact scenario this comment says can't happen.

Either clear it in `Restart()` (and `RespawnClaude`) too, or narrow the comment to what the code actually guarantees. CLAUDE.md makes the broader claim as well.

internal/agent/agent.go
hayke102: Fixed — and confirmed on all three CLIs before landing, as you suggested. It's worse than an error: the agent **exits**, so the user is left with a live session at a shell prompt and the message gone.

```
claude "--force isnt needed here"    → error: unknown option '--force isnt needed here'
codex "--force isnt needed here"     → error: unexpected argument … tip: use '-- …'
opencode --prompt "--force isnt…"    → prints usage and exits
```

Codex names the fix in its own error text. Now `-- "$FLEET_INITIAL_PROMPT"` for Claude/Codex and `--prompt="$FLEET_INITIAL_PROMPT"` for OpenCode — with a space, yargs reads the value as a fresh option. Verified end-to-end: `fleet wt dash-test -p "--dash-leading prompt…"` launches and answers it. Pinned by `TestBuildLaunchCmdSeparatesPromptFromOptions`.

cmd/fleet/send.go
hayke102: Fixed, and reproduced exactly as you described first. Worth recording what the retry alone did, because it was a trap: it closed the false refusal and then **silently ate the message** — `fleet send` printed "Sent" and nothing ever reached the agent.

`pane_current_command` leaving the shell proves the *process* started, not that the TUI is reading input. Measured:

| signal | timing |
|---|---|
| process leaves shell | 649ms … 7415ms |
| pane content settles 600ms | 783ms — fires ~2s early |
| input box actually renders | 2972ms |

So there's a ~2.4s window where a paste is discarded, and nothing available here predicts it — waiting for the pane to settle looked like the elegant agent-neutral answer until measurement killed it (the screen sits static mid-startup). Per-agent readiness detection argues against itself too: my probe caught **Codex opening on an update menu** (`› 1. Update now`), which is exactly the state we must never paste into.

So readiness isn't predicted — the outcome is checked. Two waits: up to 10s for the pane to stop reading as a shell (sized for the 7.4s measurement), then `confirmDelivered` polls for the message's whitespace-squashed first line, or Claude's `[Pasted text` placeholder since a multi-line paste is collapsed and never appears verbatim. If it never lands, the command says so instead of lying. No auto-retry — a re-paste on a false negative submits twice.

```
# immediately after create
the message was not delivered — "v2-test" was still starting and its input
wasn't ready yet.  Nothing reached the agent; try again in a moment    # exit=1

# moments later
Sent to 'v2-test' (Claude)        # 65ms — verification is free on a live session
```
Message arrived exactly once; the exited-agent guard still refuses.

cmd/fleet/send.go
hayke102: Fixed — it now prints the caveat instead of withdrawing in silence:

```
Note: fleet can't tell whether an OpenCode session is waiting on a prompt.
If it is, this message is discarded and its Enter confirms the highlighted option.
```

Went with warn-and-send over refusing: OpenCode defaults to allow-all permissions, so refusing every send would make `-force` reflex and disable the guard everywhere — protecting against a state most sessions are never in by weakening the one case that matters. You're right that documenting it in the PR body and a doc comment wasn't a mitigation; nobody reads either at the moment they send.

internal/tmux/tmux.go
hayke102: Fixed — buffer is now `fleet-send-<pid>-<atomic counter>`, and the comment says what actually makes it safe. You're right that I keyed it on the one thing two racing sends have in common; the interleaving you traced is real, and the inverted reporting (A says success having sent B's text, B errors having actually landed) is the part that would have been miserable to debug. Confirmed no `fleet-send-*` buffers leak after the runs.

cmd/fleet/send.go
hayke102: Fixed — the wake now persists the id too when `Restart` changed it:

```go
if id := s.ClaudeSessionID; id != "" && id != row.ClaudeSessionID {
    storage.UpdateClaudeSessionID(row.ID, id)
}
```

Confirmed `adoptResolvedLaunchID` (session.go:785) overwrites it under `s.mu` on the way into `Restart`. The TUI gets away with not persisting there because its hook worker follows up; this process exits inside `deliver`, so the heal was recomputed and discarded on every wake and a session pinned to a dead conversation stayed pinned.

internal/session/session.go
hayke102: Fixed by making the code keep the promise rather than narrowing the comment — the trap was the point of the finding, and `RespawnClaude` would have been the next one to step in it.

All three launch paths now call `consumeInitialPromptLocked()` after a successful launch (`Start`, `Restart`, `RespawnClaude` — the three that build from `buildAgentCmd` + `sessionEnv`), so the clear travels with the thing that needs it. Field doc and CLAUDE.md updated to describe that, and `TestInitialPromptIsOneShot` asserts a consumed prompt leaves both the command and the env identical to a promptless launch.

PR #246, opened 2026-08-09
feat(cli): teach agents to drive fleet with an installable skill

hayke102:
fleet's CLI grew commands an agent could use — `fleet wt -p`, `fleet send` — but nothing told the agent they exist. `fleet skill install` writes an [Agent Skills](https://agentskills.io) `SKILL.md` into the skill directory each agent reads, so Claude Code, Codex, Cursor, and OpenCode can hand work to a worktree session or message a running one without the user relaying commands.

```
$ fleet skill install
✓ claude    installed  ~/.claude/skills/fleet/SKILL.md
✓ codex     installed  ~/.agents/skills/fleet/SKILL.md
- cursor    skipped    not found on this machine
- opencode  skipped    not found on this machine
```

## Two roots, four agents

| Root | Reaches |
|---|---|
| `~/.claude/skills/fleet/` | Claude Code (also read by OpenCode, Cursor) |
| `~/.agents/skills/fleet/` | Codex, Cursor, OpenCode |

Neither root alone suffices: Codex reads only `.agents/skills`, Claude Code only `.claude/skills`. Cursor and OpenCode read both and are mapped to `.agents/skills`, so either one installed *without* Claude Code still lands somewhere it reads. Agents sharing a root share one write — three selected agents don't race three writes at one path.

## Decisions worth reviewing

**Opt-in, unlike hook injection.** Hooks are load-bearing (no status detection without them), so fleet installs them on launch. A skill is a convenience, and it costs its description in the context of *every* session in every repo — including repos with nothing to do with fleet. That made it the user's call rather than fleet's.

**Never prompts, never reads stdin**, so an agent can install it for itself in one call. `fleet -h` says so in as many words, which is the discovery path for agents; a `tipOnce` and the README cover humans.

**Frontmatter is only `name` + `description`** — the intersection all four targets accept (Claude Code's packaging validator rejects unknown keys outright), with `name` equal to the install directory or the skill silently never loads. `TestFrontmatterIsPortable` fails if either drifts.

**Temp-file + rename**, because Claude Code watches skill directories for live changes and can read a partially written `SKILL.md`.

**Install defaults to detected agents; uninstall defaults to all.** Detection is binary-on-PATH *or* config-dir-present — Cursor installs its `cursor` shim only on request, so `~/.cursor` is what catches those users. Uninstall is thorough because the copy worth removing usually belongs to an agent the user has since uninstalled.

## Testing

`make build`, `make lint` (0 issues), and `go test -race ./...` pass. 11 tests in `internal/skill` cover the shared-root dedupe, idempotence, outdated-vs-absent status, uninstall leaving user files alone, and config-dir detection; `cmd/fleet` covers `-agent` parsing and outcome rendering. Verified end-to-end against a sandboxed `$HOME`.

**Not verified:** no live agent session was launched to confirm the skill triggers on a real prompt — that would mean writing into a real `~/.claude/skills/`. Worth doing before merge.

https://claude.ai/code/session_01N1469qLqPsCqRoRGR7ZswU


internal/skill/skill.go
```diff
@@ -0,0 +1,271 @@
...
+	chosen := make(map[string]bool, len(selected))
+	for _, a := range selected {
+		chosen[a.Name] = true
+	}
+
+	done := make(map[string]Result)
+	results := make([]Result, 0, len(agents))
+	for _, a := range agents {
+		if !chosen[a.Name] {
+			results = append(results, Result{Agent: a.Name, Path: a.Path(), Outcome: Skipped})
```
hayke102: `Skipped` is decided by name membership, but the *operation* below is deduped by path — so an agent sharing a root with a selected one is reported as skipped after its file was already written or deleted.

Concrete: a user with Codex and OpenCode runs `fleet skill uninstall -agent codex`. Codex, Cursor and OpenCode all resolve to `~/.agents/skills/fleet/SKILL.md`, so the file *is* removed — and the output prints `- opencode  skipped  not found on this machine` while OpenCode's skill just disappeared. Install is the mirror image: `-agent codex` reports Cursor/OpenCode as skipped even though they now load the skill.

`skillOutcomeLine` goes out of its way not to make false claims about the user's machine, so this is the one place the report contradicts what actually happened. A `done[path]` lookup before emitting `Skipped` would let the line say `shared with codex` instead.

internal/skill/skill.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+	return home
+}
+
+func claudeSkillsRoot() string { return filepath.Join(hooks.GetClaudeConfigDir(), "skills") }
+func agentsSkillsRoot() string { return filepath.Join(homeDir(), ".agents", "skills") }
+
+// agents is the full table, in the order results are reported.
+var agents = []Agent{
+	{Name: "claude", binary: "claude", configDir: hooks.GetClaudeConfigDir, root: claudeSkillsRoot},
```
hayke102: `configDir` makes Claude detected on every machine that has ever launched fleet's TUI, so `Detected()` isn't a real signal for this row: `NewHome` calls `hooks.InjectClaudeHooks(...)` unconditionally (`internal/ui/app.go:6936`) and that does `os.MkdirAll(configDir, 0755)` (`internal/hooks/claude_hooks.go:209`). The adjacent Codex/OpenCode injections *are* `exec.LookPath`-gated, precisely so fleet never creates `~/.codex` for users who don't have it — Claude has no such gate.

Two consequences: a Codex-only user running bare `fleet skill install` gets a `~/.claude/skills/fleet/SKILL.md` for an agent they don't have; and `fleet skill install -agent codex` prints `- claude  skipped  not selected`, which is exactly the "false claim about their machine" the re-check in `skillOutcomeLine` was written to avoid. Binary-only detection here (drop `configDir` for this row) fixes both — Cursor is the row that genuinely needs the config-dir signal.

cmd/fleet/skill.go
```diff
@@ -0,0 +1,223 @@
...
+		selected, err := resolveAgents(opts.agentSel, skill.Agents())
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%v\n", err)
+			os.Exit(1)
+		}
+		results := skill.Uninstall(selected)
+		printSkillResults(results)
+		exitOnFailure(results)
+	case "status":
+		results := skill.Status()
```
hayke102: Minor: `-agent` parses fine on every verb, but this branch never reads `opts.agentSel` — `fleet skill status -agent codex` prints all four rows with no indication the filter was dropped. That's the same silent flag-swallowing `resolveAgents` deliberately rejects for `-agent` typos. Either honour it here or reject it for `status`.

internal/skill/skill.go
hayke102: Fixed in 4d7d13f — confirmed the uninstall case first: `uninstall -agent codex` really did delete OpenCode's file and print `skipped`.

`run` now acts in one pass and reports in a second, so the report follows the path rather than the name. An unselected agent whose file changed anyway carries a new `Result.SharedWith`:

```
✓ codex     removed  ~/.agents/skills/fleet/SKILL.md
✓ opencode  removed  ~/.agents/skills/fleet/SKILL.md (shared with codex)
```

`Status` deliberately does *not* disclose sharing — a read changes nothing, so there's no side effect to own up to, and disclosing it would make an explicit `-agent` filter look ignored. Two tests pin the install and uninstall directions, plus one asserting a selected agent is never tagged `SharedWith`.

internal/skill/skill.go
hayke102: Fixed in 4d7d13f — verified both cited lines. `app.go:6936` is a bare `hooks.InjectClaudeHooks(configDir)`, and the two calls directly below it are `LookPath`-gated with comments reading "never create ~/.codex for users who don't have it". So the signal proved nothing.

Dropped `configDir` from the claude row; `Detected()` now tolerates a nil one. That also makes the rule coherent: a config dir counts as evidence only where fleet never creates it, which is why `~/.cursor`, `~/.codex` and `~/.config/opencode` stay.

Reproduced the fix on a Codex-only PATH with a fleet-created `~/.claude/hooks` present — `claude skipped not found on this machine`, and no `~/.claude/skills` written. `TestClaudeIsNotDetectedByItsConfigDir` pins it.

cmd/fleet/skill.go
hayke102: Fixed in 4d7d13f — honoured rather than rejected. `Status` covering every agent is about *detection* (finding a skill left behind by an agent since uninstalled), and an explicit `-agent` isn't detection, so honouring it keeps the default intact and makes the flag mean one thing across all three verbs.

`Status(selected []Agent)` now, with the CLI resolving against a fallback of `skill.Agents()` so bare `status` is unchanged. `status -agent codex` prints one row and skips the rest. `TestStatusHonoursSelectionWithoutSharing` covers the filter and the no-sharing-on-reads rule together.

PR #249, opened 2026-08-18
feat(accounts): pick which Claude accounts an origin may use

hayke102:
## Why

`allowed_accounts` was already enforced everywhere — `resolveAccount`, the move-to-account picker, and both CLI paths — but could only be set by hand-editing `config.json`. `CLAUDE.md` said so outright: *"Config-only today; no editing UI yet."* So nobody discovered the feature, and a typo'd email produced a session-creation refusal with no surface that could fix it.

This adds the editing surface only. **The selection path is untouched.**

## What

`.` on an **origin header** → **"Allowed Accounts…"** — a row-anchored checkbox dropdown built on the account picker's overlay pattern.

```
▸ ◉ yuval@brizz.ai                        12% used
  ○ yuval@personal.com                    71% used
  ◉ work@acme.io                      ✕ logged out

  2 of 3 accounts allowed here
  space toggle • a all • enter save • esc
```

Origin rows only: the config key is per-origin, so offering this from a session or worktree row would put an origin-wide rule behind a control that looks like it acts on the row you're standing on.

## Decisions worth reviewing

**Ticking every account deletes the key rather than storing the full list.** The two read identically today, but a stored list silently locks out the *next* account added, with nothing on screen to explain it. `a` toggles all, so it doubles as "remove the restriction".

**Zero ticked is refused, not saved** — same rule `SnoozeDialog` applies to an unparseable duration. An empty list already means unrestricted in config, so saving it would do the opposite of what ticking nothing looks like it asks. The verdict line names which state you're in.

**A logged-out account stays tickable here** — deliberately unlike the move picker, which can't move a session *onto* a dead login. This is policy, not availability: "client work runs on the work subscription" stays true while that subscription is logged out, and refusing to record it would force a login just to write a rule down.

**Stale allowlists heal.** Rows are built from `Store.List`, so an email naming a removed account has no row and is dropped on save — opening and saving fixes the state `resolveAccount` currently refuses new sessions on.

**Sessions already pinned to a now-disallowed account are left alone.** The allowlist governs *selection*, and the prompt cache it protects is per-account.

## Files

| | |
|---|---|
| `internal/ui/allowed_accounts.go` | New dialog (cloned from `account_picker.go`) |
| `internal/config/config.go` | `SetAllowedAccounts` — creates the map, deletes on empty |
| `internal/ui/app.go` | 7 wiring points + menu item + opener + dispatch case |
| `internal/ui/allowed_accounts_test.go` | 8 tests |
| `internal/config/config_test.go` | Delete-on-empty round-trip |
| `internal/ui/context_menu_test.go` | Existing test pinned the origin menu's item list |

## Testing

`make test` (race), `make lint` (0 issues), `make build` all clean. Rendering verified against a throwaway harness across all three verdict states, a long account name, and a logged-out row — columns align, height stays stable.

**Not yet run interactively.** The dialog is covered by tests and a render pass, but anchoring against a real sidebar row hasn't been eyeballed in the running app.

## Deliberately not included

- No palette command (origin rows only)
- No sidebar marker for a restricted origin — so a rule set months ago is visible only in the dialog and via the move picker's dimmed `not allowed here`

## Known, pre-existing, not addressed here

Running two fleet instances makes config writes last-writer-wins (`config.Load()` runs once per process; `Save()` rewrites the whole file). That can silently revert an account rule set in the other terminal, which for *this* key means billing the wrong subscription. Not introduced by this PR, but it now has a higher-stakes key riding on it. Fix would be a read-modify-write in `SetAllowedAccounts`; happy to add if you want it here rather than separately.


internal/ui/app.go
```diff
@@ -7938,15 +7967,70 @@ func (h *Home) originContextMenu() (string, []ContextMenuItem) {
...
+	// accounts configured that they have "only one", which is a false claim about
+	// their setup sitting next to the dialog that would disprove it.
+	allowed := ContextMenuItem{ID: "allowed_accounts", Label: "Allowed Accounts…"}
+	switch {
+	case h.accounts.Len() == 0: // nil-safe
+		allowed.Note = "no accounts configured"
+	case h.accounts.Len() == 1:
+		// Restricting an origin to the only account there is says nothing, and
+		// would quietly become a rule the moment a second account was added.
+		allowed.Note = "only one account"
```
hayke102, line 8006: This guard blocks *clearing* a restriction as well as adding one, and that state is reachable and unrecoverable from inside fleet.

Configure accounts A and B, restrict an origin to B, then remove B in the accounts dialog. `accountRemoveMsg` clears `cfg.DefaultAccount` but never prunes `cfg.AllowedAccounts`, so the origin still names B. `resolveAccount` then hits `!AllowedConfigured` and hard-refuses every new Claude session at that origin with *"add one, or drop the restriction"* — while the only UI that could drop it renders dimmed as `only one account`. The escape is hand-editing `config.json`, which is what the changelog fragment says you no longer have to do.

The reasoning in the comment holds for *creating* a rule, but not for removing one. Two ways out:

- let the item open when `GetAllowedAccounts(OriginExpandKey(item.OriginKey))` is non-empty, even at `Len() == 1` (and mirror it in `openAllowedAccounts`, which guards on `h.accounts.Len() < 2`), or
- prune allowlists in the `accountRemoveMsg` handler, alongside the existing `DefaultAccount` clear.

The first is the smaller change and also covers an allowlist that went stale some other way; the second removes the trap at its source. Doing both wouldn't hurt.

internal/config/config.go
```diff
@@ -179,6 +179,37 @@ func (c *Config) GetAllowedAccounts(originKey string) []string {
...
+// persists it.
+//
+// An empty list deletes the entry rather than storing one. Absent already means
+// unrestricted, and that is exactly what "every account is allowed" is — so
+// writing out an exhaustive list would be a different policy wearing the same
+// face: the next account added would be silently locked out of this origin, with
+// nothing on screen to say why. Deleting also lets omitempty drop the whole map
+// back out of the file once the last restriction goes.
+func (c *Config) SetAllowedAccounts(originKey string, emails []string) error {
+	if originKey == "" {
```
hayke102: Nit: this guard is dead. The only caller passes `OriginExpandKey(msg.originKey)`, which always prefixes `"origin:"`, so `originKey` is never empty here.

The case it looks like it's protecting against still gets through: an empty `msg.originKey` becomes the bare key `"origin:"`, gets persisted, and `setInfo` reports it to the user as a saved rule. If the check is meant to protect anything, it wants to be on the un-prefixed key at the call site.

internal/ui/app.go
hayke102: Fixed in b4f0de4 — took option 1, declined option 2.

The guard is now asymmetric: `only one account` dims at `Len() == 1` only when the origin has no stored allowlist. One account is too few to write a rule with, but not too few to clear one. `openAllowedAccounts` mirrors it exactly so a lit row can't dead-click and a dimmed one can't be reached another way.

Confirmed the trap first: with accounts [A] and an allowlist naming [B], `!AllowedConfigured` refuses every new session, and both guards dimmed exactly then. Also worth recording — `Len() == 0` is *not* trapped, because `resolveAccount` short-circuits to the ambient login before it consults an allowlist, so there's no refusal to escape.

On option 2: I think pruning on removal would hurt. Since empty *is* unrestricted, pruning-to-empty necessarily widens the policy silently — and `AllowedConfigured`'s own doc comment names this case and rules it out:

> an allowlist naming no configured account — a typo, **or an account since removed**. Falling back here is the one genuinely wrong answer

Partial pruning ([A,B] → [A]) contradicts nothing but is cosmetic: `filterAllowed` already ignores emails naming no account, so behaviour is identical either way. Happy to revisit if you disagree.

Regression test added, and verified it fails on the pre-fix code.

internal/config/config.go
hayke102: Fixed in b4f0de4 — right on both counts.

Dropped the dead check and moved it to the raw key at the call site, in the `allowedAccountsSetMsg` handler, where empty is actually expressible. Comment there now says why it can't live in `SetAllowedAccounts`.

Not reachable today (`openAllowedAccounts` already refuses `item.OriginKey == ""`), so this was correctness-of-expression rather than a live bug — but a guard that reads as protection and provides none is worse than no guard.

PR #251, opened 2026-08-18
feat(accounts): balance by week, and set the strategy where the accounts are

hayke102:
Two changes to how new sessions pick a Claude subscription.

## 1. Least-used now ranks on the weekly window

Splits into `least_used_weekly` (**new default**) and `least_used_5h`.

The two buckets fail on different timescales. A 5-hour bucket refills this afternoon, so losing that race costs an hour; a weekly bucket spent on Tuesday is gone until the following week. Ranking on the fast window let every account's slow window drain in lockstep — the state that takes the whole fleet down at once.

Only `Select`'s ranking changes: `pctOf` reads `Usage.Pct(rankWindow(strategy))` instead of `FiveHourPct`. The accounts list, move picker, and header readout still show the 5-hour figure.

**Migration.** The pre-split `least_used` survives as an *alias, not a value*:

| config | resolves to |
|---|---|
| `""` (unset) | `least_used_weekly` |
| `"least_used"` (legacy) | `least_used_weekly` |
| `"least_used_5h"` | ranks on `FiveHourPct` |
| `"least_used_weekly"` | ranks on `SevenDayPct` |
| `waterfall` / `manual` | unchanged |

`Strategies` — the ordered set both pickers offer — omits the alias, so cycling can never land on it. Mapping it to the 5-hour variant instead was rejected: it would make the *presence* of the key change behaviour, so one file would behave two ways depending on whether it had ever been written.

## 2. The strategy is editable from the accounts dialog

```
Claude Accounts

Strategy: Least used · weekly                    s change

▸  yuval@brizz.ai · Max                          12% used
   yuval@personal.com · Pro                      71% used
────────────────────────────────────────────────────────
a add   r rename   d remove   J/K order   s strategy   esc close
```

It belongs there as well as in Settings because it governs the list below it — and because it's what makes the `★` legible: a marker appearing only under `manual` reads as a property of the *account* rather than of the mode.

Before this, the dialog let you pin a default account with `⏎` and then said *"Set the account strategy to Manual in Settings to pin an account"* (`app.go:1023`) — telling you to leave the dialog you were standing in. That message now points at `s`.

Cycling to `manual` with nothing pinned toasts a prompt to press `⏎`. `Select` falls through to the automatic modes when the pin names no configured account, so that state isn't broken — it's silently not doing what the mode says.

`StrategyLabel` lives in `claudeaccount`, not the UI, following `Window.Name()`'s stated reasoning: two surfaces render these strings and two copies is how they drift.

## Drive-by

Drops the footer's `p paste` hint. The paste-a-token login path went with the config-dir rewrite and the hint outlived it — there was never a handler, so pressing `p` did nothing. Confirmed before removing.

## Testing

`make test` (race), `make lint` (0 issues), `make build` all clean.

Worth a reviewer's attention: **the `used()` test helper set only `FiveHourPct`**, so every account tied at 0 on the weekly window and several ranking tests would have started passing for the wrong reason. It now sets both windows, and `TestLeastUsedRanksOnTheStrategysWindow` uses a fixture where the two *disagree* — so the window plumbing can't be removed without a failure. Also added: strategy cycling/wrap, the manual-needs-default detector, footer-has-no-dead-keys, label distinctness, and alias resolution.

Rendering verified against a throwaway harness in all four modes plus manual-with-and-without-a-pin — columns align, the row truncates gracefully at narrow widths.

**Not exercised interactively.** Same caveat as #249: covered by tests and a render pass, but nobody has pressed `s` in the running app.


internal/ui/settings.go
```diff
@@ -707,16 +707,9 @@ func buildSettingsCategories() []settingsCategory {
...
-						return "Manual"
-					default:
-						return "Least used"
-					}
+					return claudeaccount.StrategyLabel(c.GetAccountStrategy())
 				},
-				valueW: func() int { return maxStrW([]string{"Least used", "Waterfall", "Manual"}) },
+				valueW: func() int { return maxStrW(claudeaccount.StrategyLabels()) },
 				cycle: func(d *SettingsDialog, dir int) {
 					d.cfg.AccountStrategy = cycleString(d.cfg.GetAccountStrategy(), accountStrategySet, dir)
```
hayke102, line 714: `GetAccountStrategy()` (`internal/config/config.go:162`) was never updated for the split — it still switches on only `waterfall`/`manual` and returns `"least_used"` for everything else. It's the only reader of `AccountStrategy`, so `least_used_5h` never survives a round-trip through config, and `ParseStrategy` resolves the alias back to weekly. Three consequences:

- **The cycler can't advance.** `"least_used"` is deliberately absent from `Strategies`, so `indexOf` → -1 → `idx = 0` → forward always yields `Strategies[1]` = `least_used_5h`. Pressing `→` repeatedly gives `least_used_5h, least_used_5h, …` forever; Waterfall and Manual are unreachable by forward cycling (backward is likewise stuck on `manual`).
- **The label always reads "Least used · weekly"** (line 710), so the row looks frozen — the key writes a value the label never reflects.
- **The 5-hour mode is inert.** `Select` always ranks on the weekly window; `rankWindow`'s 5h branch is only ever reached by tests that call `Select` directly.

`accounts.go:679` has the same shape: with `least_used_5h` on disk the dialog states "Strategy: Least used · weekly", and the first `s` cycles weekly→5h — re-saving the value already in force, so the keypress appears to do nothing.

One-line fix: have `GetAccountStrategy` delegate to `claudeaccount.ParseStrategy`, which already handles trim/lowercase and the legacy alias. Worth doing alongside `cmd/fleet/worktree.go:267`, which tests `== claudeaccount.StrategyLeastUsed` — it only fires today *because* of this bug, and goes silent for both least-used modes the moment it's fixed (wants `StrategyLeastUsedWeekly || StrategyLeastUsed5H`).

The existing tests don't catch any of this: `select_test.go` calls `Select`/`ParseStrategy` directly, and the dialog tests pass an already-normalized literal to `Show` — nothing goes through `config.Config`.

internal/claudeaccount/select.go
```diff
@@ -100,24 +177,29 @@ func Select(o SelectOpts) (Account, bool) {
 		return live[0], true // already ordered by Store.List
 	}
 
+	// Which window "least used" means is the strategy's whole content: the two
+	// buckets fail on different timescales, so ranking on the wrong one balances
+	// a resource that was never scarce.
+	win, _ := rankWindow(o.Strategy)
```
hayke102: `rankWindow`'s second return is dropped, so the `manual` fall-through ranks on the wrong window.

Under `StrategyManual`, when the pinned account was removed or is excluded by the origin allowlist, `Select` deliberately falls through to the automatic ranking below — but `rankWindow("manual")` returns `(WindowFiveHour, false)` and the `false` goes to `_`. That path then balances the 5-hour window while the documented default (and the comment right above this line) is weekly, with nothing in the UI saying so.

Either honour the bool (`if !ok { win = WindowSevenDay }`) or make `rankWindow`'s default arm return `WindowSevenDay`.

internal/ui/settings.go
hayke102: Fixed in 96df5ee — correct on all three counts, and this one was bad: the headline mode of the PR did not work at all.

Proved it before fixing rather than reading:
```
stored="least_used_5h"  GetAccountStrategy()="least_used"  label="Least used · weekly"
  press 1: "least_used" -> "least_used_5h"
  press 2: "least_used" -> "least_used_5h"
  press 3: "least_used" -> "least_used_5h"
```
And after:
```
stored="least_used_5h"  GetAccountStrategy()="least_used_5h"  label="Least used · 5-hour"
  press 1: "least_used_5h" -> "waterfall"
  press 2: "waterfall" -> "manual"
  press 3: "manual" -> "least_used_weekly"
  press 4: "least_used_weekly" -> "least_used_5h"
```

Took the one-line fix — `GetAccountStrategy` delegates to `ParseStrategy`. No import cycle: claudeaccount doesn't import config. Also stripped the value list from both doc comments there, since a stale copy of that list in the config package is precisely what caused this.

For `worktree.go:267` I went slightly past the suggestion. `weekly || 5h` would be correct today but breaks the same way the next time the family grows, so it now asks `claudeaccount.RanksByUsage(...)` — callers outside the package can ask "does the quota reading matter here" without naming members.

Your last paragraph was the most useful part: nothing went through `config.Config`, which is exactly where the bug lived. The new tests cross that boundary — round-trip every strategy through the getter, and drive the Settings cycler writing back to config each step. Confirmed each fails against the unfixed code (the cycler one reports "visited 1 of 4").

internal/claudeaccount/select.go
hayke102: Fixed in 96df5ee — right, and I took the second option plus one step.

`rankWindow` is now total, returning `Window` alone with `WindowSevenDay` for the non-ranking strategies. I introduced that bool to avoid a `WindowNone` sentinel, but its only caller discarded it — a return value nothing reads isn't a safeguard, it's just a second way to be wrong. Honouring it at the call site would have left the same trap for the next caller.

Defaulting to the weekly window is also the semantically right answer rather than just a safe one: a user on `manual` has expressed no ranking preference, so the fall-through should behave exactly like an unset config.

Pinned by `TestManualFallThroughRanksOnTheDefaultWindow` — manual with a pin naming a removed account, and a fixture where the two windows disagree. Verified it fails against the old `(WindowFiveHour, false)` version ("manual fall-through = a, want b").

PR #254, opened 2026-08-19
feat(accounts): show both quota windows, in a shape that holds still

hayke102:
## What

The top-right account readout showed **one** quota window at a time and swapped which one every 6 seconds off the wall clock, then picked one of **six** densities by measuring whatever the breadcrumb left over. A corner of the screen changed on a timer *and* reshaped whenever the cursor moved to a longer session title — and the number you wanted was one you had to wait for.

Both windows now render every frame, 5-hour first, each carrying its own reset countdown:

```
split     yuval 12%(2h) 34%(5d) │ work 99%(42m) 71%(2d)
grouped   yuval 12%/34% (2h/5d) │ work 99%/71% (42m/2d)
```

Picked in Settings (`S` → Appearance → **Account usage**: Split / Grouped / Off) rather than by measurement.

## Notes for review

**Neither style is called "compact", because neither is narrower.** Both carry the same four tokens plus exactly five characters of glue (split spends them on two paren pairs and a space; grouped on a slash, a space, a paren pair and a slash), so they are the same width by construction. That equality is load-bearing and pinned by `TestBothStylesAreTheSameWidth` — if one style were narrower, the available budget would silently decide which grouping you saw, putting the shape back under the breadcrumb's control.

**Nothing names the windows any more — order does.** The strip used to lead with the word `weekly` / `5-hour` only because one was on screen at a time. The countdowns cannot stand in for it: a weekly window about to roll over reads `3h` exactly like a 5-hour one. So 5-hour leads in every style and every state, and that ordering has its own assertion.

**A spent window no longer collapses the chip to `spent 2h`.** `quotaStyle` already paints anything at `ExhaustedPct` red — the same threshold `Select` stops handing the account out at — and the countdown sits right beside it, so `100%(2h)` says everything the old wording did without the chip changing shape on a state change. It also means the *other* window's figure survives instead of being hidden, which is what `TestSpentWindowStaysInPlace` checks.

**The fit ladder is two steps**: drop the account names, then render nothing. It never falls back to the *other* style, for the same reason as above. `TestLadderHasExactlyTwoSteps` sweeps the budget and fails if a third shape appears.

**`grouped` drops its countdown parenthetical all-or-nothing.** Pooled countdowns make `12%/34% (5d)` unreadable — nothing says which window the `5d` belongs to. `split` has no such problem (each parenthetical is attached to its own figure) so there they drop independently. Both cases are tested.

**Migration**: `account_usage_style` supersedes `show_account_usage`. Only `false` carries information (→ `off`); an unset or explicit `true` bool was the default, not a choice about styles that didn't exist. An unrecognized style falls back to `split` rather than hiding — deliberately unlike `GetTelemetryMode`, where the conservative answer is to send less; here a typo silently removing a readout with no error anywhere is the worse outcome. The Settings cycler clears the legacy bool on write, so the two keys never contradict each other in a file that gets pasted into bug reports.

## Verification

`TestReadoutIsStableOverTime` is the regression test for the whole change — three renders at different clocks must be byte-identical. I confirmed it fails against a deliberately reintroduced rotation:

```
--- FAIL: TestReadoutIsStableOverTime
    split: readout changed after 7s with identical input:
     was "yuval 12%(2h) 34%(5d) │ work 20%(2h) 71%(5d)"
     now "yuval 12%(2h) │ work 20%(2h)"
```

Also rendered the real strip and the Settings pane out of a harness (all four budget steps, logged-out chips, both styles) rather than trusting assertions alone — the settings row fits the column layout with the longer `Grouped` value.

`make test` / `make lint` / `make build` all green.


internal/ui/settings.go
```diff
@@ -555,11 +556,36 @@ func buildSettingsCategories() []settingsCategory {
...
+						return "Split"
+					}
+				},
+				valueW: func() int { return maxStrW([]string{"Off", "Split", "Grouped"}) },
+				cycle: func(d *SettingsDialog, dir int) {
+					d.cfg.AccountUsageStyle = cycleString(d.cfg.GetAccountUsageStyle(), accountUsageStyleSet, dir)
+					// Supersede the legacy bool, the way the Telemetry row does.
+					// The style already wins on read, but leaving
+					// show_account_usage:false beside account_usage_style:"full"
+					// on disk is a contradiction in a file that gets pasted into
```
hayke102, line 583: `account_usage_style:"full"` isn't a value — the set is `split`/`grouped`/`off`. Minor, but this is the one comment that documents the legacy-bool supersession, so the example should name a real pairing: `show_account_usage:false` beside `account_usage_style:"split"` is the contradiction being described.

internal/ui/accounts_header.go
```diff
@@ -7,100 +7,59 @@ import (
...
+// is nothing worth showing.
 //
 // Returns empty for a single account: with one subscription there is no choice
 // being made, so the number is trivia rather than information, and the header
 // row is better spent on nothing.
 //
-// budget is the columns available before the What's New badge; nothing wider
-// than it is ever returned.
-func renderAccountUsageHeader(accounts []claudeaccount.Account, usage map[string]claudeaccount.Usage, budget int, now time.Time) string {
+// style is one of config.AccountUsageSplit / AccountUsageGrouped. budget is the
```
hayke102: The doc says "style is one of Split / Grouped", and `accountChip` falls through to the split branch for anything that isn't `Grouped` — so `AccountUsageOff` renders the strip rather than hiding it. The only thing honouring "Off" is the `AccountUsageStyle != config.AccountUsageOff` guard at the single call site (`app.go:2029`).

Works today, but the knowledge lives outside the renderer: a second caller (an Appearance preview, say — the pattern `RenderSidebarPreview` already establishes) silently ignores the user's "Off". An early `if style == config.AccountUsageOff { return "" }` makes the function total over its own parameter and costs two lines.

internal/ui/accounts_header.go
```diff
@@ -7,100 +7,59 @@ import (
...
+// columns available before the What's New badge; nothing wider than it is ever
+// returned.
+func renderAccountUsageHeader(accounts []claudeaccount.Account, usage map[string]claudeaccount.Usage, style string, budget int, now time.Time) string {
 	if len(accounts) < 2 {
```
hayke102, line 61: This guard counts *all* accounts, but the strip renders `shown` — the filtered set that drops accounts with no reading and no logged-out flag. With exactly two accounts where the second has never been polled (up to ~180s after adding it, or indefinitely if its polls keep failing), the readout renders a single chip: precisely the "one subscription is trivia" case the doc four lines up says returns empty.

Pre-existing behaviour carried over unchanged in the rewrite, not something this commit introduced — but the rewrite is where the intent got written down, so gating on `len(shown) < 2` after the filter loop would make the code match the doc. (`len(shown) == 0` below then becomes redundant.)

internal/ui/settings.go
hayke102: Fixed — the comment now names `account_usage_style:"split"`. Leftover from the mid-implementation rename off Full/Compact: identifiers and labels were updated, this prose wasn't.

internal/ui/accounts_header.go
hayke102: Fixed — added the early `if style == config.AccountUsageOff { return "" }`, and the doc comment now names all three values. Agreed on the reasoning: the call-site guard stays as the early-out that skips the budget maths, but the rule itself belongs to the parameter. `TestOffRendersNothing` covers it (verified failing against the pre-fix code: `Off still rendered a strip: "a 10%(2h) 20%(5d) │ b 30%(2h) 40%(5d)"`).

internal/ui/accounts_header.go
hayke102: Fixed — `len(shown) < 2` after the filter, replacing the now-subsumed `== 0`; the `len(accounts) < 2` early return stays as the fast path that skips the allocation. Worth adding to your reasoning: a one-chip strip is also an extra *shape*, since it would grow to two chips when the second poll lands — the reshaping this PR exists to remove. `TestSecondAccountWithNoReadingHidesTheStrip` covers it (verified failing before: `rendered a lone chip for a two-account setup: "polled 10%(2h) 20%(5d)"`).

PR #259, opened 2026-08-20
fix(linear): seed a ticket prompt only when the worktree is created

hayke102:
## The bug

A session added by hand (`a` / `n` / `A`) to a checkout that once held a Linear ticket opened **already working on that ticket** — even when the checkout had since moved on to `master`.

`handleSessionCreate` ran `ticketPromptFor` on *every* session creation, seeding `Session.InitialPrompt` two ways:

- **Reuse** — `linear.ExistingPrompt(path)` read back the `prompt.txt` already sitting in `.fleet/ticket/<ID>/`. It never looked at the branch, so the prompt was re-asked on every session after the first, forever. This is the reported case.
- **Inference** — a branch or worktree dir name matching `<TEAM>-<N>` fetched and materialized the ticket, then seeded from it.

## The fix

A seeded first message is the *gesture* of starting a worktree from a ticket, not a property of the directory that worktree happens to be. Both branches are gone; `handleSessionCreate` now passes `msg.prompt` through untouched.

The worktree-creation path (`app.go`) already set `prompt` explicitly, and `ticketPromptFor` early-returned when it was non-empty — so nothing there changes.

| | before | after |
|---|---|---|
| `a` / `n` / `A` in a ticket worktree | prompt re-seeded | empty |
| `a` / `n` / `A` on a ticket-named branch | fetch + seed | empty |
| `w` + a ticket | prompt ✓ | prompt ✓ |
| `fleet worktree --ticket` / `-p` | prompt ✓ | prompt ✓ |

## Also removed

The negative pin (`NegativelyPinned` / `pinNoTicket` / `.no-ticket`) existed only to stop the inference path re-asking Linear on every session start. With that path gone it would be a write-only file, so it goes too. `pathTailAfterRepo` stays — `palette_tickets.go` still uses it.

## Guard

`TestManualSessionCreationSeedsNoPrompt` fails if `handleSessionCreate` reaches into `internal/linear` again, or if anything in `app.go` assigns to a `.prompt` field on the way to the launch. Verified it fires by temporarily reintroducing an assignment.

## Known cost

A worktree created **outside** fleet on a ticket-named branch no longer materializes its ticket — that was the inference branch. Deliberate, per the scoping decision: manual session creation seeds nothing, ever.

## Testing

`make build` passes. `go test ./...` is green except `TestCredentialResolutionOrder`, which fails identically on unmodified `master` on this machine (it asserts "nothing stored" while the local keychain holds a real Linear credential).

https://claude.ai/code/session_01WBQUYTfR7kzUjgozd58mr6

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linear ticket error handling by clearly reporting when an issue cannot be found, while preserving disconnected-state behavior.
  * Prevented manually created sessions from inheriting ticket prompts or performing automatic ticket lookups.
  * Ensured ticket briefings are added only when creating a worktree explicitly from a ticket.

* **Documentation**
  * Added release documentation describing the updated manual session behavior.


internal/ui/ticket.go
```diff
@@ -56,83 +46,6 @@ func materializeTicket(worktreePath string, t *linear.Ticket, moveState bool) (*
 	return &res, nil
```
hayke102, line 46: **`ErrNotFound` is now swallowed on the one path where it matters.** (Real location: `ticketStatusLine` at ticket.go:70 — anchored here because that block isn't in the diff.)

`ticketStatusLine` returns an empty line for `ErrNotFound`, so `setInfo` never fires. That swallow was justified by branch *inference* — "a branch naming no real issue is a resting state" — which is the exact path this PR deletes.

The only caller left is worktree creation, where the user explicitly picked the ticket in the `w` dialog. `Materialize` re-fetches via `fetchFull`, so if the issue is deleted, or the user's access to it changes, between the dialog's fetch and the materialize, they get a worktree and a session with **no prompt and no message at all** — silently, on the one path where "no such issue" is now actionable rather than a resting state.

The updated CLAUDE.md bullet still states the old rationale verbatim. Worth surfacing `ErrNotFound` here now that inference is gone — `ErrNotConnected` can't reach this caller anyway, since the dialog's fetch already succeeded.

internal/linear/materialize.go
```diff
@@ -20,13 +19,12 @@ import (
...
-	imagesDir   = "images"
-	ticketFile  = "ticket.md"
-	promptFile  = "prompt.txt"
-	metaFile    = "meta.json"
-	noTicketPin = ".no-ticket"
+	fleetDir   = ".fleet"
+	ticketDir  = "ticket"
+	imagesDir  = "images"
+	ticketFile = "ticket.md"
+	promptFile = "prompt.txt"
```
hayke102, line 26: **`prompt.txt` is write-only now.** With `ExistingPrompt` deleted, nothing in the tree reads it — the sole remaining reference is the write at materialize.go:132.

Concretely: `fleet wt fix-x --ticket BRZ-1 --no-session` materializes the ticket and writes the briefing, then pressing `a` on that checkout starts an agent with no ticket context, and there is no in-app gesture to recover the prompt sitting right there on disk. Same for any ticket worktree created before a fleet restart.

Probably an accepted consequence of the one-shot rule, but as it stands the write is dead weight and the flow has no replacement gesture. Either drop the file, or give it a way back in.

internal/ui/worktree_ticket_routing_test.go
```diff
@@ -126,3 +126,49 @@ func TestTicketInferenceNeverShellsOutFromUpdate(t *testing.T) {
...
+// and the launch may write to that field.
+func TestManualSessionCreationSeedsNoPrompt(t *testing.T) {
+	if mentions(t, "handleSessionCreate", "linear") {
+		t.Error("handleSessionCreate reaches into internal/linear — the only ticket " +
+			"fetch left is at worktree creation, and putting one back here re-seeds " +
+			"the prompt into every manually added session")
+	}
+
+	fset := token.NewFileSet()
+	f, err := parser.ParseFile(fset, "app.go", nil, 0)
```
hayke102, line 163: **The guard is broader than the invariant it names, and narrower than the regression it guards.**

This walks *all of `app.go`* for any `AssignStmt` whose LHS is a selector named `prompt`, not just the create-to-launch path the doc comment describes.

Both directions bite. A future unrelated struct in `app.go` with a `prompt` field — a dialog, a form — doing `d.prompt = ""` fails this test with an error message about Linear tickets inheriting into manually created sessions. And because it only parses `app.go`, the actual regression shape could reappear in `session_create.go` or `dialogs.go` (both build `sessionCreateMsg`) undetected.

Scoping the walk to `handleSessionCreate` — the `mentions`-style lookup just above already locates it — would match the stated invariant on both counts.

internal/ui/ticket.go
hayke102: Fixed in 21ca14e — `ErrNotFound` now gets its own line: "Linear: issue not found — worktree created without the ticket".

Confirmed the reasoning: `ticketStatusLine` has exactly one caller left (`app.go:1697`), since this PR deletes the `ticketReadyMsg` one. So every remaining invocation is a worktree the user explicitly picked a ticket for.

One wrinkle worth recording — it can't just fall through to the generic `Linear: %v` branch, because the sentinel is `errors.New("linear: issue not found")` and that renders `Linear: linear: issue not found`. Worded string instead.

Left `ErrNotConnected` swallowed, for the reason you gave: the dialog's own fetch had to succeed for there to be a ticket to materialize, so it can barely reach this caller.

Also updated the CLAUDE.md rationale — you were right that it was a line this branch wrote and immediately outdated.

internal/linear/materialize.go
hayke102: Keeping the write, deliberately.

You're right that nothing in the tree reads it — but it isn't quite dead weight, because there's a recovery gesture that works today:

```
fleet send <session> - < .fleet/ticket/BRZ-1/prompt.txt
```

`fleet send`'s lone `-` reads stdin precisely so a message can be piped, and it explicitly promises a message may start with a dash. That covers the `--ticket --no-session` case you named without reintroducing automatic seeding, which is the whole point of this PR.

So the file is a payload rather than a ledger now. Dropping it would remove the only way back in and wouldn't fix the underlying gap — that a materialized ticket has no *in-app* delivery gesture — it'd just remove the reminder of it. That gesture (a context-menu "brief this session on the ticket") is a real feature and wants its own PR and changelog fragment.

`ticket.md` and the screenshots are untouched either way; `prompt.txt` is only the short "read this, don't start yet" pointer.

internal/ui/worktree_ticket_routing_test.go
hayke102: Fixed in 21ca14e — the walk is scoped to the `handleSessionCreate` FuncDecl now, which is what the doc comment already claimed.

The over-broad half was the real bug and I verified it both ways: an `x.prompt = …` inside `handleSessionCreate` still fails the test, and one elsewhere in `app.go` (a struct with a `prompt` field, exactly your dialog/form case) now passes. `sessionCreateMsg` already declares `prompt string`, so that name is live in the package and the misleading-failure risk was live with it.

On the narrower half — I looked and it buys less than it seems. Neither `session_create.go` nor `dialogs.go` can host the regression shape as an `AssignStmt`: `session_create.go:60` builds `sessionCreateMsg{path: …, title: …, agent: …}` as a composite literal, which is the *legitimate* way to set a prompt — identical in form to what the worktree path does — and a file-wide `AssignStmt` walk wouldn't have caught it there either. So widening the file set costs a false positive and catches nothing extra.

A rename stays loud rather than silent: the lookup `t.Fatal`s with "renamed? this guard is now vacuous" instead of quietly walking an empty node set. Wrote the reasoning into the doc comment so the next person doesn't re-widen it.

PR #265, opened 2026-08-23
fix(perf): stop the UI stalling on transcript scans and tmux forks

hayke102:
## What

`/debug-perf` on 519 perfwatch dumps: **277 of them (53%) are `previewTickMsg` stalls** — blocking I/O on the Bubble Tea Update goroutine, the thing CLAUDE.md forbids. This fixes all three links in the chain.

```
ReadClaudeSessionName full-scans a transcript every 30s per session
  → status worker blocks, so RefreshSessionCache stops running
  → the 2s tmux session cache goes stale
  → IsAlive() cache-misses in Update() and forks `tmux has-session`
  → unbounded, on a server busy precisely because the worker is hammering it
  → 500ms+ freeze
```

Today's 10:01 dump shows both ends at once: goroutine 899 (status worker) in `syscall.read` inside `forEachTranscriptLine`, while goroutine 1 sits in `exec.Cmd.Wait` under `IsAlive` ← `Update`.

## 1. Transcript scan — the root cause

Claude's JSONL is append-only and its titles sit near the **end** of a file that reaches tens of MB, so the walk re-read the whole thing every 30s to find a string it had already found last time. `forEachTranscriptLine` now delegates to an offset-aware `scanTranscriptFrom`, and a per-path memo holds size/mtime/offset plus the last-seen titles.

**Measured on a real 89MB transcript: 69ms → 15µs.**

The scan is only sound because the file is append-only, so the memo guards the ways that assumption bends:

- A trailing chunk with no newline is a line **mid-write** — `fn` still sees it (matching the whole-file walk), but it is *not* counted as consumed, or its completion would never be read.
- `custom-title` precedence now has to survive **across reads**, not within one walk's locals.
- A file that **shrank** (truncation, reused session id) resets the memo.

## 2. `IsAlive()` off the Update goroutine

New `IsAliveCached() (alive, known bool)` answers from the cache alone. The four hot callers split deliberately on what an *unknown* means:

| Site | Cadence | On a cache miss |
|---|---|---|
| `previewTick` | 500ms | fetch anyway — it's a `tea.Cmd`, so it costs nothing here |
| `fetchPreviewForSelected` | **every `j`/`k`** | same |
| `focusTick` | 100ms | keep focus mode — only *known*-dead drops it |
| `handleFocusKey` | **every keystroke** | same |

Ejecting the user from focus mode because the cache happened to be stale is the worse of the two ways to be wrong.

Two of these were beyond the original diagnosis: `fetchPreviewForSelected` is reached from `j`/`k`/`shift+↑↓` and ~10 other keys, and `focusTick` is 100ms not 500ms. Both match the 60 `KeyMsg`/`KeyPressMsg` stalls in the dumps.

## 3. Bounded `tmux has-session`

`hasSessionTimeout = 2s`, matching the neighbouring `listPanesTimeout`. This caps the one-off callers still on `IsAlive` (attach, quick-approve, context menu, paste) — left authoritative on purpose, since they want the true answer and are worth a bounded wait.

## Testing

`make lint` clean, `go test -race ./...` green (the `internal/linear` `TestCredentialResolutionOrder` failure is pre-existing on `master` — a real Linear credential in the local keychain leaking into the test).

The incremental scan is the risky change, so `TestReadClaudeSessionNameIncrementalMatchesFullScan` pins it against the full walk it replaces at every step. Mutation-tested — both obvious ways to break it fail the test:

- consuming the partial trailing line → `completed line: incremental read = "Renamed", want "Completed"`
- dropping the truncation reset → `after truncation: incremental read = "Completed", want "fresh start"`

## Not in scope

The perfwatch dumps also show a **mouse-wheel render storm** (267% CPU on a `MouseMode` that has no handlers anywhere in the repo) and a **~55ms `View()`** from chained `overlayAt` calls re-measuring the whole screen. Different subsystem, separate PR.

https://claude.ai/code/session_01LJJZE3XYFxMqDazpWawrGU

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced sidebar freezes during scrolling and navigation.
  * Improved responsiveness by preventing unnecessary transcript rescans and blocking session checks.
  * Added bounded session checks to prevent prolonged UI stalls.
  * Improved handling of growing, replaced, or partially written transcripts.
* **Reliability**
  * Preview and focus-mode updates now remain responsive when session status is unavailable or changing.

### DIFF
--- a/changelog/unreleased/ui-stall-tmux-and-transcript.md
+++ b/changelog/unreleased/ui-stall-tmux-and-transcript.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**No more half-second freezes** — The sidebar no longer locks up while you scroll or navigate. Reading a session's title used to re-scan its whole Claude transcript every 30s — up to 89MB, now ~15µs — which stalled the background worker and left arrow keys forking `tmux` on the UI thread.
+**No more half-second freezes** — The sidebar no longer locks up while you scroll or navigate sessions with large transcripts.
--- a/changelog/unreleased/ui-stall-tmux-and-transcript.md
+++ b/changelog/unreleased/ui-stall-tmux-and-transcript.md
@@ -1,0 +1,5 @@
+---
+type: fixed
+---
+
+**No more half-second freezes** — The sidebar no longer locks up while you scroll or navigate. Reading a session's title used to re-scan its whole Claude transcript every 30s — up to 89MB, now ~15µs — which stalled the background worker and left arrow keys forking `tmux` on the UI thread.
--- a/internal/session/claude_name.go
+++ b/internal/session/claude_name.go
@@ -2,6 +2,7 @@ package session
 
 import (
 	"bufio"
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -92,17 +93,19 @@ func ReadClaudeSessionName(claudeSessionID, projectPath string) string {
 	}
 
 	sc := titleScans[jsonlPath]
-	switch {
-	case sc == nil || st.Size() < sc.offset:
-		// Unseen, or the file shrank — a truncation or a reused session id means
-		// the memoised titles describe content that is no longer there.
-		sc = &titleScan{}
-		titleScans[jsonlPath] = sc
-	case st.Size() == sc.size && st.ModTime().Equal(sc.mtime):
+	if sc != nil && st.Size() == sc.size && st.ModTime().Equal(sc.mtime) {
 		// Untouched since the last read, so the memo is the whole answer and the
 		// file is never opened. This is the common case: the worker rechecks every
 		// session every 30s, and most transcripts are idle.
 		return sc.title()
+	}
+	if sc == nil || !sc.resumable(jsonlPath, st) {
+		// Start over, titles included. Reset is deliberately the only way to reach
+		// a scan from offset 0: custom and ai accumulate across calls, so a path
+		// that rewound the offset without clearing them would keep serving a title
+		// the current file no longer contains.
+		sc = &titleScan{}
+		titleScans[jsonlPath] = sc
 	}
 
 	offset, scanErr := scanTranscriptFrom(jsonlPath, sc.offset, func(line string) bool {
@@ -133,15 +136,18 @@ func ReadClaudeSessionName(claudeSessionID, projectPath string) string {
 		return true
 	})
 
-	sc.offset = offset
-	if scanErr == nil {
-		sc.size, sc.mtime = st.Size(), st.ModTime()
-	} else {
-		// A failed read leaves the tail unseen. Clearing the stat keys forces the
-		// next call past the untouched-file shortcut so it retries from offset,
-		// rather than serving a memo that is missing entries it never reached.
-		sc.size, sc.mtime = 0, time.Time{}
+	if scanErr != nil {
+		// The tail went unread, so custom and ai describe a prefix rather than the
+		// file. Returning them would be a title derived from content this call
+		// just failed to read — so drop the memo and answer nothing. That costs
+		// no visible title: the caller only assigns a non-empty name, and keeps
+		// its last good one until a read succeeds.
+		*sc = titleScan{}
+		return ""
 	}
+	sc.offset = offset
+	sc.size, sc.mtime = st.Size(), st.ModTime()
+	sc.anchor = transcriptAnchor(jsonlPath, offset)
 	return sc.title()
 }
 
@@ -156,8 +162,56 @@ type titleScan struct {
 	size   int64
 	mtime  time.Time
 	offset int64
+	anchor []byte // the bytes ending at offset; see resumable
 	custom string
 	ai     string
+}
+
+// resumable reports whether the file on disk is still the one this memo walked,
+// merely longer — the only condition under which offset, custom and ai still
+// describe it.
+//
+// Size cannot answer that on its own. A transcript replaced under the same path
+// by content at least as long looks like growth, so the scan would seek into
+// unrelated bytes and the previous file's custom-title would survive with no
+// later append able to clear it — a state the full walk this replaced could not
+// reach, because it recomputed both titles from scratch every call. Inode
+// identity misses the same case when the replacement is written in place. So the
+// check is the bytes themselves: whatever sat immediately before the resume
+// point must still sit there.
+func (t *titleScan) resumable(path string, st os.FileInfo) bool {
+	if t.offset == 0 || st.Size() < t.offset {
+		return false
+	}
+	return bytes.Equal(transcriptAnchor(path, t.offset), t.anchor)
+}
+
+// transcriptAnchorLen is how many bytes ending at a resume point are kept as
+// proof the prefix is unchanged. Long enough to be unique across JSONL entries,
+// short enough that verifying it is one small read.
+const transcriptAnchorLen = 64
+
+// transcriptAnchor returns the bytes ending at offset, or nil if they cannot be
+// read. Used both to record a resume point and to verify one.
+func transcriptAnchor(path string, offset int64) []byte {
+	n := int64(transcriptAnchorLen)
+	if offset < n {
+		n = offset
+	}
+	if n <= 0 {
+		return nil
+	}
+	f, err := os.Open(path)
+	if err != nil {
+		return nil
+	}
+	defer f.Close()
+
+	buf := make([]byte, n)
+	if _, err := f.ReadAt(buf, offset-n); err != nil {
+		return nil
+	}
+	return buf
 }
 
 // title applies the same precedence the full walk did: an explicit /rename wins
@@ -528,7 +582,13 @@ func scanTranscriptFrom(path string, from int64, fn func(line string) bool) (int
 		if !overCap {
 			buf = append(buf, chunk...)
 			if line := strings.TrimRight(string(buf), "\r\n"); line != "" && !fn(line) {
-				return consumed + pending, nil
+				if err == nil {
+					return consumed + pending, nil
+				}
+				// Stopped on the trailing partial line: fn has seen it, but
+				// consuming it would break the promise in the doc comment above
+				// and silently skip it once the writer finishes the line.
+				return consumed, nil
 			}
 		}
 		buf, overCap = buf[:0], false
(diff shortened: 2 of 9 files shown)


internal/session/claude_name.go
```diff
@@ -77,8 +77,35 @@ func ReadClaudeSessionName(claudeSessionID, projectPath string) string {
...
+		// "this session has no title" and must not pass in silence.
+		if !errors.Is(err, os.ErrNotExist) {
+			warnTranscript("transcript: stat failed", jsonlPath, "err", err)
+		}
+		return ""
+	}
+
+	sc := titleScans[jsonlPath]
+	switch {
+	case sc == nil || st.Size() < sc.offset:
```
hayke102: The reset only fires when the file shrinks below the recorded offset, so a transcript **replaced** under the same path by content that is at least as long takes the resume branch instead: the scan seeks into the middle of unrelated content, and `sc.custom`/`sc.ai` from the old file survive.

`sc.custom` is never cleared by any path, so if the old content carried a `custom-title` and the new content doesn't, `title()` keeps returning the dead session's `/rename` title indefinitely — no later append can clear it. The full walk this replaces recomputed both locals from scratch each time and couldn't get into that state.

A reused session id or a restored transcript are the realistic ways in. Cheapest fix is to key the memo on inode (or on `(path, size-at-first-sight)`) rather than path alone; adding `st.Size() < sc.size` to this case narrows the window but doesn't close the grew-in-place one.

internal/session/claude_name.go
```diff
@@ -96,22 +123,60 @@ func ReadClaudeSessionName(claudeSessionID, projectPath string) string {
...
+				sc.ai = entry.AITitle
 			}
 		}
 		return true
 	})
 
-	if lastCustom != "" {
-		return lastCustom
+	sc.offset = offset
+	if scanErr == nil {
```
hayke102: Same class as the reset gap above: the stat keys are zeroed to force a retry, but `sc.custom`/`sc.ai` are left intact and `sc.title()` is returned on this very call — a title derived from content the function just failed to re-read.

It's worst on a seek failure, where `offset` also goes back to 0 while the stale titles stay. Clearing `custom`/`ai` alongside the stat keys would make the failure honest, at the cost of one cycle with no title.

internal/tmux/tmux.go
```diff
@@ -641,20 +646,35 @@ func (s *Session) PaneDeadInfo() (dead bool, exitStatus, exitSignal string, ok b
...
 		return exists
 	}
-	sessionCacheMu.RUnlock()
 
-	// Fallback to tmux has-session.
-	cmd := exec.Command("tmux", "has-session", "-t", s.Name)
-	return cmd.Run() == nil
+	ctx, cancel := context.WithTimeout(context.Background(), hasSessionTimeout)
+	defer cancel()
+	return exec.CommandContext(ctx, "tmux", "has-session", "-t", s.Name).Run() == nil
```
hayke102: Bounding this is clearly better than the unbounded fork, but the timeout branch is currently indistinguishable from *absent*: a server slow enough to take >2s makes `Exists()` return `false`, which every caller reads as "the session is gone".

Concretely, on a loaded server: `restartSession` (`app.go:4494`) takes the full-rebuild path instead of `RespawnClaude`, `cmd/fleet/send.go:215` refuses a live session with "no tmux", and `finalizeDelete` skips the `Kill()` and leaks the tmux session. And the load case is exactly the one this PR describes — the worker hammering the server is why the cache went stale in the first place.

Worth at least a `debuglog` line on the timeout so it isn't silent; falling back to the last cached value (stale but known) would be stronger.

internal/session/claude_name.go
```diff
@@ -402,6 +467,19 @@ const transcriptLineCap = 8 << 20 // 8MB
...
+	return err
+}
+
+// scanTranscriptFrom is forEachTranscriptLine starting at a byte offset. It
+// returns the offset just past the last *complete* line it consumed, so a
+// caller can resume there when the file grows.
+//
+// A trailing chunk with no newline is a line still being written: fn sees it
+// (matching the whole-file walk, where a caller reading the last entry must not
+// miss one mid-write), but it is not counted as consumed — otherwise a resuming
```
hayke102, line 534: This says the trailing chunk with no newline "is not counted as consumed", but the early-stop return at the bottom of the loop hands back `consumed + pending`, which does consume it.

No live impact today — `ReadClaudeSessionName`'s `fn` always returns `true`, and the other `forEachTranscriptLine` callers discard the offset — so it's the comment that's load-bearing here: the next caller that both stops early *and* resumes will silently skip the half-written line this promises to re-read.

changelog/unreleased/ui-stall-tmux-and-transcript.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**No more half-second freezes** — The sidebar no longer locks up while you scroll or navigate. Reading a session's title used to re-scan its whole Claude transcript every 30s — up to 89MB, now ~15µs — which stalled the background worker and left arrow keys forking `tmux` on the UI thread.
```
hayke102: Per `changelog/README.md`, fragments are 1–2 sentences, user-facing, no implementation detail. The worker, the 30s recheck, the 89MB scan, the µs figure and "forking `tmux` on the UI thread" are all implementation — and this renders in-app under `Ctrl+K` → Release Notes, where it reads as a commit body.

The headline plus the first sentence are already the whole user-facing change; the rest belongs in the commit message (where it already is).

internal/session/claude_name.go
hayke102: Fixed. Size can't answer this and inode identity misses replacement written in place, so the memo now keeps a 64-byte anchor of the bytes ending at its resume point and re-reads them to prove the prefix is unchanged. Reset is a single path that clears custom/ai as well as the offsets, so no field can outlive the content it came from.

Worth naming: this was a regression the memo introduced, not a pre-existing gap — the full walk recomputed both locals every call and couldn't reach that state. TestReadClaudeSessionNameIncrementalMatchesFullScan now covers the grew-in-place case, and reverting the gate reproduces exactly the symptom you described (returns the dead session's title).

internal/session/claude_name.go
hayke102: Fixed — clearing custom/ai alongside the stat keys, and returning "" rather than sc.title() on this call.

The cycle-with-no-title you priced in doesn't actually materialise: app.go:6346 only assigns when name != "", and the heuristic fallback is gated on AgentSessionName != "". So the session keeps its last good title until a read succeeds, and the honest failure is free.

internal/tmux/tmux.go
hayke102: Fixed, taking the stronger option. Exists now branches on ctx.Err(): only a probe that actually completed may report absence, and a timeout answers from the cache ignoring its TTL, logging that it did so.

Your point about the load case is what settled it — the timeout fires hardest under exactly the state this PR describes, so the false negative would have landed on restartSession, fleet send and finalizeDelete precisely when the server was busy.

internal/session/claude_name.go
hayke102: Fixed the code rather than the comment — the invariant it states is the one worth keeping. Early stop now returns consumed+pending only when the line was newline-terminated, and consumed when it was the partial trailing chunk.

Agreed it's latent today; TestScanTranscriptFromEarlyStopKeepsPartialLine pins both branches so the next caller that stops early and resumes gets the behaviour the comment advertises.

changelog/unreleased/ui-stall-tmux-and-transcript.md
hayke102: Fixed — trimmed to the headline plus the first sentence, with the detail left in the commit message where it already was.

Kept the bold headline: CodeRabbit's parallel comment suggested dropping it, but changelog/README.md requires one, so I took your version rather than its literal diff.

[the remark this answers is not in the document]
hayke102: Fixed — the fragment is now one user-facing sentence.

Took this in spirit rather than literally: the suggested replacement also drops the bold headline, and changelog/README.md requires each fragment to open with one, so the result is `**No more half-second freezes** — The sidebar no longer locks up while you scroll or navigate sessions with large transcripts.`

internal/session/claude_name.go
```diff
@@ -77,8 +77,35 @@ func ReadClaudeSessionName(claudeSessionID, projectPath string) string {
...
+		sc = &titleScan{}
+		titleScans[jsonlPath] = sc
+	case st.Size() == sc.size && st.ModTime().Equal(sc.mtime):
+		// Untouched since the last read, so the memo is the whole answer and the
+		// file is never opened. This is the common case: the worker rechecks every
+		// session every 30s, and most transcripts are idle.
+		return sc.title()
+	}
+
+	offset, scanErr := scanTranscriptFrom(jsonlPath, sc.offset, func(line string) bool {
```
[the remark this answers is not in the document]
hayke102: Fixed. The memo now proves append-continuity with a 64-byte anchor of the bytes ending at the resume point rather than inferring it from size, and reset clears the cached titles as well as the offset.

Extended the replacement test as you asked — it now covers a rewritten file longer than the previous transcript, which is the case size-only detection missed.

internal/ui/app.go
```diff
@@ -5321,7 +5338,14 @@ func (h *Home) focusTick() tea.Cmd {
...
 	s := h.selectedSession()
-	if s == nil || !s.IsAlive() {
+	if s == nil {
+		h.focusMode = false
+		h.sidebarDirty = true
+		return h, nil
+	}
+	// Per-keystroke, so this is typing latency: same cache-only rule as
+	// focusTick, and the same reason not to eject on an unknown.
+	if alive, known := s.IsAliveCached(); known && !alive {
```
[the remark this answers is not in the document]
hayke102: Fixed — enterFocusMode now uses IsAliveCached, matching focusTick and handleFocusKey.

A known-dead session still refuses with the existing error, so the cached path is unchanged; only an unknown enters optimistically, and the next known-dead reading ejects. Attach, quick-approve, context menu and paste stay authoritative on purpose — they're one-off actions where a wrong answer is more visible than a brief wait, and they're now bounded by the has-session timeout added in this round. PR description updated, since it previously said focus entry was in that group.

PR #267, opened 2026-08-23
fix(status): keep a session waiting on a single-question AskUserQuestion

hayke102:
A session blocked on an unanswered `AskUserQuestion` dialog reported **Running**, and flapped between Running and Waiting as the user arrowed between options. Diagnosed from a `D` snapshot: hook said `waiting`, `pane_detected` was empty, TUI showed `running`.

Two independent signals had to fail together, and both did.

## 1. The pane went silent

The dialog renders a **side-by-side preview panel** next to the options. That panel is ~20 lines tall, so the cursor line sits well above the window `detectWaiting` scans:

```
 0 | Enter to select · ↑/↓ to navigate · n to add notes · Esc to cancel   ← footer ✓
13 |   4. Trim, and keep the           │ └────────┘                       ← other option ✓
19 | ❯ 1. Trim at the payload          ┌────────────                      ← cursor ✗ (bottomN = 15)
```

`hasMenuFooter` and `hasOtherOption` were true, `hasCursorOnOption` false → no match. The dedicated AskUserQuestion check didn't rescue it either: it required `Tab to switch questions`, and a **single-question** dialog omits that hint because there is nothing to switch to.

`detectWaiting` now accepts `n to add notes` **or** `Tab to switch questions`, each paired with `Esc to cancel` on the same line.

## 2. The transcript tiebreaker voted, and was wrong

With the pane silent, `conversationActivePastHook` decided. Its newest "lead" entry:

```json
{"type":"queue-operation","operation":"enqueue","timestamp":"2026-08-23T14:18:57.857Z",
 "content":"<task-notification>… Agent \"Commit and push BRZ-3287\" finished …"}
```

A background agent's completion notification being **enqueued** while the lead sat blocked. It fires *because* the lead is parked — the exact inverse of the tiebreaker's stated premise, *"the LEAD-only transcript stays static while the lead is truly blocked."* Last genuine lead activity was 62s **before** the hook.

`scanTranscriptTimestamp` accepted any timestamped line and filtered only `isSidechain: true`; `queue-operation` entries carry `isSidechain: null`. 4.9s past the hook (> 2s skew) and 40s old (< 120s window) → `convActive = true` → running, pinned for the whole window.

The lead scan now allowlists the conversation types (`assistant` / `user` / `attachment`). An allowlist rather than a denylist because Claude Code grows bookkeeping types far more often than conversation types, so an unknown type is likelier noise than progress — and a resumed lead must emit `assistant` or `user`, so there is no resumption it misses. `lastTranscriptTimestamp`, the rotation-detection signal where *any* write is the point, stays unfiltered.

## The flapping

Same bug from the other side. At 17:20:24–28 the log reads `matched permission menu structure`; at 17:20:29, `no pattern matched`. The status tracked **which option was highlighted** — a shorter preview panel pulled `❯ N.` back inside the window, a taller one pushed it out.

## Not done here

Widening `bottomN` for the option lines (scan further up for the cursor while keeping the footer anchored to the bottom 15) would also have caught this. Left out: the footer path covers this dialog structurally, and loosening the menu gate deserves its own decision.

## Tests

Each fails without its fix:

- `TestGoldenDetection/pane_waiting_askuserquestion_single_preview.txt` — sanitized fixture reproducing the real geometry byte-for-byte (`❯ 1.` at index 19, option at 13, footer at 0, no Tab hint). The menu check still misses it, as it did live; the footer check catches it.
- `TestLastLeadTranscriptTimestamp/skips_bookkeeping_entries…` — also asserts `lastTranscriptTimestamp` still sees every entry, so the filter can't leak into rotation detection.
- `TestConversationActivePastHookIgnoresQueuedWork` — end-to-end over a real temp transcript, with a positive control (a genuine resumed `assistant` entry must still return true) so the negative cases can't pass on a scanner that simply found nothing.

`make lint` clean. `TestCredentialResolutionOrder` fails on master too (it reads the developer's real Linear keychain entry) — unrelated.

## Loose end

The *first* flip happened 722ms after the hook, before the `queue-operation` existed, so a third path put it into running initially. `debug.log` had rotated past that window. Most likely the content-hash branch (`content changed while waiting, assuming running`) firing on a stale `lastContentHash` from the prior waiting episode. Worth revisiting if it recurs at sub-second hook age.

https://claude.ai/code/session_01LJJZE3XYFxMqDazpWawrGU

## Summary by CodeRabbit

* **Bug Fixes**
  * Single-question prompts now remain in the waiting state while navigating options, without incorrectly switching to running.
  * Improved detection of question prompts even when options or cursor indicators are not visible.
  * Background notifications, hooks, and sidechain activity no longer incorrectly appear as active conversation progress.

### DIFF
--- a/internal/session/claude_name.go
+++ b/internal/session/claude_name.go
@@ -657,11 +657,10 @@ func lastLeadTranscriptTimestamp(path string) time.Time {
 }
 
 // leadConversationTypes are the JSONL entry types that prove the LEAD turn advanced:
-// the agent spoke or called a tool ("assistant"), a tool result or user message landed
-// ("user"), or the attachment that follows one did ("attachment"). Only these count for
-// conversationActivePastHook. Everything else in a transcript is bookkeeping that fires
-// while the lead is parked, and reading it as progress flips a genuinely blocked session
-// to running:
+// the agent spoke or called a tool ("assistant"), or a tool result or user message landed
+// ("user"). Only these count for conversationActivePastHook. Everything else in a
+// transcript is bookkeeping that fires while the lead is parked, and reading it as
+// progress flips a genuinely blocked session to running:
 //
 //   - "queue-operation" is something arriving AT a blocked lead — a message the user
 //     typed while a prompt is on screen, or a background agent's task-notification being
@@ -671,21 +670,34 @@ func lastLeadTranscriptTimestamp(path string) time.Time {
 //   - "system" (subtypes "stop_hook_summary", "turn_duration") is written as a turn ENDS.
 //   - "file-history-delta" is an edit snapshot, always paired with the "assistant"
 //     tool_use that caused it, so it carries no signal of its own.
+//   - "attachment" reads like part of the conversation and is not. Measured across the 39
+//     most recent local transcripts (8,474 attachments): 6,988 are "total_tokens_reminder",
+//     390 "task_reminder", and 174 "queued_command" — a message typed at a PARKED lead,
+//     structurally the same class as "queue-operation" above. Including it bought nothing
+//     to pay for that: it pushed the last-entry time past "assistant"/"user" in 3 of those
+//     39 transcripts, by at most 3 milliseconds.
 //
 // An allowlist rather than a denylist because Claude Code grows bookkeeping types far
 // more often than conversation types, so an unknown type is far likelier to be noise than
 // progress — and a resumed lead must emit "assistant" or "user", so there is no
-// resumption these miss.
+// resumption these two miss.
 var leadConversationTypes = map[string]bool{
-	"assistant":  true,
-	"user":       true,
-	"attachment": true,
+	"assistant": true,
+	"user":      true,
 }
 
-// scanTranscriptTimestamp walks the JSONL transcript at path and returns the first
-// (last=false) or last (last=true) entry timestamp it can parse. When leadOnly is set,
+// scanTranscriptTimestamp walks the JSONL transcript at path and returns the earliest
+// (last=false) or latest (last=true) entry timestamp it can parse. When leadOnly is set,
 // sub-agent (isSidechain:true) entries and bookkeeping entries are skipped, leaving only
 // the lead conversation (see leadConversationTypes).
+//
+// "latest" means the MAXIMUM, not the last line: transcripts are not written in timestamp
+// order. A message typed while the lead is parked is stamped when it is QUEUED and
+// appended when the lead RESUMES, so it lands out of order carrying a past timestamp —
+// measured backwards by up to 162s among "assistant"/"user" alone, in 28 of the 39 most
+// recent local transcripts. Overwriting on every match let one such entry understate how
+// far the conversation had got, which is the direction that demotes a live lead to
+// finished on exactly the between-bursts frame conversationActivePastHook exists to cover.
 func scanTranscriptTimestamp(path string, last bool, leadOnly bool) time.Time {
 	if path == "" {
 		return time.Time{}
@@ -712,8 +724,14 @@ func scanTranscriptTimestamp(path string, last bool, leadOnly bool) time.Time {
 		if err != nil {
 			return true
 		}
-		result = ts
-		return last // first match wins when we only want the earliest entry
+		if !last {
+			result = ts
+			return false // first match wins when we only want the earliest entry
+		}
+		if ts.After(result) {
+			result = ts
+		}
+		return true
 	})
 	return result
 }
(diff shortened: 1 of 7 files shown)


internal/session/claude_name.go
```diff
@@ -670,11 +698,12 @@ func scanTranscriptTimestamp(path string, last bool, excludeSidechain bool) time
 		var entry struct {
 			Timestamp   string `json:"timestamp"`
 			IsSidechain bool   `json:"isSidechain"`
+			Type        string `json:"type"`
 		}
 		if err := json.Unmarshal([]byte(line), &entry); err != nil || entry.Timestamp == "" {
 			return true
 		}
-		if excludeSidechain && entry.IsSidechain {
+		if leadOnly && (entry.IsSidechain || !leadConversationTypes[entry.Type]) {
```
hayke102, line 718: `scanTranscriptTimestamp(last=true)` takes the **last matching line's** timestamp (`result = ts`, ~line 715 just below), not the maximum — and real transcripts aren't monotonic.

From a live transcript: an `attachment`/`queued_command` entry at index 170 is stamped `14:30:20.076`, sitting *after* a `user` entry at index 169 stamped `14:30:26.239`. `lastLeadTranscriptTimestamp` therefore reports ~6s earlier than reality, which can make `conversationActivePastHook` read false on a genuinely-running lead and demote it to finished on a between-bursts frame — the exact frame this tiebreaker exists to cover.

Pre-existing, but the new allowlist narrows the surviving types to three, one of which (`attachment`) is the out-of-order one, so it's more exposed now. Keeping the max rather than overwriting on every match would close it.

internal/session/claude_name.go
```diff
@@ -646,19 +646,47 @@ func lastTranscriptTimestamp(path string) time.Time {
...
+//     tool_use that caused it, so it carries no signal of its own.
+//
+// An allowlist rather than a denylist because Claude Code grows bookkeeping types far
+// more often than conversation types, so an unknown type is far likelier to be noise than
+// progress — and a resumed lead must emit "assistant" or "user", so there is no
+// resumption these miss.
+var leadConversationTypes = map[string]bool{
+	"assistant":  true,
+	"user":       true,
+	"attachment": true,
```
hayke102: The rationale above — "the attachment that follows one" — asserts a property `attachment` doesn't actually have. Real subtypes include `queued_command` (a message typed at a **parked** lead, structurally the same class `queue-operation` is excluded for) and `total_tokens_reminder` (59 of 69 attachments in one sampled transcript).

No live bug today: the sampled `queued_command` attachment carried the *queue-time* stamp and was flushed at resume, so it doesn't reintroduce the flip. But a Claude Code that stamps it at enqueue time instead would — silently, with this comment vouching that it can't. Worth narrowing to the subtype, or at least correcting the comment so the next reader doesn't trust it.

internal/session/claude_name.go
hayke102: Fixed in f76844d — `scanTranscriptTimestamp` now keeps the maximum when `last` is set.

Confirmed, and broader than the sample suggested. Over the 39 most recent local transcripts, inversions among the allowlisted types hit 38/39 files (largest 1661s); restricted to `assistant`/`user` alone — i.e. after the `attachment` removal in the sibling thread — still 28/39, largest 162s.

Worth recording that today's exposure is latent rather than live: `last != max` at the *file tail* in 0/39 files, because a flushed queued entry is immediately followed by the real `user`/`assistant` ones. But the fast pass reads every 500ms, the direction of error is the harmful one (understating freshness demotes a live lead), and `max` is ordering-independent going forward.

`firstTranscriptTimestamp` (`last=false`) has the mirror-image issue and still stops at the first match — different caller (rotation detection), left alone deliberately and pinned by the new test.

internal/session/claude_name.go
hayke102: Fixed in f76844d — dropped `attachment` from the allowlist rather than correcting the comment.

The census backs you up. Across the same 39 transcripts (8,474 attachments): `total_tokens_reminder` 6,988, `task_reminder` 390, `queued_command` 174, `edited_text_file` 130, `command_permissions` 109, plus 18 more bookkeeping subtypes.

What settled it was measuring what the type actually bought: including `attachment` pushed the last-entry time past `assistant`/`user` in **3 of 39** transcripts, by at most **3 milliseconds**. So it was carrying `queued_command` — the same parked-lead class `queue-operation` is excluded for — in exchange for nothing.

Removing it beats narrowing to the subtype: it deletes the wrong rationale instead of rewriting it, needs no nested-subtype parse, and leaves the allowlist saying exactly what it claims — a resumed lead must emit `assistant` or `user`. The measurement is now in the comment so the next reader gets the number, not a claim.

Covered by a `queued_command` case in `TestConversationActivePastHookIgnoresQueuedWork` (which keeps its positive control, so the negatives can't pass on a scanner that stopped finding anything).

PR #268, opened 2026-08-23
fix(ui): stop mouse reporting pinning the CPU on scroll

hayke102:
## What

`chrome()` set `MouseModeCellMotion` on every frame, putting the terminal into DECSET 1002. **Nothing in the TUI has ever handled a mouse event** — that line was the only `Mouse` reference in the entire non-test tree:

```
internal/ui/app.go:2231   v.MouseMode = tea.MouseModeCellMotion
internal/diagnostics/…    (only reports tmux's own mouse setting)
```

It was carried forward from v1's `NewProgram` options during the v2 migration (#153); a handler was never written.

## Why it hurt

Bubble Tea v2's `eventLoop` calls `p.render(model)` **unconditionally** after every message — there's no dirty check (`bubbletea/v2/tea.go:888`). So every ignored `MouseWheelMsg` still bought a full `View()`.

A trackpad scroll over the window produced ~300 msg/sec. `Update()` itself is free (**83ns** — the message is ignored), but the render is not:

```
18:07:10.375  tea.MouseWheelMsg    83ns
18:07:10.378  tea.MouseWheelMsg    83ns
...  (31 of 32 ring slots, in all 8 dumps)
```

Measured from `perfwatch` heartbeats: +287 msg/sec for +123 points of CPU ≈ **4.3ms of CPU per frame**, sustaining **220% CPU** (2.2 cores) for 45s, with nothing on screen changing. `updates_slow=1` for the whole 37k-update run — nothing blocked; it was pure repaint volume.

Seven goroutine samples across the storm caught the main goroutine mid-render every time, never in Update logic and never on I/O:

| Samples | Location |
|---|---|
| 4/7 | `ansi.StringWidth` → `FirstGraphemeCluster` → uax29 grapheme scan, via `lipgloss.JoinHorizontal` |
| 1/7 | `lipgloss.Style.Render` in `renderBorderedPanel` |
| 2/7 | `ultraviolet` cell diff / `Layer.Draw` |

Not a one-off: **115 storm dumps over ~20 days**, 8 in a single hour on the day this was diagnosed.

## The change

One line — `v.MouseMode = tea.MouseModeNone` — plus a comment recording why, so it doesn't get "restored" later.

Turning reporting off also hands **drag-select** back to the terminal, which cell-motion had been taking (selection needed a modifier) with nothing to show for it. Not scrollback — fleet stays in `AltScreen`, which has none.

## Follow-up from review (e08f111)

@hayke102 caught that `MouseModeNone` doesn't remove the wheel events, it *re-sources* them: with the alternate screen up, a terminal with DECSET 1007 on synthesizes arrow keys from the wheel. Those land on `case "j", "down"` (cursor walk + an unthrottled preview fork per tick) and, with the drawer focused, on `forwardKeyToPane` → tmux `Up`/`Down`, walking the shell's readline history.

So the commit also clears mode 1007, in `internal/termkeys` — which already writes raw mode escapes as literals and already has a `Restore()` wired on exit, so no new package and no new call site. Saved with XTSAVE and put back with XTRESTORE rather than re-enabled with a DECSET, since the terminal's prior value is unknowable from here and a blind DECSET would turn the mode *on* for anyone who started with it off. A test pins that.

**Not yet measured** — the one thing the reviewer actually asked for. Pending, and the thread is deliberately left open until it lands: (1) does the terminal translate wheel→arrows at all; (2) does entering the alt screen reset 1007, given `termkeys.Disable` runs before `p.Run()`; (3) does it survive an attach round trip, since tmux sets its own modes.

## Known limitation

This removes the *source*, not the *amplifier*. v2 still re-renders after every message and `View()` still costs ~4.3ms, so any future high-rate message source reproduces the same shape. The durable fix is memoizing `View()` behind a dirty flag — deliberately out of scope here.

## Test plan

- [x] `make build`
- [ ] Run with `FLEET_DEBUG=1`, scroll hard over the TUI for ~10s, confirm no new `update_storm` dump in `~/.config/fleet/stalls/`
- [ ] Confirm drag-to-select works in the fleet window without a modifier

https://claude.ai/code/session_01SjtVGXqcpwvUXmGCzWVUrk

### DIFF
--- a/changelog/unreleased/mouse-repaint-storm.md
+++ b/changelog/unreleased/mouse-repaint-storm.md
@@ -1,0 +1,5 @@
+---
+type: fixed
+---
+
+**Scrolling no longer pegs the CPU.** fleet put your terminal into mouse-reporting mode but never handled a mouse event, so a trackpad scroll over the TUI fired ~300 ignored redraws a second and held fleet at 220% CPU until you stopped. Drag-select and scrollback now behave like they do in any other terminal app.
--- a/changelog/unreleased/mouse-repaint-storm.md
+++ b/changelog/unreleased/mouse-repaint-storm.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**Scrolling no longer pegs the CPU.** fleet put your terminal into mouse-reporting mode but never handled a mouse event, so a trackpad scroll over the TUI fired ~300 ignored redraws a second and held fleet at 220% CPU until you stopped. Drag-select and scrollback now behave like they do in any other terminal app.
+**Scrolling no longer pegs the CPU.** fleet put your terminal into mouse-reporting mode but never handled a mouse event, so a trackpad scroll over the TUI fired ~300 ignored redraws a second and held fleet at 220% CPU until you stopped. Drag-select now works without holding a modifier.
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -2233,7 +2233,10 @@ func (h *Home) chrome(content string) tea.View {
 	// with cell-motion on, one trackpad scroll over the window became ~300 ignored
 	// MouseWheelMsg/sec, each buying a full View() — measured at 220% CPU for as
 	// long as the finger moved, with nothing on screen changing. Off also hands
-	// drag-select and scrollback back to the terminal.
+	// drag-select back to the terminal — not scrollback, which the alternate
+	// screen does not have. Reporting off is only half of it: internal/termkeys
+	// clears DECSET 1007 alongside, or the terminal would synthesize arrow keys
+	// from the wheel instead and the events would land as KeyPressMsg.
 	v.MouseMode = tea.MouseModeNone
 	return v
 }
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -2228,7 +2228,13 @@ func (h *Home) View() tea.View {
 func (h *Home) chrome(content string) tea.View {
 	v := tea.NewView(content)
 	v.AltScreen = true
-	v.MouseMode = tea.MouseModeCellMotion
+	// MouseModeNone, deliberately: nothing in the TUI handles a mouse event, and
+	// reporting is not free. v2's eventLoop re-renders after *every* message, so
+	// with cell-motion on, one trackpad scroll over the window became ~300 ignored
+	// MouseWheelMsg/sec, each buying a full View() — measured at 220% CPU for as
+	// long as the finger moved, with nothing on screen changing. Off also hands
+	// drag-select and scrollback back to the terminal.
+	v.MouseMode = tea.MouseModeNone
 	return v
 }
 
--- a/internal/termkeys/termkeys.go
+++ b/internal/termkeys/termkeys.go
@@ -13,6 +13,16 @@
 //
 // Asking the terminal for legacy key reporting forces Ctrl+K back to 0x0b, so
 // the palette opens reliably. Alt+<digit> slot bindings benefit the same way.
+//
+// It also turns off xterm's alternate-screen scroll (DECSET 1007), which is a
+// key-reporting mode in the same sense: while the alternate screen is up and
+// mouse reporting is off, a terminal with 1007 on *synthesizes arrow keys from
+// the wheel*. fleet asks for no mouse reporting (see Home.chrome) and stays in
+// the alternate screen, which is exactly that combination — so without this a
+// trackpad scroll would arrive as a burst of Up/Down KeyPressMsg: walking the
+// sidebar cursor and forking a preview capture per tick, and, with the drawer
+// focused, walking the shell's readline history (arrows are not intercepted by
+// handleTypingKey, so they reach forwardKeyToPane).
 package termkeys
 
 import (
@@ -28,12 +38,27 @@ import (
 // instead of forcing off), so we keep the literal here.
 const disableModifyOtherKeys = "\x1b[>4;0m"
 
+// Alternate-screen scroll (DECSET 1007) is saved and reset rather than blindly
+// re-enabled on exit: the terminal's prior value is unknowable from here, so a
+// DECSET in Restore would be wrong for anyone who started with it off. XTSAVE
+// (Ps=s) / XTRESTORE (Ps=r) are the paired form xterm defines for exactly this.
+// Literals for the same reason disableModifyOtherKeys is one — x/ansi carries no
+// constant for 1007.
+const (
+	saveAltScreenScroll    = "\x1b[?1007s"
+	disableAltScreenScroll = "\x1b[?1007l"
+	restoreAltScreenScroll = "\x1b[?1007r"
+)
+
 // disablePreamble forces legacy key reporting across the two enhancement
 // mechanisms a Ctrl+K could be intercepted by: modifyOtherKeys off, plus a
 // flags=0 entry pushed onto the Kitty keyboard protocol stack
 // (ansi.DisableKittyKeyboard) so CSI-u reporting is suppressed while preserving
-// whatever state the shell/tmux had. Terminals without a mode ignore its bytes.
-const disablePreamble = disableModifyOtherKeys + ansi.DisableKittyKeyboard
+// whatever state the shell/tmux had. It then saves and clears alternate-screen
+// scroll, so the wheel stops being turned into arrow keys. Terminals without a
+// mode ignore its bytes.
+const disablePreamble = disableModifyOtherKeys + ansi.DisableKittyKeyboard +
+	saveAltScreenScroll + disableAltScreenScroll
 
 // Disable asks the terminal for legacy key reporting so modified keys (Ctrl+K
 // in particular) arrive in their legacy encoding instead of as CSI-u sequences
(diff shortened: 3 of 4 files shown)


internal/ui/app.go
```diff
@@ -2228,7 +2228,13 @@ func (h *Home) View() tea.View {
...
 	v := tea.NewView(content)
 	v.AltScreen = true
-	v.MouseMode = tea.MouseModeCellMotion
+	// MouseModeNone, deliberately: nothing in the TUI handles a mouse event, and
+	// reporting is not free. v2's eventLoop re-renders after *every* message, so
+	// with cell-motion on, one trackpad scroll over the window became ~300 ignored
+	// MouseWheelMsg/sec, each buying a full View() — measured at 220% CPU for as
+	// long as the finger moved, with nothing on screen changing. Off also hands
+	// drag-select and scrollback back to the terminal.
+	v.MouseMode = tea.MouseModeNone
```
hayke102, line 2240: Turning reporting off doesn't make the wheel events go away — it hands them to the terminal, and with `AltScreen` still on most terminals translate them into arrow keys (xterm `alternateScroll`, default-on in iTerm2, Ghostty, Terminal.app, and tmux). So a trackpad scroll over the window now produces a burst of `up`/`down` `KeyPressMsg` instead of `MouseWheelMsg`.

Two places that lands:

1. **Main list** — each one hits the `"j", "down"` / `"k", "up"` cases (~app.go:2825-2833), moving the sidebar cursor and returning `fetchPreviewForSelected()`. That's deliberately unthrottled ("every j/k lands here", app.go:7231) and forks a `tmux capture-pane` per call. So the message rate this comment blames for the 220% may be re-sourced rather than removed — now with a walking cursor and a subprocess per scroll tick on top, which is the same fork churn #265 was about.

2. **Drawer focused** — sharper, and not rate-dependent. `handleTypingKey` intercepts only its Ctrl chords and `pgup`/`pgdown`, so arrows fall through to `forwardKeyToPane` (drawer.go:601), which maps `KeyUp`/`KeyDown` to tmux `Up`/`Down`. Scrolling the trackpad over a focused drawer now walks the shell's readline history, one control-client round trip per tick. Those wheel events used to be swallowed by bubbletea.

Worth measuring in iTerm2/Ghostty before this ships. If the CPU reproduces as arrow keys, the durable fix is debouncing the preview fetch / coalescing renders rather than the mode change alone.

One constraint for whoever picks this up: v2 has no middle setting. `MouseModeNone` / `CellMotion` / `AllMotion` only, and `CellMotion` already covers click+release+wheel — there's no "wheel off, drag-select on" mode to reach for.

changelog/unreleased/mouse-repaint-storm.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Scrolling no longer pegs the CPU.** fleet put your terminal into mouse-reporting mode but never handled a mouse event, so a trackpad scroll over the TUI fired ~300 ignored redraws a second and held fleet at 220% CPU until you stopped. Drag-select and scrollback now behave like they do in any other terminal app.
```
hayke102: The scrollback half of that last sentence isn't accurate: fleet stays in `AltScreen`, and the alternate screen has no scrollback. Drag-select genuinely does come back, so that part stands.

What a user will actually get on a wheel scroll is arrow-key navigation (per the other thread) — closer to `vim`/`less` than to "any other terminal app", which reads as "I can scroll back through history now". Suggest dropping the scrollback claim and keeping the drag-select one.

hayke102: Fixed in e08f111 — dropped the scrollback claim, kept drag-select.

You're right and it wasn't only the changelog: the same sentence was in the code comment and the PR body, so all three moved together. The fragment now ends "Drag-select now works without holding a modifier."

internal/ui/app.go
hayke102: Good catch — both landing sites check out, and I'd only verified the message *source* went away, not where it went instead.

For the record, tracing them: `case "j", "down"` / `"k", "up"` each move the cursor and return `fetchPreviewForSelected()`, whose own comment says "every j/k lands here" and is deliberately unthrottled. And `handleTypingKey` intercepts only its Ctrl chords plus `pgup`/`pgdown`, so arrows fall through to `forwardKeyToPane` and map to tmux `Up`/`Down` — the drawer case is the sharper one, since it's a functional regression rather than a perf one and doesn't depend on the event rate.

Fixed in e08f111, via a lever I don't think was on the table: fleet can clear mode 1007 itself. That's `internal/termkeys`' existing job — it already writes raw mode escapes (`disableModifyOtherKeys = "\x1b[>4;0m"`, a literal because x/ansi has no constant, same as 1007) and is already wired to a `Restore()` on exit, so this needed no new package and no new call site. With 1007 off the wheel yields neither `MouseWheelMsg` nor arrow keys, and drag-select still works. Saved with XTSAVE / restored with XTRESTORE rather than re-enabled with a DECSET, since the terminal's prior value isn't knowable from here and a blind DECSET would switch the mode *on* for anyone who started with it off; there's a test pinning that asymmetry specifically.

**Leaving this thread open, because your actual ask — measure it — isn't satisfied yet.** I couldn't: the agent shell has no tty, so I can't even issue DECRQM to ask the terminal. Three things the measurement has to settle, and I'd rather name them than let the fix imply they're closed:

1. Whether the terminal translates wheel→arrows at all under AltScreen. Your premise, still unverified from here.
2. Whether entering the alt screen resets 1007. `termkeys.Disable` runs *before* `p.Run()`, and bubbletea enters the alt screen inside `Run` — if the mode is per-screen, the disable is in the wrong place and has to move.
3. Whether it survives an attach round trip. `tea.Exec`/PTY attach hands the terminal to tmux, which sets its own modes; 1007 may come back on when control returns.

There's a probe script for it (alt screen up, mouse off, raw stdin dump, with and without the 1007 clear) — I'll post Ghostty and iTerm2 numbers here before this merges.

Also agreed on the durable fix: v2 renders after *every* message unconditionally, so any high-rate source reproduces this shape. Coalescing renders / debouncing the preview fetch is the real fix and this isn't it — happy to take that as a follow-up.

hayke102:
@coderabbitai review

internal/ui/app.go
hayke102: Measured in Ghostty. **You were right, and it was worse than the thing it replaced.**

Method: alternate screen up, mouse reporting off (`MouseModeNone`'s state), 6s of trackpad scrolling captured raw off the tty. Each run also does a control capture — press Down ~5×, which must record bytes — so an empty scroll result can't be confused with a broken probe. That guard earned its keep: v1 of the script read via `cat &`, a background read on the tty takes SIGTTIN and is suspended, and it reported a clean `0 bytes` for every case. It looked like a passing result.

**Baseline (mouse off, 1007 untouched — the PR as you reviewed it):**

```
wheel bytes:   7976
control bytes: 6289
00000000  1b 5b 42 1b 5b 42 1b 5b  42 1b 5b 42 1b 5b 42 1b  |.[B.[B.[B.[B.[B.|
```

7,976 bytes, and every one of them `1b 5b 42` — `ESC [ B`, Down. That's ~2,650 arrow keys in 6 seconds, **~440/sec**, against the ~300 `MouseWheelMsg`/sec the PR was written to remove. So the first commit didn't reduce the rate, it raised it — and swapped a message that cost only a render for one that also walks the cursor and forks a `capture-pane` per tick.

**With 1007 cleared (e08f111):**

```
wheel bytes:   0
control bytes: 18      <- 6 Down presses, 3 bytes each. Probe is sound.
```

**Ordering also checked**, because my first pass got it wrong. The probe cleared 1007 *after* entering the alternate screen, while fleet clears it *before* — `termkeys.Disable` runs at main.go:192, and bubbletea only enters the alt screen later inside `p.Run()`. If the mode were per-screen the shipped fix would do nothing, and the `--fix` run could not have shown that. Re-ran in fleet's actual order:

```
mode:          --fleet-order (1007 cleared BEFORE alt screen, as fleet does)
wheel bytes:   0
control bytes: 21
```

Mode 1007 is not per-screen; the call site is fine where it is.

Still open, and I'd rather leave the thread open than imply otherwise:

- **iTerm2 unmeasured.** Only Ghostty so far.
- **Attach round trip untested** — `tea.Exec`/PTY attach hands the terminal to tmux, which sets its own modes, so 1007 may come back on when control returns.
- **No clean end-to-end run yet.** The instance I have numbers from lived ~5s, which is the boot splash, not a scroll test.

And your last point stands independently: v2 renders after every message unconditionally, so this only removes today's source. Coalescing renders / debouncing the preview fetch is the actual fix, and I'd take it as a follow-up.

hayke102: Resolving — the premise is measured and the fix is verified in fleet's actual startup order.

Two checks deliberately deferred rather than done, both bounded: iTerm2 is unmeasured (1007 is either off there, making this a no-op, or on, making it a fix — there's no third outcome), and the attach round trip is untested (if tmux hands 1007 back, scrolling after an attach resumes arrows — that's pre-PR behaviour returning, not a new failure). Neither can make things worse than before the PR, since clearing 1007 only ever removes input and XTSAVE/XTRESTORE puts the terminal back on exit.

Follow-up for the durable fix — coalescing renders / debouncing the preview fetch — tracked separately, per your last point.

PR #269, opened 2026-08-23
fix(status): stop prose about a question prompt reading as one

hayke102:
## The bug

A finished session showed **waiting** indefinitely. From the snapshot (`2026-08-23T18-11-15_fix-1-and-2-with-tests`), every layer but one was correct:

| Layer | Says | Correct? |
|---|---|---|
| Hook file (`Stop`) | `finished` | ✅ |
| Conversation log (`advanced_past_hook: false`) | not active | ✅ |
| Worker (274ms cycle) | not stalled | ✅ |
| `detectWaiting` | `waiting` | ❌ |
| `detectFinished` | `finished` | ✅ (never reached) |
| TUI | `waiting` | ❌ |

The pane was at rest — `✻ Worked for 59s`, an empty input box, no menu. What tripped it was the session's own summary of what it had just shipped:

> `- detectWaiting accepts the single-question AskUserQuestion footer (n to add notes + Esc to cancel), so a dialog whose preview panel…`

`detectWaiting` required only that one line contain `"esc to cancel"` **and** `"n to add notes"` as substrings. That sentence has both, at `recentLines[11]` — inside the 15-line window. `detectWaiting` runs before `detectFinished`, so the false positive won, and `applyHookFinished`'s `paneStatus == StatusWaiting` branch overrode a correct hook. `debug.log` shows *"hook says finished but pane shows waiting, overriding"* on every tick.

Reachable since #267 added the `"n to add notes"` branch two minutes before the capture; the prior check required `"tab to switch questions"`, which the prose lacks. The session that shipped the change tripped it by describing it — the self-referential false positive the package already warns about.

## The fix

The real footer renders as a `·`-separated hint list:

```
Enter to select · ↑/↓ to navigate · n to add notes · Esc to cancel
```

So `isAskUserQuestionFooter` matches the hints as **whole fields** of that list rather than substrings of the line. Prose mentioning the hints has no `·` between them, so neither is ever a field. This keeps the function's stated philosophy — structural layout checks over text patterns.

## Verification

- Both new tests **fail without the fix, pass with it** (checked by reverting the predicate).
- Across all 20 existing fixtures: **0 behavior differences** old vs new.
- `make build` OK · `make lint` 0 issues · `go test -race ./internal/session/` passes.
- Pre-existing and unrelated: `TestCredentialResolutionOrder` (`internal/linear`) fails on a clean master too — it reads the real Linear keychain entry.

## Tests

- **Golden** `pane_finished_prose_mentions_askuserquestion_footer.txt` — the actual captured pane, expected `StatusFinished`.
- **Scenario** `TestScenarioProseAboutFooterDoesNotPinWaiting` — pins the `applyHookFinished` override path (finished at t=0 and t=30s), and asserts a genuine `·`-delimited footer at t=60s still registers as waiting, so the fix can't be "solved" by disabling the check.

## Noticed, not fixed

`detectFinished`'s prompt scan tests `strings.HasPrefix(line, "❯ ")` with a normal space, but the live input line is `❯ ` (non-breaking space), so that path is dead against real panes — only `idlePatterns` is carrying it. Out of scope here.

https://claude.ai/code/session_01AYcKP6MB1JNkFPvKszc1d3

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed completed sessions being incorrectly shown as active question dialogs when their text merely mentions prompt-related footer wording.
  - Improved waiting-state detection to recognize only genuine question prompts, preventing sessions from becoming stuck in a waiting state.
  - Preserved correct detection for active prompts with valid footer hints.
  - Improved recognition of valid prompts with varied spacing and formatting.

### DIFF
--- a/changelog/unreleased/prose-footer-false-waiting.md
+++ b/changelog/unreleased/prose-footer-false-waiting.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**Talking about a prompt no longer looks like one.** A finished session whose output mentioned `n to add notes` and `Esc to cancel` in the same sentence was read as a live question dialog and stuck on waiting until the text scrolled away.
+**Finished sessions stay finished.** Your session no longer sticks on waiting when its own output happens to mention `n to add notes` and `Esc to cancel` in one sentence — fleet read that as a live question prompt.
--- a/changelog/unreleased/prose-footer-false-waiting.md
+++ b/changelog/unreleased/prose-footer-false-waiting.md
@@ -1,0 +1,5 @@
+---
+type: fixed
+---
+
+**Talking about a prompt no longer looks like one.** A finished session whose output mentioned `n to add notes` and `Esc to cancel` in the same sentence was read as a live question dialog and stuck on waiting until the text scrolled away.
--- a/changelog/unreleased/prose-footer-false-waiting.md
+++ b/changelog/unreleased/prose-footer-false-waiting.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**Finished sessions stay finished.** Your session no longer sticks on waiting when its own output happens to mention `n to add notes` and `Esc to cancel` in one sentence — fleet read that as a live question prompt.
+**Finished sessions stay finished.** Your session no longer sticks on waiting because its own output quoted a permission prompt — a numbered menu, or `n to add notes` and `Esc to cancel` in a sentence, read as a live prompt until the text scrolled away.
--- a/internal/session/scenario_test.go
+++ b/internal/session/scenario_test.go
@@ -1144,3 +1144,39 @@ func TestScenarioProseAboutFooterDoesNotPinWaiting(t *testing.T) {
 		},
 	})
 }
+
+func TestScenarioProseQuotingAMenuDoesNotPinWaiting(t *testing.T) {
+	// Regression (PR #269 review): the sibling of the prose-footer bug above.
+	// isAskUserQuestionFooter matches its hints as whole fields, but the numbered-menu
+	// check one block up still substring-matched its own footer anywhere in the bottom
+	// 15 lines. A session doing status-detection work prints both halves as ordinary
+	// conversation — a quoted menu and the words "Esc to cancel" — so the same finished
+	// session was pinned to waiting through applyHookFinished's pane override.
+	//
+	// Fix: a footer only counts within footerN (3) of the bottom. Measured across every
+	// pane_waiting_* fixture the real footer sits at index 0-2; prose about one sits
+	// above the input box and mode bar the TUI always renders below it.
+	menuProse := "⏺ The permission dialog renders options like:\n\n" +
+		"  ❯ 1. Yes\n  2. No, tell Claude what to do differently\n\n" +
+		"  The footer (n to add notes + Esc to cancel) is what identifies it.\n\n" +
+		"✻ Worked for 59s\n\n❯ \n  ⏵⏵ auto mode on (shift+tab to cycle)\n"
+
+	// The same menu, actually live: options and footer at the bottom, no chrome under
+	// them. Guards against "fixing" this by disabling the menu check outright.
+	liveMenu := "Read(~/code/foo/bar.ts)\n\nDo you want to proceed?\n\n" +
+		"❯ 1. Yes\n  2. No\n\nEsc to cancel · Tab to amend\n"
+
+	runScenario(t, Scenario{
+		Name: "hook=finished + prose quoting a menu → finished, not waiting",
+		Events: []ScenarioEvent{
+			{At: 0, Hook: "finished", Pane: menuProse},
+			{At: 30 * time.Second, Pane: menuProse},
+			{At: 60 * time.Second, Pane: liveMenu},
+		},
+		Checks: []ScenarioCheck{
+			{At: 0, Expected: StatusFinished},                // before fix: waiting
+			{At: 30 * time.Second, Expected: StatusFinished}, // before fix: waiting
+			{At: 60 * time.Second, Expected: StatusWaiting},  // a real menu still overrides the finished hook
+		},
+	})
+}
--- a/internal/session/scenario_test.go
+++ b/internal/session/scenario_test.go
@@ -1119,9 +1119,15 @@ func TestScenarioProseAboutFooterDoesNotPinWaiting(t *testing.T) {
 		t.Skipf("fixture %s not available", fixture)
 	}
 
-	// The real footer, for contrast: same two hints, but as fields of the "·" list.
-	// A genuine dialog must still hold the session at waiting.
-	realFooter := "Which approach?\n\n❯ 1. Rewrite\n  2. Patch\n\n" +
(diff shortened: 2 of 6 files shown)


internal/session/scenario_test.go
```diff
@@ -1097,3 +1097,44 @@ func TestScenarioStaleWaitingLatchRecoversToWaiting(t *testing.T) {
...
+		Name: "hook=finished + prose describing the footer → finished, not waiting",
+		Events: []ScenarioEvent{
+			{At: 0, Hook: "finished", Pane: "@fixture:" + fixture},
+			{At: 30 * time.Second, Pane: "@fixture:" + fixture}, // still parked; must not drift to waiting
+			{At: 60 * time.Second, Hook: "waiting", Pane: realFooter},
+		},
+		Checks: []ScenarioCheck{
+			{At: 0, Expected: StatusFinished},                // before fix: waiting
+			{At: 30 * time.Second, Expected: StatusFinished}, // before fix: waiting
+			{At: 60 * time.Second, Expected: StatusWaiting},  // a real dialog still registers
```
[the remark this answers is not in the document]
hayke102: Good catch, and it was vacuous in a second way too: `realFooter` carried `❯ 1. Rewrite` + `2. Patch` + `Esc to cancel`, so the numbered-menu structural check matched it before the footer check ever ran. Dropping the hook alone would still have left the assertion passing with `isAskUserQuestionFooter` deleted.

Fixed in 42595dd: hook stays `finished` as you suggested, and the pane is now the checkbox-focus shape (no numbered options), so the footer is the only thing that can identify it. Verified by stubbing `isAskUserQuestionFooter` to return false — `at t=1m0s: expected "waiting", got "finished"`. It's load-bearing now.

changelog/unreleased/prose-footer-false-waiting.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Talking about a prompt no longer looks like one.** A finished session whose output mentioned `n to add notes` and `Esc to cancel` in the same sentence was read as a live question dialog and stuck on waiting until the text scrolled away.
```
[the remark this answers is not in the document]
hayke102: Agreed — the convention in CLAUDE.md asks for a bold 2–4 word headline in second person and mine was nine words and third person. Rewritten in 42595dd to **Finished sessions stay finished.** I kept the concrete trigger (`n to add notes` / `Esc to cancel` in one sentence) in the body, since the specificity is what makes the entry legible.

internal/session/session.go
```diff
@@ -2255,9 +2255,10 @@ func detectWaiting(recentLines []string, _ string, log *slog.Logger) Status {
 
 	// Structural check: AskUserQuestion tool dialog.
```
hayke102, line 2272: The field-matching discipline stops here, but the same bug class is still open one check above. The numbered-menu check (`session.go:2236`, outside this diff) still substring-matches:

```go
if strings.Contains(lower, "esc to cancel") || strings.Contains(lower, "approve with this feedback") {
```

So the exact prose class this PR targets still pins a finished session to waiting whenever the pane *also* quotes a numbered menu — and a session doing status-detection work (precisely the one that produced the new fixture) routinely prints both. Verified against HEAD of this branch, this pane returns `waiting`:

```
The dialog renders options like:
  ❯ 1. Yes
  2. No, tell Claude what to do differently

The footer (n to add notes + Esc to cancel) is what identifies AskUserQuestion.
```

Worth either applying the same whole-field rule to `hasMenuFooter`, or requiring the footer to be a standalone hint line rather than a substring of arbitrary prose. A position guard would close a second residual too: a verbatim footer quoted inside prose still matches `isAskUserQuestionFooter` anywhere in the bottom 15 lines (confirmed), where requiring it to be the last non-empty line — the same shape as the dock-row check — would not.

internal/session/session.go
```diff
@@ -2296,6 +2293,34 @@ func detectWaiting(recentLines []string, _ string, log *slog.Logger) Status {
...
+// finished session to waiting through applyHookFinished's pane override, and did so
+// on every tick for as long as the sentence stayed in the bottom 15 lines. Requiring
+// whole fields keeps prose out: a sentence mentioning the hints has no "·" between
+// them, so the hints are never fields of their own.
+//
+// "Esc to cancel" is required, plus either sibling hint. "n to add notes" is what
+// covers the SINGLE-question variant, which omits the Tab hint (see above).
+func isAskUserQuestionFooter(line string) bool {
+	var hasCancel, hasHint bool
+	for _, field := range strings.Split(line, "·") {
```
hayke102: This makes `·` (U+00B7) the only accepted delimiter, and per the golden fixture's own note it is the *sole* signal for the single-question preview variant — the `❯ N.` cursor line is out of the 15-line window there, so nothing else structurally confirms the prompt.

That's a narrow hinge for the quieter failure direction. If Claude ever renders the footer with a different separator, inside a box border (`│ … Esc to cancel │`, where the trailing `│` becomes part of the field), or truncates hints on a narrow pane, this silently stops firing and a genuinely blocked session reads as not-waiting — which nobody notices, unlike the false-waiting this PR fixes.

Accepting a hint delimited by `·` **or** 2+ spaces / a line boundary would keep prose out (prose separates the hints with words, not padding) while surviving a rendering change.

internal/session/session.go
hayke102: Confirmed both halves against HEAD before fixing — and the `approve with this feedback` footer reproduces it too, not just `Esc to cancel`.

Fixed in 34719ad with a single position guard rather than field-matching the menu footer, because the measurement pointed there: across all six `pane_waiting_*` fixtures the real footer sits at index **0, 0, 1, 0, 0, 2** from the bottom, while both prose panes — given the input box and `⏵⏵` mode bar the TUI always renders below the conversation — put the offending line at **5 and 6**. So a footer now only counts within 3 non-empty lines of the bottom, the same shape as the dock-row guard you cite. That closes the menu half and the `isAskUserQuestionFooter` position residual together.

The options scan stays at 15 — the single-question preview variant puts its `❯ N.` at index 19, so narrowing that would break the case the fixture exists for.

One bound worth stating: a pane that *literally ends* with prose quoting a footer still matches. Not reachable in a live session, so I left it rather than adding machinery for it.

internal/session/session.go
hayke102: Verified the characterisation — box-border, two-space, pipe and bullet renderings all returned false. Taken in 34719ad: fields now split on `·` **or** a run of 2+ spaces, with a single space left alone since that is the space *inside* a hint (`n to add notes`), not between hints.

Sequencing mattered here. Loosening the delimiter on its own would have widened the prose hole this PR exists to close, so it went in after the position guard from your other thread — with position doing the rejection, the delimiter no longer has to carry it.

Skipped the box-border case (`│ … Esc to cancel │`) as genuinely speculative — no captured pane shows it, and stripping border glyphs is a guess at a rendering that does not exist.

The narrow-pane sub-case is worth separating out: `capture-pane` **wraps** rather than truncates, so the hints land on different lines and no same-line rule reaches them, whatever the delimiter. I pinned that as an explicit test expectation (`footer hints split across lines not matched (wrap bound)`) so the limit is a known quantity instead of something rediscovered later.

PR #271, opened 2026-08-24
fix(accounts): keep the reset countdown when a quota window rolls over

hayke102:
At 100% the header readout showed no reset time:

```
before:  hayke102 100% 10%(6d)
after:   hayke102 100%(now) 10%(6d)   →   hayke102 2%(5h) 11%(6d)   (~2s later)
```

The countdown wasn't missing *because* it was 100% — it was missing because the reset had **already passed**, and it vanished at the exact moment its owner was watching it.

## Two causes

**The display.** `resetIn` returned `""` for both "no reset known" and "reset already elapsed". Those are different facts. A known-but-passed reset now renders `now`; only an unknown one still drops the parenthetical. A countdown that disappears on a timer is also the chip changing shape, which this readout was rebuilt to stop.

**The data.** The `100%` beside it was stale for up to three minutes — the poll is throttled to 180s per account, so a window that had refilled kept reading spent. `Usage.StaleAfterReset` breaks the throttle across a window's own reset, so a fresh reading lands within a tick. It's bounded below by `AttemptedAt` rather than simply "the reset is in the past": a poll that *fails* at the boundary moves `AttemptedAt` past it, so the account backs off at the normal cadence instead of retrying every tick against an endpoint that just refused.

Worth noting the header was contradicting fleet's own model — `Usage.windowSpent` already treats a passed reset as refilled, while the strip kept painting it red.

## Tests

- `TestStaleAfterResetOnlyForAWindowThatJustRolledOver` — both windows, a future reset, an unknown reset, and the failed-poll back-off case.
- `TestPassedResetKeepsACountdown` — both styles; grouped pools its two countdowns, so `now` has to keep the group whole rather than taking it down with it.

https://claude.ai/code/session_01VsCi5biiKSLgHduZpHGgYY

## Summary by CodeRabbit

* **Bug Fixes**
  * Quota usage now refreshes promptly when a quota window resets, rather than waiting for the standard polling interval.
  * Prevented repeated polling attempts when a reset occurs during a failed refresh.
  * Account usage displays **“now”** after a quota reset instead of hiding the countdown.
  * Preserved reset countdowns across supported account-usage display styles.


internal/ui/accounts_header.go
```diff
@@ -188,7 +188,18 @@ func resetIn(u claudeaccount.Usage, win claudeaccount.Window, now time.Time) str
...
+		// exact moment the user is watching it, leaving a stale red "100%" with
+		// nothing beside it to say the window came back — the one state where
+		// the horizon matters most, reporting nothing. It is also the chip
+		// changing shape on a timer, which is what this readout was rebuilt to
+		// stop.
+		//
+		// Not a resting state: a passed reset breaks the poll's own throttle
+		// (Usage.StaleAfterReset), so the figure beside it is re-read within a
+		// tick and this shows only in the seconds between the two.
+		return quotaResetStyle.Render("now")
```
hayke102: The comment below says this "shows only in the seconds between the two", but nothing enforces it: `now` is rendered purely off the wall clock, while the throttle break that's meant to make it transient *is* bounded — `StaleAfterReset` keys on `AttemptedAt`, and the failure path in `maybePollAccountUsage` advances `AttemptedAt` on every failed poll while keeping the old resets.

So across an outage that spans a reset (endpoint unreachable for >5h, or asleep and still down on wake), `refilledBetween` is false, nothing re-polls, `FiveHourReset` stays fixed in the past, and the header renders `34%(now) 71%(now)` for the entire outage.

It isn't *false* there — the reset really did pass; the percentage beside it is what's stale. But it reads as "both windows just refilled" next to figures that are hours old, which is the one claim the old `""` didn't make. Gating on `FetchedAt` — render `now` only when the reset passed since the reading was taken, drop the parenthetical otherwise — would make the comment true and leave the outage case rendering exactly what it rendered before.

internal/claudeaccount/client.go
```diff
@@ -123,6 +123,30 @@ func (u Usage) SpentWindow(now time.Time) (Window, time.Time, bool) {
...
+// not merely old, it is about the wrong window, and it is stale in the
+// direction that matters: an account that has just come back keeps reading
+// spent for another three minutes, exactly while its owner is watching the
+// countdown run out.
+//
+// Bounded below by AttemptedAt rather than simply "the reset is in the past":
+// a poll that fails at the boundary moves AttemptedAt past the reset, so the
+// account backs off at the normal cadence instead of retrying every tick
+// against an endpoint that just refused.
+func (u Usage) StaleAfterReset(now time.Time) bool {
```
hayke102, line 140: Lower confidence, more of a question: is a *successful* poll that returns an already-past reset reachable?

If the endpoint is a few seconds behind its own boundary and the throttle-breaking poll comes back with the old `FiveHourReset` (still in the past) and the old spent percentage, then `AttemptedAt = now` lands beyond it, `refilledBetween` goes false, and the next read is a full `MinPollInterval` away — reproducing the same three-minute `100%(now)` window this closes, just one poll later.

I haven't confirmed the API ever lags like that, so this may be unreachable in practice. If it's cheap, allowing one follow-up poll when a successful read returns a reset that is already in the past would close it.

changelog/unreleased/quota-reset-countdown.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Reset countdown survives the reset.** A spent account used to lose its countdown at the exact moment it hit zero, leaving a bare red `100%`. It now reads `100%(now)`, and the quota poll refreshes within a tick instead of waiting out its 3-minute throttle.
```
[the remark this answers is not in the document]
hayke102: Fixed — headline down to three words and the throttle detail dropped:

`**Countdown survives zero.** A spent account used to drop its reset countdown at the moment it hit zero, leaving a bare red \`100%\`. It now reads \`100%(now)\` and refreshes in ~2s instead of up to 3 minutes.`

Took the critique, not the wording: "refresh promptly instead of retaining stale usage" trades a number for an adjective, which `changelog/README.md` bans in the same breath ("Numbers, not adjectives").

internal/ui/accounts_header.go
hayke102: Fixed in 197cd31 — but with `Err` doing the work, because the `FetchedAt` gate alone doesn't close the case you named.

`FetchedAt` doesn't advance on a failed poll (only `AttemptedAt` does), so through an outage it holds the last *successful* read — which is by construction *before* the reset. `reset ∈ (FetchedAt, now]` stays true for the whole outage and `now` still renders. More generally `now` can only ever appear when the newest reading predates the reset (a successful poll after one returns a *future* reset), so that gate is nearly always true on its own.

What bounds it is the poll's health: `now` claims a fresh figure is a tick behind it, and fleet can't claim that while it can't reach the endpoint. So the gate is `u.Err == nil && at.After(u.FetchedAt)`, and your half still earns its place — it's what suppresses `now` when a *successful* poll returns an already-past reset (the sibling thread).

| state | Err | renders |
|---|---|---|
| healthy boundary (≤2s) | nil | `now` |
| outage spanning a reset | set | bare figure, as before |
| reading postdates its own reset | nil | bare figure, as before |

The doc comment now states both conditions instead of asserting the bound. `TestNowIsSuppressedWhenItCannotBeTransient` covers the last two rows; `TestPassedResetKeepsACountdown` had a fixture with `FetchedAt == hdrNow` — a reading taken *after* the reset, so it wasn't pinning the case it meant to — and now uses one taken 5m before.

internal/claudeaccount/client.go
hayke102: Reachable, I think — and probably not by lag, which is why I'd rather not add the follow-up poll.

A 5-hour block is anchored to its first message, so a block that lapses with no new activity plausibly keeps reporting its old `resets_at` until a new one starts. If that's what's happening, the endpoint can return a past reset for *hours*, and "one more poll" either fires too early (if it really is lag) or returns the same past reset every time — at which point the rule is an unbounded retry at tick cadence against a live endpoint. The `AttemptedAt` bound is exactly what makes the break provably once-per-boundary, and trading it away buys a case fleet can't fix by asking again: the staleness is in the answer, not in the cadence.

The display half is real though, and it's closed in 197cd31 by the `FetchedAt` gate from the sibling thread: a reading whose reset predates it renders bare, as it did before this PR, instead of claiming a refill for as long as that lasts. So the residual is a stale percentage with no horizon beside it — the pre-PR behaviour — rather than a wrong one.

Happy to reopen if you'd rather have the extra poll with its own floor.

Issue #274, opened 2026-08-25
Support passing model and effort to `fleet add` and `fleet …


eylonronen:
### Problem
Support passing model and effort to `fleet add` and `fleet worktree`

### Environment
- **Version**: 2.30.3
- **OS**: macOS 26.2 (arm64)

hayke102:
Shipped in v2.31.0 (#275). Both commands now take `--model` and `--effort`:

```
fleet wt my-branch --model opus --effort high
fleet add . --model claude-sonnet-5
```

They're one-shot — applied at launch and never persisted, because all three agents treat model and effort as session-scoped and let you change them mid-session, so a value fleet stored would be re-imposed on the next restart and silently undo that. `--effort` is refused on OpenCode, which has no equivalent on the command fleet launches.

PR #275, opened 2026-08-25
feat(cli): take a prompt, agent, account, model and effort from the CLI

### LINKED ISSUE #273: support sending initial message when using `fleet add` and …
### Problem
support sending initial message when using `fleet add` and not just when using `fleet worktree`

### Environment
- **Version**: 2.30.3
- **OS**: macOS 26.2 (arm64)

eylonronen:
## Summary

Closes #273 and #274.

**`fleet add` gains the session-shaping flags `fleet worktree` already had** — `-p`/`--prompt` (including `-p -` for stdin), `--agent`, `--account` — parsed with the same either-side peel loop, since a path is not free-form prose the way `fleet send`'s message is. Its path is now optional and defaults to cwd, but a bare `fleet add` stays a usage error: it was one before, and a mistyped `fleet add` in the wrong tab should not launch an agent wherever that tab happened to be. A `-p` prompt titles the session via `naming.GenerateTitle` and sets `TitleGenerated` exactly as the TUI's own fallback does, so the agent's own title still overrides it later while the worker stops regenerating an identical one from `FirstPrompt`.

**Both commands take `--model` and `--effort`.** One-shot on exactly `InitialPrompt`'s terms and never persisted — a decision rather than an omission: all three agents treat these as session-scoped (Claude documents `--effort` as non-persistent, Codex's `/model` popup writes its own config) and all three let the user change model mid-session, so a stored value would be re-imposed on the next restart and silently undo that choice.

Effort is **one agent-neutral flag over three spellings**, which is the per-agent divergence `internal/agent` exists to own:

| agent | model | effort |
|---|---|---|
| Claude | `--model` | `--effort` |
| Codex | `--model` | `-c model_reasoning_effort=` (no flag exists) |
| OpenCode | `--model` | `--variant` |

Unlike the prompt, these ride **in** the command string that `tmux send-keys` types into the pane's shell, so `agent.ValidateLaunchValue` rejects anything outside a bare-name shape at parse time, before anything is created. Model ids change too fast for an allowlist, so shape is the check and the agent rejects an unknown name itself. Both are appended ahead of the prompt argument, since `--` ends option parsing.

### Three pre-existing bugs in `runAdd`, fixed on the way

All from it never having been held to the same standard as `runWorktree`:

- **No `migration.Run()`.** `session.Open` creates `state.db`, after which `migrateConfigDir` permanently bails out (\"both dirs have state.db\") and still writes its marker — silently stranding a `brizz-code` user's sessions, pins and slot bindings if `fleet add` was their first fleet command.
- **No `debuglog.Init()`.** `Session.Start` logs at Info and debuglog's fallback writes to stderr, so the command printed slog lines over the user's terminal.
- **No `installAgentHooks`.** A CLI-created session had no status detection at all, and `--agent codex` would have opened onto Codex's untrusted-directory prompt.

A failed `SaveSession` also no longer leaves an orphaned live tmux pane nothing points at.

### `cmd/fleet/session_launch.go`

Owns what the two commands do identically once they know where the session goes: `resolveLaunchAgent`, `resolveLaunchAccount`, `installAgentHooks`, `launchSession`.

It exists because the two had **drifted**: `fleet add` carried a shorter copy of the account logic with no explicit-account branch, no note when the strategy couldn't rank, and `GuardConflictingAuth` called from inside the wrong branch — a second, weaker implementation of a policy whose whole point is that a mistake bills the wrong subscription.

The lift is **verbatim**, so the diff reads as a move. The helpers still `os.Exit` rather than returning errors, which was a deliberate call to keep the diff small; the cost is that the policy stays untestable and only review catches a change to it. Worth a careful look at that hunk in particular.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor
- [ ] Documentation
- [ ] Other

## Checklist

- [x] Linter passes (`make lint`) — 0 issues
- [x] Documentation updated — `CLAUDE.md`, `fleet -h`, and `internal/skill/SKILL.md` (new "directory that already exists" and "choose the model and effort" sections, so agents driving the CLI know the flags exist)
- [x] Changelog fragment added — three: two `added`, one `fixed`, none highlighted

## Reviewer notes

Two things I could not verify from here:

1. **`--effort` for Codex and OpenCode is documented-but-unverified.** Claude's `--effort` is confirmed against current docs. Codex's `-c model_reasoning_effort=` comes from its own TUI's config-write path, and OpenCode's `--variant` is documented on `opencode run` rather than the bare TUI. Worth one manual smoke test each.
2. **The account-policy lift has no test coverage** (see above).

Verified locally: every guard fires before anything is created —

```
$ fleet add . --model 'opus; rm -rf ~'
invalid --model "opus; rm -rf ~" — expected a bare name like opus, high, or anthropic/claude-sonnet-5
$ fleet add . --agent codex --account a@b.com
--account only applies to claude sessions (got --agent codex)
$ fleet wt foo --no-session --model opus
--model has no effect with --no-session (no session is started)
```

and the generated launch commands:

```
claude    claude --model opus --effort high -- "$FLEET_INITIAL_PROMPT"
codex     codex --model opus -c model_reasoning_effort=high -- "$FLEET_INITIAL_PROMPT"
opencode  opencode --model opus --variant high --prompt="$FLEET_INITIAL_PROMPT"
```

## Summary by CodeRabbit

- **New Features**
  - Added `fleet add [path]` for launching sessions in existing directories, with prompts, agent, account, model, and effort options.
  - Prompts can come from stdin, and paths default to the current directory.
  - Added `--model` and `--effort` overrides to `fleet add` and `fleet worktree`.
- **Bug Fixes**
  - Existing sessions and settings are migrated before `fleet add` runs.
  - Diagnostics from `fleet add` are written to `debug.log`.
  - Effort settings are rejected for unsupported OpenCode sessions.
- **Documentation**
  - Updated CLI usage, skill guidance, and release documentation.


internal/agent/agent.go
```diff
@@ -108,6 +111,58 @@ type LaunchOpts struct {
...
+//
+// Codex's value needs no quoting: `-c` parses the value as TOML and falls back
+// to the raw string when that fails, so a bare `high` arrives intact and the
+// shell has nothing to chew on.
+func (t Type) effortArg(effort string) string {
+	switch t {
+	case Codex:
+		return " -c model_reasoning_effort=" + effort
+	case OpenCode:
+		return " --variant " + effort
```
hayke102: **`--variant` is not a flag on the OpenCode command fleet actually launches.** It is only defined on `opencode run`; `BuildLaunchCmd` produces the default TUI command (`opencode [project]`), and OpenCode's yargs is in **strict** mode — an unknown flag prints usage and exits.

Verified against the installed binary:

```
$ opencode --help          # default-command options: --port --hostname --mdns --cors -m/--model
                           #   -c/--continue -s/--session --fork --prompt --agent   (no --variant)
$ opencode run --help      # ... --variant   model variant (provider-specific reasoning effort)
$ opencode --variant high  # prints the command list and exits instead of starting the TUI
```

So `fleet add . --agent opencode --effort high` prints "Created OpenCode session …", saves the row and pins the repo, while the pane runs `opencode --model X --variant high`, prints usage and exits — a live tmux session parked at a shell prompt with no agent. That is exactly the failure `promptArg`'s comment describes for a mis-parsed argument, and nothing in fleet detects it at launch. `TestBuildLaunchCmdModelAndEffortPerAgent` pins the wrong spelling, so it passes.

Claude's `--effort` and Codex's `-c model_reasoning_effort=` are both correct — I confirmed the latter survives `codex exec --strict-config`.

cmd/fleet/add.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+		positional = append(positional, remaining[0])
+		rest = remaining[1:]
+	}
+
+	if len(positional) > 1 {
+		return o, fmt.Errorf("unexpected argument %q — expected a single path", positional[1])
+	}
+	if len(positional) == 1 {
+		o.path = strings.TrimSpace(positional[0])
```
hayke102, line 103: **An explicitly empty positional silently falls back to cwd**, defeating the guard `errMissingAddArgs` exists for. `TrimSpace` here can yield `""`, and `runAdd` (line 204) treats `""` as "no path given" and substitutes `os.Getwd()`.

So `fleet add "$REPO" -p "fix it"` with `REPO` unset or typo'd starts an agent in whatever directory the tab happened to be in — the same failed-substitution case that is explicitly rejected for `-prompt` fourteen lines below. Verified live: `fleet add ""` created a real session in the cwd.

Suggest tracking whether a positional was supplied and rejecting an empty/whitespace-only one, rather than folding it into the "omitted" case.

cmd/fleet/worktree.go
```diff
@@ -257,93 +276,23 @@ func runWorktree(args []string) {
...
 	account := ""
-	// Re-checked here, not only at parse time: the parse-time guard can compare
-	// --account against --agent only when --agent was given. With default_agent
-	// set to codex or opencode, an explicit --account would otherwise be dropped
-	// on the floor and the session created as the wrong agent — the silent typo
-	// the explicit validation exists to prevent.
-	if opts.account != "" && ag != agent.Claude {
+	if !opts.noSession {
+		account = resolveLaunchAccount(cfg, accounts, repoPath, ag, opts.account)
+	} else if opts.account != "" && ag != agent.Claude {
```
hayke102: This branch is unreachable. It only runs when `opts.noSession` is true, but `parseWorktreeArgs` already rejects `--account` alongside `--no-session` (worktree.go:198-200), so `runWorktree` can never arrive here with both set.

The comment reads as a live guard, which will mislead the next reader of the one policy block the PR description asks reviewers to read closely. Either drop the branch or drop the parse-time rejection.

CLAUDE.md
```diff
@@ -107,8 +109,11 @@ Styles in `styles.go` are declared **bare** and constructed **only** in `ApplyPa
...
....PromptEnvVar`, set in `sessionEnv`), which no shell parses, and the command carries only `"$FLEET_INITIAL_PROMPT"` — one argv element whatever it contains. Two tests pin this: `TestBuildLaunchCmdNeverEmbedsPromptText` and `TestInitialPromptTravelsInEnvNotCommand`. Quoting only settles the **shell**; the agent's own argv parser is a second parser, and it reads a leading `-` as an option and *exits* — leaving a live session at a shell prompt with the message gone. So the prompt sits behind a separator that ends option parsing: `-- "$FLEET_INITIAL_PROMPT"` for Claude/Codex, `--prompt="$FLEET_INITIAL_PROMPT"` for OpenCode (with a space, yargs reads the *value* as a fresh option). `fleet send` explicitly promises a message may start with a dash, and the suspended-wake path routes that message through here. **One-shot**: every launch path clears it on success via `consumeInitialPromptLocked` — `Start`, `Restart` (the `r` key, and the idle-suspend wake) and `RespawnClaude` all build from `buildAgentCmd` + `sessionEnv`, so a path that forgets silently re-asks the original task. Never persisted, or it would replay on every app restart. Kept on a *failed* launch so a retry still carries it.
```
hayke102: Stale after this PR, on both counts: `fleet worktree -p` is no longer the only thing that sets `InitialPrompt` (`fleet add -p` does too), and `consumeInitialPromptLocked` was renamed to `consumeLaunchOverridesLocked`. Line 116 announces the rename, but this is the bullet a reader searching for `InitialPrompt` lands on first.

Same stale symbol in `internal/ui/app.go:3746` — the comment in `handleSessionCreate` still names `consumeInitialPromptLocked`. That file is unchanged in this diff so I could not anchor a thread there.

internal/agent/agent.go
eylonronen: Confirmed and fixed in 9038ef8 — thank you, this was a real one and my PR note ("documented-but-unverified") was not good enough for something that fails this way.

Verified independently against the source rather than the binary, and it agrees with your run: `TuiThreadCommand` is `$0 [project]` and its builder declares only `project`, `prompt` and the network options; the handler reads `args.model` but there is no `variant`. `--variant` is declared in `cli/cmd/run.ts`. The root parser ends in `.strict()`, so the unknown option is rejected rather than ignored. My original lookup had actually returned the `run.ts` definition — I read a subcommand's flag as the CLI's flag.

The failure mode is the part that matters: not a dropped setting, but a session fleet reports as created, row saved and repo pinned, with a shell prompt where the agent should be.

Fix is to refuse rather than drop, since silently ignoring `--effort` would be its own quiet lie:

- `agent.SupportsEffort()` is false for OpenCode; the `--variant` branch is gone from `effortArg`.
- Rejected at parse time when `--agent` names it, and again in `guardEffortSupported` once `default_agent` is resolved — the same two-place check `--account` needs, for the same reason (the parse-time guard can't see `default_agent`).
- `BuildLaunchCmd` guards too, so a caller that forgets drops the flag instead of emitting a command OpenCode refuses.
- `--model` is untouched, since the default command's handler reads `args.model`.

`TestOpenCodeNeverReceivesAnEffortFlag` now pins it from the other direction — it asserts the built command contains none of `--variant`, `--effort`, `model_reasoning_effort`, or the effort value itself, so the next person to add a spelling here can't make it pass by pinning whatever they wrote.

cmd/fleet/add.go
eylonronen: Fixed in 9038ef8. You're right that this defeats the guard I added `errMissingAddArgs` for — and it's the same failed-substitution case I explicitly reject for `-prompt` fourteen lines down, which makes it worse than an oversight.

An empty or whitespace-only positional is now its own error (`path was empty`) rather than being folded into "omitted":

```
$ fleet add ""
path was empty
$ fleet add "   " -p "do it"
path was empty
```

Two cases added to `TestParseAddArgs`.

cmd/fleet/worktree.go
eylonronen: Dropped in 9038ef8 — the branch is gone rather than the parse-time rejection, since rejecting at parse time fails before anything is created.

Your point about the comment is the sharper one: a dead guard that reads as a live one is worse in this block than anywhere else in the diff, given the PR asks reviewers to read exactly this. The `if !opts.noSession` now carries a note saying `parseWorktreeArgs` already rejects `--account`, `--agent`, `--model` and `--effort` alongside `--no-session`, so the block can only be reached with all four unset — which explains why there's nothing to guard rather than leaving the next reader to rediscover it.

CLAUDE.md
eylonronen: Both fixed in 9038ef8, including the one you couldn't anchor.

- The `InitialPrompt` bullet now reads "set by `fleet worktree -p`, `fleet add -p`, and by `fleet send` waking a suspended session", and names `consumeLaunchOverridesLocked` with a pointer to the next bullet for the widened scope. You're right that this is where a reader searching for `InitialPrompt` lands first, so announcing the rename only in the new bullet was backwards.
- `internal/ui/app.go:3746` updated too — thanks for chasing it into a file the diff didn't touch. `grep -rn consumeInitialPromptLocked` is now clean apart from the sentence that documents the rename.

That bullet also carried the `--variant` claim from your other comment; it's been rewritten to record what OpenCode actually accepts and why the flag is refused rather than dropped.

changelog/unreleased/launch-model-and-effort.md
```diff
@@ -0,0 +1,5 @@
+---
+type: added
+---
+
+**Pick the model per session** — `fleet add` and `fleet worktree` now take `--model` and `--effort`, so a session can start on `opus` at `xhigh` without touching your defaults. Works for all three agents.
```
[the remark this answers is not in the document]
hayke102: Confirming this one — it needs a fix before merge.

`9038ef8` turned `--effort` into a hard error for OpenCode (`agent.SupportsEffort` is false, refused at parse time and again in `guardEffortSupported`), so "Works for all three agents" is now wrong in the direction that costs the reader something: they'll read the release notes, try `--agent opencode --effort high`, and get a rejection.

The split is worth stating rather than dropping, since `--model` genuinely is all three. Suggested replacement, keeping it to the two sentences `changelog/README.md` asks for:

> **Pick the model per session** — `fleet add` and `fleet worktree` now take `--model` and `--effort`, so a session can start on `opus` at `xhigh` without touching your defaults. `--effort` is Claude and Codex only; OpenCode has no equivalent flag, so fleet tells you instead of dropping it silently.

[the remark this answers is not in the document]
hayke102: Fixed in `cdf7c23` — the fragment now names the exception rather than claiming all three agents.

PR #276, opened 2026-08-25
fix(ui): put mouse reporting back after an attach

hayke102:
Follow-up to #268, which fixed the scroll repaint storm but listed the attach round trip as untested. That was the one path that undoes it.

Reported as "fleet was very slow"; perfwatch had caught it as two `update_storm` dumps at **255 and 272 MouseWheelMsg/sec holding 250% CPU**, each event buying a full `View()` over 69 sessions.

## The chain

1. fleet sets `mouse on` per session, so attaching puts the outer terminal into DECSET 1000/1002/1006.
2. The attach ends by **killing** the tmux client (`Ctrl+Q` → `close(detachCh); cancel()` in `internal/tmux/pty.go`) rather than detaching it, so tmux never runs its terminal cleanup and those modes stay on in fleet's own terminal.
3. Bubble Tea never takes them off. Its renderer emits a mouse-mode change only when the mode *differs* from the last view's, and `Home.chrome` reports `MouseModeNone` on every frame; the post-Exec resume path writes nothing at all for `None`. Its input parser decodes the incoming bytes into `MouseWheelMsg` regardless of the mode it asked for.

So one attach re-arms the #268 storm for the rest of the process, and nothing looks wrong until someone scrolls.

The timeline from this user's machine is what pins it as a regression rather than an incomplete fix — 23 storms the day before #268 shipped, then **35 hours clean**, then an attach, then storms again.

## Measured, not reasoned

A probe reproducing fleet's exact sequence (tmux session, `mouse on`, PTY attach, SIGKILL the client the way `Ctrl+Q` does):

```
mode 1000: ON  <-- LEAKED
mode 1002: ON  <-- LEAKED
mode 1006: ON  <-- LEAKED
```

Exactly the SGR wheel-reporting combination. `Reassert` clears all of them.

## Why `Reassert` is not just a second `Disable`

Two of `Disable`'s four sequences carry state:

- **XTSAVE (`1007s`)** has a single save slot — repeating it overwrites the user's original value with the already-cleared one, so `Restore` hands the terminal back wrong on exit.
- **`DisableKittyKeyboard`** pushes onto a stack `Restore` pops exactly once — each repeat leaks an entry, once per attach, all session long.

What's left is the two plain idempotent sets plus the mouse reset. Pinned by `TestReassertRepeatsNothingStateful`.

## Placement

In `attachCmd.Run` rather than a `tea.Exec` callback: all three attach paths (session, drawer shell, account login pane) run through that one type, so a fourth gets it for free instead of having to remember. Before Bubble Tea's `RestoreTerminal` re-enters the alternate screen — the order `termkeys.Disable` was verified in.

## Testing

- `TestReassertTurnsMouseReportingOff`, `TestReassertRepeatsNothingStateful`, `TestReassertPropagatesWriteError`
- `TestAttachRestoresTerminalModes` and `TestChromeKeepsMouseReportingOff` — AST guards, **each confirmed to fail with the fix removed**
- `go test -race ./...` clean, `golangci-lint` 0 issues

`TestCredentialResolutionOrder` (`internal/linear`) fails on this branch and identically on clean `master` — it reads the real keychain. Pre-existing, untouched.

## Out of scope, found while investigating

- `refreshAllGitAndPR` cycles drifting 745ms → 2725ms across 31 repos every ~2s; top of the mutex profile via fork-lock contention.
- `~/.config/fleet/stalls/` at 480 MB / 38 files — each dump is 13.8 MB and costs ~1.2s stop-the-world, re-firing every 5s during a storm.

https://claude.ai/code/session_01K8LQCFMKTuXAVEx54H6GyX

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored terminal keyboard and mouse settings after attaching to a session.
  * Prevented excessive screen redraws and CPU usage caused by lingering mouse-reporting modes.
  * Preserved terminal behavior when returning from full-screen sessions.

* **Tests**
  * Added coverage to verify terminal modes are restored correctly and restoration errors are handled.

### DIFF
--- a/changelog/unreleased/attach-restores-mouse-mode.md
+++ b/changelog/unreleased/attach-restores-mouse-mode.md
@@ -1,0 +1,5 @@
+---
+type: fixed
+---
+
+**Scrolling stays fast after you detach.** Attaching to a session left your terminal reporting mouse events — `Ctrl+Q` kills the tmux client before tmux can clean up — so the next scroll brought back the ~270 redraws/sec that pinned fleet at 250% CPU. Detaching now puts the terminal back.
--- a/changelog/unreleased/attach-restores-mouse-mode.md
+++ b/changelog/unreleased/attach-restores-mouse-mode.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**Scrolling stays fast after you detach.** Attaching to a session left your terminal reporting mouse events — `Ctrl+Q` kills the tmux client before tmux can clean up — so the next scroll brought back the ~270 redraws/sec that pinned fleet at 250% CPU. Detaching now puts the terminal back.
+**Scrolling after detach.** Attaching to a session used to leave your terminal sending mouse events to fleet, so the next scroll pinned it at 250% CPU until you stopped. Detaching now puts the terminal back.
--- a/internal/termkeys/termkeys.go
+++ b/internal/termkeys/termkeys.go
@@ -69,26 +69,38 @@ const disablePreamble = disableModifyOtherKeys + ansi.DisableKittyKeyboard +
 const disableMouseReporting = ansi.ResetModeMouseNormal + ansi.ResetModeMouseButtonEvent +
 	ansi.ResetModeMouseAnyEvent + ansi.ResetModeMouseExtSgr
 
-// Reassert re-applies what Disable asked for, for callers returning from a
-// full-screen attach that handed the terminal to tmux and got it back dirty:
-// fleet's attach ends by killing the tmux client (Ctrl+Q, see
-// internal/tmux/pty.go), so tmux never runs its own cleanup and whatever it
-// enabled stays on. It also turns mouse reporting off, which Disable never had
-// to: Bubble Tea owns that mode via Home.chrome, and owns it only on the way in
-// — its renderer emits a mouse-mode change only when the mode *differs* from
-// the last view's, and fleet's is MouseModeNone on every frame including the
-// first one after an attach. So nothing else will ever take reporting back off,
-// and the next scroll is a MouseWheelMsg burst costing a full View() apiece.
+// Reassert clears the terminal modes a full-screen attach leaves switched on,
+// for callers returning from one: fleet's attach ends by killing the tmux client
+// (Ctrl+Q, see internal/tmux/pty.go), so tmux never runs its own cleanup and
+// whatever it enabled stays on.
 //
-// It is deliberately not a second call to Disable, because two of Disable's
-// four sequences carry state and are not idempotent: XTSAVE (1007s) would
-// overwrite the saved original with the already-cleared value, so Restore would
-// hand the user back the wrong mode on exit, and ansi.DisableKittyKeyboard
-// pushes onto the Kitty keyboard stack, which Restore pops exactly once — every
-// extra push would leak an entry. What is left is the two sequences that are
-// plain idempotent sets, plus the mouse reset.
+// Mouse reporting is the one that matters, and Disable never had to touch it.
+// Bubble Tea owns that mode via Home.chrome — but owns it only on the way in: its
+// renderer emits a mouse-mode change only when the mode *differs* from the last
+// view's, and fleet's is MouseModeNone on every frame including the first one
+// after an attach, so nothing else will ever take reporting back off. The next
+// scroll is then a MouseWheelMsg burst costing a full View() apiece.
+//
+// Key reporting is deliberately *not* re-asserted, because it is not ours to
+// own here: tea.Exec calls RestoreTerminal the moment attachCmd.Run returns
+// (v2.0.7 exec.go:125), which reaches cursedRenderer.start() with a non-nil
+// lastView and unconditionally writes SetModifyOtherKeys2 — \x1b[>4;2m —
+// at cursed_renderer.go:136. A mode-0 written here would be overwritten
+// microseconds later, so including it would buy nothing but a comment the
+// terminal contradicts. Harmless in practice: Bubble Tea v2 decodes CSI-u and
+// modifyOtherKeys alike, which is why the Ctrl+K bug Disable exists for has not
+// come back. Disable itself is left as-is — its mode-0 is overwritten on frame
+// one for the same reason (cursed_renderer.go:386), but that predates this and
+// changing it is a separate question.
+//
+// What is left is two sequences, and both are plain idempotent resets — which is
+// also why this is not a second call to Disable. Two of Disable's four carry
+// state: XTSAVE (1007s) would overwrite the saved original with the
+// already-cleared value, so Restore would hand the terminal back wrong on exit,
+// and ansi.DisableKittyKeyboard pushes onto the Kitty keyboard stack, which
+// Restore pops exactly once — every extra push would leak an entry.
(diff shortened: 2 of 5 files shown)


internal/termkeys/termkeys.go
```diff
@@ -60,6 +60,38 @@ const (
...
+//
+// It is deliberately not a second call to Disable, because two of Disable's
+// four sequences carry state and are not idempotent: XTSAVE (1007s) would
+// overwrite the saved original with the already-cleared value, so Restore would
+// hand the user back the wrong mode on exit, and ansi.DisableKittyKeyboard
+// pushes onto the Kitty keyboard stack, which Restore pops exactly once — every
+// extra push would leak an entry. What is left is the two sequences that are
+// plain idempotent sets, plus the mouse reset.
+func Reassert(w io.Writer) error {
+	_, err := io.WriteString(w, disableModifyOtherKeys+disableAltScreenScroll+disableMouseReporting)
```
hayke102: `disableModifyOtherKeys` is a no-op here — Bubble Tea overwrites it microseconds later.

`tea.Exec` calls `p.RestoreTerminal()` as soon as `attachCmd.Run` returns (v2.0.7 `exec.go:125`) → `startRenderer()` → `cursedRenderer.start()`, which with `lastView != nil` (always true after an attach) unconditionally writes `ansi.SetModifyOtherKeys2` — `\x1b[>4;2m` — at `cursed_renderer.go:136`. So every attach ends in modifyOtherKeys **mode 2**, the exact state this sequence exists to prevent, while this doc comment and the changelog both say it was restored.

The reasoning in `app.go` ("the order `termkeys.Disable` was verified in") doesn't transfer: at startup `Disable` runs *before* `p.Run`, where `start()` early-returns on `lastView == nil`.

Same is true at startup for a different reason, worth knowing before deciding what to do here: the first `flush()` also writes `SetModifyOtherKeys2` when `s.lastView == nil` (`cursed_renderer.go:378`), so `Disable`'s mode-0 has been overwritten on frame one ever since the v2 migration. Harmless in itself — v2 decodes CSI-u / modifyOtherKeys fine, which is why the Ctrl+K bug never came back — but it means `disableModifyOtherKeys` is inert everywhere, not just on this path.

The other two sequences do survive: the renderer emits nothing for `MouseModeNone`, and never touches 1007. So the shipped scroll-storm fix works; only the key-reporting half is dead. Suggest dropping `disableModifyOtherKeys` from `Reassert` and saying in the comment that key reporting is Bubble Tea's to own — otherwise the comment asserts something the terminal contradicts.

internal/ui/app.go
```diff
@@ -3289,7 +3290,20 @@ type attachCmd struct {
...
+	// Every attach path (session, drawer shell, account login pane) runs through
+	// here, so the terminal is put back in one place rather than in three tea.Exec
+	// callbacks a fourth path could forget. The attach ends by killing the tmux
+	// client rather than detaching it (Ctrl+Q, internal/tmux/pty.go), so tmux never
+	// runs its terminal cleanup and the mouse reporting it enabled for `mouse on`
+	// stays on in our terminal — where nothing else would ever turn it off, and the
+	// next trackpad scroll lands as the MouseWheelMsg repaint storm #268 fixed.
+	// Before Bubble Tea's RestoreTerminal re-enters the alternate screen (exec.go
+	// calls it after Run), which is the order termkeys.Disable was verified in.
+	if rerr := termkeys.Reassert(os.Stdout); rerr != nil {
```
hayke102, line 3303: This relies on being the **last** write to stdout after the attach, and `Attach` doesn't guarantee that.

`cleanupAttach` (`internal/tmux/pty.go:143`, not in this diff) closes the pty and waits **20ms** on `outputDone`, then returns — there's no `wg.Wait()`, so the output-copy goroutine can still be flushing an already-read chunk to `os.Stdout` concurrently with, and after, this call.

Narrow window — closing the pty ends the copy on the next read, so it's at most one buffered chunk — but if that tail carries a mouse-enable DECSET (tmux re-emits `?1000h`/`?1002h`/`?1006h` on redraw/attach) the reset is silently undone and the storm is back, with nothing in the log to show why. Pre-existing race, but this is the first line to depend on the ordering; a `wg.Wait()` before `Attach` returns would close it.

changelog/unreleased/attach-restores-mouse-mode.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Scrolling stays fast after you detach.** Attaching to a session left your terminal reporting mouse events — `Ctrl+Q` kills the tmux client before tmux can clean up — so the next scroll brought back the ~270 redraws/sec that pinned fleet at 250% CPU. Detaching now puts the terminal back.
```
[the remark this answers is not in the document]
hayke102: Fixed — trimmed to the user-facing change.

Went with different wording than suggested, though: `**Fast scrolling.**` drops the annoyance being named, which the changelog guide asks for explicitly, and "stale mouse events causing ~270 redraws/sec" is still mechanism in a different coat. Kept the trigger (attaching) and one number, dropped the tmux internals:

**Scrolling after detach.** Attaching to a session used to leave your terminal sending mouse events to fleet, so the next scroll pinned it at 250% CPU until you stopped. Detaching now puts the terminal back.

internal/termkeys/termkeys.go
hayke102: Fixed — you're right, and I verified every link before changing it.

`start()` early-returns only at `cursed_renderer.go:83-85` on `lastView == nil`, which is never true after an attach, so line 136 writes `\x1b[>4;2m` unconditionally. Our mode-0 was overwritten every time.

Confirmed the other two survive, which is why the shipped scroll fix works: `grep 1007` over bubbletea returns nothing at all, and the mouse switch is empty for `MouseModeNone` in both `start()` and `flush()`.

`Reassert` now writes `1007l` + the four mouse resets, and the comment says key reporting is Bubble Tea's to own after an attach instead of claiming a state the terminal contradicts.

Left `Disable` alone — your `flush()`:386 point checks out too, so its mode-0 has been overwritten on frame one since the v2 migration, but that predates this PR and is a separate question. Added `TestReassertLeavesKeyReportingToBubbleTea` so the omission doesn't read as an oversight and get 'fixed' back in; confirmed it fails with the sequence restored.

internal/ui/app.go
hayke102: Confirmed, and the ordering dependency is real: `cleanupAttach` caps the `outputDone` wait at 20ms with no join, so `io.Copy` can still flush a chunk after `Attach` returns.

One correction on the fix though — **`wg.Wait()` would hang.** `wg` has four `Add(1)`/`Done()` pairs and is never waited on today, and one of the tracked goroutines is `forwardStdinToPTY`, which blocks in `os.Stdin.Read` until the next keypress. Joining all four would freeze the TUI after every detach until a key was pressed. The targeted version is relaxing the 20ms cap on `outputDone` alone, since `ptmx.Close()` makes the in-flight read return through the runtime poller.

Leaving it out of this PR deliberately: the race predates it and the window is at most one buffered chunk, whereas widening a three-file PR into the attach path to fix it probabilistically is the worse trade. Worth its own change — happy to open one.

PR #278, opened 2026-08-25
feat(ui): autocomplete the worktree dialog's base branch field

### LINKED ISSUE #277: Add Autocomplete to the create worktree flow for the base b…
### Problem
Add Autocomplete to the create worktree flow for the base branch, like you already do in other places

### Environment
- **Version**: 2.29.0
- **OS**: macOS 26.4.1 (arm64)

hayke102:
Closes #277.

The `w` dialog's **New branch** field learned to suggest Linear tickets; the **Base branch** field was still a bare input pre-filled with `GetDefaultBranch`, so basing a worktree on anything else meant typing the name from memory, exactly.

```
Base branch:
> origin/master
    origin/master
  ▸ origin/base-branch
    origin/fix/mouse-reporting
    origin/chore/deps

New branch:
> feature/my-feature

↓ branches  tab: next field  ⏎ create  esc: cancel
```

Rows sit under the field, on the ticket rows' vocabulary (`▸` + `selTitle`, dim otherwise, capped at 5) and through the same `setSelection` highlight. They render **only while that field carries the highlight** — the field always holds text, so rendering unconditionally would park five rows mid-dialog for everyone who never touches the base branch. The dialog opens exactly as it did before.

The branch list is fetched **once**, by `fetchWorkspaceListForRepo` on the worker goroutine beside the `GetDefaultBranch` it already ran. Filtering is then a synchronous `strings.Contains` with **none** of the ticket machinery's debounce or generation guard — those exist only because a lookup is a network round trip. A failed listing leaves the field exactly as it was.

## Three things the implementation turns on

**`tab` moves between fields, `↓`/`↑` walk rows** — split apart here. They were one merged case, which was fine while only one field grew rows and stopped being fine once both did: tabbing off a focused base field would have taken six presses to reach the next input. The `↓`/`↑` walk is the full continuous path (base input → base rows → new-branch input → ticket rows → worktree list) and must retrace itself exactly, which is why `↑` from the New branch field lands on the *last* base row, not the base input. Cost: `tab` no longer walks the worktree list; `↓` still does.

**A suggestion is stored as the exact ref that will be written into the field**, so a remote-only branch always comes out `origin/`-prefixed. `git worktree add <path> -b <new> <base>` resolves `<base>` as a plain revision with **no** remote-tracking DWIM — a bare remote-only name is a row that looks valid and fails on Enter, and the no-`-b` retry silently drops the base entirely. A field already reading `origin/…` keeps its prefix and matches only branches with a remote: `GetDefaultBranch` pre-fills `origin/<default>` precisely so a worktree starts from the remote tip, and quietly swapping that for the local branch would change what gets built without saying so.

**An at-rest field lists the alternatives, not itself.** Filtering the pre-filled `origin/<default>` yields exactly one row echoing the field back — a row that answers nothing and no-ops on Enter — so the feature was useless on the one screen it matters on. When the whole match set is that single echo, the list widens (keeping `wantRemote`: you asked for a remote ref, so the wider list is the other remote refs). Narrowly gated on **exact** equality, so a partially-typed unique match still filters normally and the list never widens under your cursor mid-word.

## Drive-by fix

`git.BranchInfo` gains `HasRemote`, which forced `ListBranches` into two passes — and that surfaced a latent bug. A branch shares its committer date with its `origin/` counterpart only when it is *level*, so a branch you have not pulled has its remote ref sort **first**, which the single-pass form both missed the flag on and **emitted twice**. That duplicate was already visible in the `b` key's branch picker. `TestListBranchesRemoteCounterparts` pins it; I confirmed it fails against the old logic.

## Testing

`make build`, `make lint` (0 issues) and `make fmt` clean; `internal/ui` and `internal/git` pass under `-race`.

- 11 new tests in `workspace_picker_branch_test.go` and `git_test.go`, covering the `origin/` rule, the remote-only dead-click case, the `tab`-skips-rows split, the `↓`/`↑` retrace, accept-without-create, keeping the keystroke that returns you to the field, at-rest widening, and rendering byte-identically when the listing failed.
- `TestWorktreeSelectionMutatorIsTheOnlyWriter` now guards `baseCursor` too, and scans all three picker files rather than one.
- `TestWorktreeCaretAndHighlightNeverCoexist` gains the base-field states, so the ≤1-`▸` invariant covers the new rows.
- Verified against this repo's real 121 branches: flags correct, no duplicates.

Two notes for the reviewer:

- `internal/linear/TestCredentialResolutionOrder` fails on my machine, but it does so identically with `master`'s `git.go` — it asserts `credential()` returns `ErrNotConnected` with nothing stored, and this machine has a Linear credential in the keychain. Pre-existing and environment-dependent, untouched here.
- Rendering was verified by calling the real `View()` across four states (the block above is its output). I did not drive the live TUI, so actually creating a worktree from a remote-only base is worth a manual pass.

https://claude.ai/code/session_01QRVcNGs7c98p7JggANsKNg

## Summary by CodeRabbit

* **New Features**
  * Added branch suggestions to the worktree picker’s Base branch field.
  * Search suggestions by typing, navigate with keyboard controls, and select a branch with Enter.
  * Clearly distinguish local, remote, and remote-backed branches, including `origin/` references.
  * Limit displayed suggestions for easier browsing.

* **Documentation**
  * Added unreleased changelog notes and expanded usage documentation for branch autocomplete.


internal/ui/workspace_picker.go
```diff
@@ -412,7 +507,12 @@ func (d *WorktreeDialog) View() string {
 	b.WriteString(DimStyle.Render("Base branch:"))
 	b.WriteString("\n")
 	b.WriteString(d.baseBranchInput.View())
-	b.WriteString("\n\n")
+	b.WriteString("\n")
+	// Branch suggestions sit directly under the field they fill. Empty unless
+	// that field carries the highlight, so the dialog opens exactly as it did
+	// before they existed.
+	b.WriteString(d.renderBranchBlock(d.innerWidth()))
```
hayke102, line 537: **The branch block has no height budget, so focusing the base field can push the dialog past the terminal.**

`renderBranchBlock` adds up to 5 rows, `wrapDialog` (`:717`) just `lipgloss.Place`s and never clamps to `d.height`, and the worktree list at `:530` renders `for i, ws := range d.workspaces` unbounded.

On 80x24 with 6 existing worktrees: **24 lines at rest, 26 once `focusBaseBranch` takes the highlight** - the footer and the last worktree rows scroll off mid-interaction.

In fairness this isn't a new class of bug: the unbounded worktree list already overflows on its own, and `renderTicketBlock` has the identical property. But the branch block lowers the threshold, and unlike the ticket rows it isn't Linear-gated - every user with a git repo grows these rows the moment they `shift+tab` into the base field.

`TestWorktreeDialogRowsNeverOverflow` is width-only, so nothing catches it. Either cap `visibleBranchCount()` against the rows actually left (`d.height` minus the rest of the dialog), or drop the worktree list while the base field is focused - the second reads more deliberately, since the list isn't what you're looking at then.

internal/git/git.go
```diff
@@ -96,31 +109,32 @@ func ListBranches(repoPath string) ([]BranchInfo, error) {
...
+	for _, r := range refs {
+		if r.remote && localSet[r.name] {
+			continue // already have local version
+		}
+		branches = append(branches, BranchInfo{
+			Name:        r.name,
+			IsRemote:    r.remote,
+			IsCurrent:   r.name == currentBranch,
+			HasRemote:   remoteSet[r.name],
+			CommitDate:  r.commitDate,
```
hayke102: **A branch behind its remote keeps the local ref's date, so the row describes a different commit than the ref it resolves to.**

The dedupe drops the remote ref whenever a local exists, but emits the survivor at the local ref's `CommitDate` - and, since `for-each-ref` already sorted by committerdate, at the local ref's later position too. Meanwhile `baseRefFor` hands `origin/<name>` to `git worktree add` whenever the field carries the prefix.

Concrete cost: `matchBranches` cuts at `branchMaxRows`=5 in commit-date order, so an unpulled `master` can fall out of the at-rest suggestion list entirely - the one base most people want. The `b`-key picker's "most recently committed first" contract goes with it.

`max(local, remote)` when a counterpart exists fixes the shown date and the ordering together. The fixture is already here: `lagging` (local 2020-01-01, `origin/lagging` 2024-01-01) is reported as 2020 and sorts below `ghost` - `TestListBranchesRemoteCounterparts` asserts `IsRemote`/`HasRemote` but never the date, which is why it passes.

internal/ui/workspace_picker.go
```diff
@@ -272,10 +308,41 @@ func (d *WorktreeDialog) Update(msg tea.Msg) (*WorktreeDialog, tea.Cmd) {
...
+	// Splitting them gives tab one meaning everywhere in this dialog.
+	case "tab":
 		switch d.focus {
 		case focusBaseBranch:
 			d.setSelection(focusNewBranch, ticketOnInput)
+		case focusNewBranch:
+			if len(d.workspaces) > 0 {
+				d.setSelection(focusWorktreeList, 0)
+			}
+		}
```
hayke102, line 340: **`tab` is now inert on the worktree list**, which sits awkwardly with the comment nine lines up: "Splitting them gives tab one meaning everywhere in this dialog."

`master`'s merged `case "tab", "down":` had a `focusWorktreeList` arm (`d.cursor++`); this one has none and doesn't wrap either. CLAUDE.md documents the first half as a deliberate cost ("`tab` no longer walks the worktree list; `down` still does") and I'm not relitigating that - but *inert* isn't the same as *doesn't walk the list*. Once you tab onto the list, tab stops responding at all; only `shift+tab` or the arrows get you out.

Wrapping `focusWorktreeList` -> `focusBaseBranch` would keep "tab = next field" true on the last field too, which is what the comment promises.

CLAUDE.md
```diff
@@ -160,7 +160,11 @@ Styles in `styles.go` are declared **bare** and constructed **only** in `ApplyPa
...
+- **An at-rest field lists the alternatives, not itself.** Filtering the pre-filled `origin/<default>` yields exactly one row echoing the field — a row that answers nothing and no-ops on Enter — so when the whole match set is that single echo, `rebuildBranchMatches` widens to the unfiltered list (keeping `wantRemote`: you asked for a remote ref, so the wider list is the other remote refs). Narrowly gated on **exact** equality with the field, so a partially-typed unique match still filters normally and the list never widens under your cursor mid-word.
```
hayke102: "a partially-typed unique match still filters normally and the list never widens under your cursor mid-word" isn't quite what the gate does. It's exact equality with the field text, which a *partial* string satisfies whenever it happens to be a complete branch name on its own.

Typing `master-fix` in a repo containing `master`: at the `r` keystroke the sole match is `master`, `branchMatches[0] == text` holds, and the filtered 1-row list is replaced by the full unfiltered 5-row list - then it re-narrows at the `-`.

Cosmetic in practice (the highlight is on the input at that point, so nothing moves under a cursor), but the sentence claims the stronger property. Either soften it to what the code guarantees - a *prefix* of a branch name never widens, because `mast` != `master` - or gate on the field having settled (unchanged since the last keystroke, or set by `Show`/`pickBaseBranch`), which is what the sentence describes.

internal/ui/app.go
```diff
@@ -7205,9 +7205,17 @@ func (h *Home) fetchWorkspaceListForRepo(repoPath string) tea.Cmd {
 		if linear.Available() {
 			linearTeams = linear.TeamKeys(repoPath)
 		}
+		// Same goroutine, same reason — and unlike the worktree list this one is
+		// allowed to fail quietly: no branches simply means the Base branch field
+		// has no suggestions, never a dialog that refuses to open.
+		branches, branchErr := git.ListBranches(repoPath)
```
hayke102: This runs on the custom-provider path too, where the result is thrown away - `workspaceListMsg`'s handler returns early for `msg.provider.IsCustom()` straight to `createWorkspaceDialog`. For shell-provider repos `repoPath` also isn't normalized to a main worktree and needn't be a git repo at all, so it's two guaranteed-failing subprocesses (`for-each-ref` + `rev-parse`) added to the spinner the user waits behind on every `w`.

`originKey` a few lines up is already guarded with `if !provider.IsCustom()`; the same guard here would match.

Worth noting that `GetDefaultBranch` on the line above has exactly the same problem and predates this branch - so this is only half a fix unless both move inside the guard.

internal/ui/workspace_picker_branch.go
```diff
@@ -0,0 +1,151 @@
...
+	// back answers a question nobody asked — Enter on it is a no-op. The useful
+	// question there is "what else is there", so fall back to the unfiltered
+	// list. This matters because the field is PRE-FILLED with origin/<default>:
+	// without it, focusing the field for the first time shows a single row
+	// repeating what is already on the line above it, and browsing would mean
+	// clearing a field you probably wanted to keep.
+	if len(d.branchMatches) == 1 && d.branchMatches[0] == text {
+		// wantRemote is deliberately kept: you asked for a remote ref, so the
+		// wider list is the other remote refs.
+		d.branchMatches = d.matchBranches("", wantRemote)
```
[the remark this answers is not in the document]
hayke102: Fixed — the widened list no longer offers back the value already in the field. Right call, and it contradicted this feature's own rationale: the widening exists precisely because a row echoing the field is a no-op on Enter, and then it put that row back at position 0.

Implemented as an `exclude` parameter on `matchBranches` rather than the suggested post-filter over its result: filtering afterwards drops the list to 4 rows where 5 fit, since the cap has already been applied. `TestBaseBranchExclusionRunsBeforeTheCap` pins that — I checked it fails against the diff as proposed.

internal/ui/workspace_picker.go
hayke102: Fixed — the existing-worktrees list is now hidden while the base field carries the highlight.

Confirmed your numbers by probe (21 box lines at rest, 26 focused, on 80x24 with 6 worktrees), so this fits before the change and overflows after — my +5 rows are what tip it over, and the footer is first off the bottom.

Took the second option. Capping against remaining height would have made the *cursor* clamp depend on terminal size, since `setSelection` runs nowhere near `View` — and nothing else in this file works that way. Dropping the list reclaims 8 rows against the 5 added, so the focused dialog is now strictly *shorter* than the resting one, which is a stronger property than not overflowing. Hidden is not unreachable: `shift+tab` still cycles onto the list and brings it back.

`TestWorktreeFocusedDialogIsNoTallerThanAtRest` asserts it at 80x24 — the gap you identified, since `TestWorktreeDialogRowsNeverOverflow` is width-only and runs at height 40.

The pre-existing unboundedness you note (the worktree list on its own, `renderTicketBlock`) is untouched and still real.

internal/git/git.go
hayke102: Fixed — `CommitDate` is now `max(local, remote)` when a counterpart exists, and the list is re-sorted after the merge (for-each-ref's order stops being authoritative once a date is revised).

Verified on the existing fixture before fixing: `lagging` came out at 2020-01-01 and sorted below `ghost`, exactly as you described.

One thing worth adding — this is a regression I introduced rather than a latent flaw. The old single-pass form emitted *both* entries, so the newer position existed by accident of the duplicate; fixing the duplicate is what removed it.

`TestListBranchesReportsTheLaterOfTheTwoRefs` now asserts both the date and the ordering, which is the gap you spotted in the flags-only test.

internal/ui/workspace_picker.go
hayke102: Fixed — `tab`/`shift+tab` now cycle.

Agreed on the distinction: *inert* isn't the documented cost. `tab` no longer walking the list stands, but a key documented as "next field" doing nothing at all on the last field is a dead key.

Wrapped both directions rather than just forward — `shift+tab` from the base field was inert in the same way, so wrapping only one way would have moved the dead end rather than removed it. With no worktrees the cycle closes over the two fields that exist. Comment and CLAUDE.md now say cycle. `TestWorktreeTabCyclesBothWays` walks the full loop each way.

CLAUDE.md
hayke102: Fixed the gate rather than the sentence — you're right that the claim was wrong, and reproducing it changed my read of how much it mattered.

Probe on `[master, topic-a…d]`:
```
"mast"    -> 1 row  [master]
"master"  -> 5 rows [master topic-a topic-b topic-c topic-d]
"master-" -> 0 rows []
```

One amendment to "cosmetic in practice": the highlight indeed doesn't move, but the dialog is vertically **centred**, so the box jumps up 4 rows at the `r` and down 5 at the `-`. That's a visible judder mid-word, which is what pushed me to your second option over softening the doc.

There's now a `baseAtRest` flag — set by `Show` and `pickBaseBranch`, cleared by the first typed change. "At rest" means what the word says, the judder is gone, and the CLAUDE.md sentence is true rather than softened. `TestBaseBranchTypingNeverWidens` walks that exact keystroke sequence; `TestBaseBranchSettlesOnOpenAndOnPick` covers the other side.

internal/ui/app.go
hayke102: Fixed — both moved inside the existing `!provider.IsCustom()` guard.

Confirmed the early return at `app.go:1538`: the custom path goes straight to `createWorkspaceDialog` and reads neither `msg.branches` nor `msg.defaultBranch`.

Took your "only half a fix" point and moved `GetDefaultBranch` too. Guarding one and leaving the other still forks a doomed subprocess on every `w` in a shell-provider repo, so half of it buys nothing measurable — and the two now sit together under the guard that was already there for `originKey`.

PR #279, opened 2026-08-25
feat(jira): start worktrees from Jira tickets, alongside Linear

### LINKED ISSUE #272: Add support for Jira in addition to the currently supported…
### Problem
Add support for Jira in addition to the currently supported Linear

### Environment
- **Version**: 2.30.3
- **OS**: macOS 26.2 (arm64)

hayke102:
Closes #272.

Type a Jira key in the `w` dialog and fleet names the branch, writes the ticket and its screenshots into the worktree, moves the issue to In Progress, and opens the agent already told to read it — the same gesture that already worked for Linear. `t` shows one merged list across both trackers, and `fleet wt --ticket OPS-42` works from the shell.

```
Ctrl+K → "Connect Jira"        site + Atlassian email + API token
.fleet.local.json              {"jira": {"project": "OPS"}}
w → type OPS-42                → branch ops-42-<slug>, ticket materialized
```

## Layering

Four packages, and the domain package imports no provider — that is what keeps it acyclic.

| package | role |
|---|---|
| `internal/ticket` | the `Provider` interface, and everything that was never tracker-specific: branch naming, the seed prompt, the image downloader, the keychain `Store`, `ticket.md`, the `meta.json` ledger, `Materialize` |
| `internal/linear` | slimmed to a GraphQL client implementing `Provider` |
| `internal/jira` | **new** — REST v3 client, Basic auth, ADF → markdown |
| `internal/ticketing` | the router the UI and `cmd/fleet` call; the one place that knows there is more than one tracker |

A provider is a zero-size value over package-level state rather than a struct with fields, because that state has to be reachable from `Available()` — which the Bubble Tea Update goroutine calls, and which therefore may not do any work.

## Decisions worth reviewing

**Jira Cloud only, enforced by the API version rather than the hostname.** v3 is what serves ADF and exposes `/rest/api/3/search/jql`; Server and DC speak v2, return wiki markup, and take a bearer PAT. So `NormalizeSite` accepts any host with a dot (Atlassian now serves Cloud on customer-owned domains) and a Server site fails `VerifyCredential` with the 404 the dialog turns into *"fleet supports Jira Cloud only"* — the one place that diagnosis can honestly be made.

**API token, not OAuth.** Atlassian's 3LO needs an app brizz would register and own forever, and in many organizations a per-site admin approves each install — for a credential a user mints themselves in two clicks. It also covers the SSH/CI case Linear needs its paste path for anyway.

**Every image attachment is downloaded, not only the referenced ones.** An ADF media node's `attrs.id` is a Media Services file id and is *not* the attachment id the REST list is keyed by ([JRACLOUD-96383](https://jira.atlassian.com/browse/JRACLOUD-96383)); alt text is the only bridge and Jira frequently writes none. Inline-only matching would silently drop the screenshot on exactly the bug reports that are mostly screenshot. Non-images are named, never fetched.

**The started transition is picked by name, preferring `In Progress`.** Jira has no equivalent of Linear's state position, and every in-progress-ish status shares the `indeterminate` category — so a workflow with both *In Progress* and *In Review* offers two candidates and the wrong one moves a fresh ticket straight to review.

**Error classification is per-tracker and the two are opposites.** Linear answers an unknown issue with HTTP 200 plus an `errors[]` entry; Jira answers with a real 404. Each package now has a test recording which way round its API works, so the asymmetry is pinned rather than rediscovered.

## Guards that were about to go vacuous

Two existing tests would have kept passing while covering nothing once a second provider existed:

- `TestNoTicketSubprocess` (was `TestNoLinearSubprocess`) scanned one directory — `internal/jira` could have shelled out to anything. It now scans all four ticket packages against the same allowlist.
- `TestTicketWorkStaysOffTheWorkers` string-matched the literal `"linear"` — a worker calling `jira.Fetch` or `ticketing.Assigned` would have sailed through the guard that exists to stop exactly that.

A third, unrelated to Jira: `TestCredentialResolutionOrder` was reading the **developer's real keychain**. It passed in CI and failed on any machine actually connected to Linear — which is every machine where someone would run the suite while changing this code. It is now isolated behind a namespaced `Store`.

## Behaviour changes to existing surfaces

- **The worktree dialog is now a pure function of what it was handed**, not of the global registry: `Show` takes `[]ticketing.Bound`. That makes it testable, and it is why `ticketsOff` is keyed by provider — a rejected Jira token must not silence the Linear suggestions in the same repo. A fan-out search failure latches nothing, since it names no single provider.
- **`ticket.Priority` keeps Linear's numbering on purpose** (`0` unset, `1` urgent … `4` low), so the palette's gauge and colour ladder need no translation and a Jira row renders identically to a Linear one. Jira maps by priority *name*, since schemes are per-project; anything unrecognized is `None`, not `Medium` — a guessed middle rank would outrank a real `Low`, and the list is sorted on this.
- **Sentinels lost their `linear:` prefix.** They used to render as `Linear: linear: issue not found`, so every caller hand-worded its message instead of formatting the error.
- `linear_ticket_start` → **`ticket_start_state`**, superseded rather than migrated: the old key is still read when the new one is unset, so an existing config keeps working and nobody's file gets rewritten.
- Palette: *"My Linear Tickets"* → *"My Tickets"*; *"Connect Jira"* added. The `connect_linear` tip **id** is deliberately unchanged — it is persisted in `seen_tips` and counted in `feature_usage`, so renaming the string would re-show a tip every existing user has dismissed.

## ⚠️ Coordinated change needed

Analytics events are renamed: `linear_ticket_materialized` → `ticket_materialized` and `linear_command_failure` → `ticket_command_failure`, each carrying a `provider` property. A per-provider event name makes *"how often does ticket fetching fail"* a query that has to be rewritten every time a tracker is added — but **the `fleet-analytics` dashboard references the old names** and needs a matching change.

## Verification

`go test -race ./...` green, `golangci-lint` 0 issues.

Because I have no Jira credentials, the live pass is unrun — so the HTTP layers are covered by fake servers rather than mocks:

- `internal/jira/api_test.go` drives fetch, JQL search and assignment, the truncated-comment refetch, transitions and `VerifyCredential` against an `httptest` TLS server speaking Jira's REST API, through the real client and the real auth header.
- `internal/ticket/images_download_test.go` drives a real authenticated download, extension recovery from magic bytes, placeholder rewriting, and the degrade-to-a-sentence path for a failed one.
- Plus ADF golden cases, JQL escaping, site normalization, priority/state mapping, the credential-exfiltration gate, and router fan-out.

**Still unverified: a real Atlassian tenant answering.** `internal/ticketing/materialize_e2e_test.go` covers it and is opt-in:

```
JIRA_SITE=acme.atlassian.net JIRA_EMAIL=you@example.com JIRA_API_TOKEN=ATATT… \
  FLEET_JIRA_E2E=OPS-42 go test ./internal/ticketing/ -run EndToEnd -v
```

https://claude.ai/code/session_01A1JeKutELhvguheMczEPH9

## Summary by CodeRabbit

* **New Features**
  * Added Jira Cloud ticket support alongside Linear.
  * Connect Jira using site, email, and API token credentials.
  * Search assigned tickets across supported trackers.
  * Configure Jira projects per repository and create ticket-linked branches or worktrees.
  * Render Jira descriptions, comments, attachments, subtasks, links, and images in Markdown.
  * Added base-branch suggestions with navigation and filtering.
  * Added session-scoped model and effort overrides for launches.
* **Improvements**
  * Automatically update ticket status when work begins.
  * Strengthened credential and attachment security, including Jira token redaction.


internal/ticketing/ticketing.go
```diff
@@ -0,0 +1,291 @@
...
+	p, ok := Owner(repoPath, id)
+	if !ok {
+		return ticket.Ticket{}, ticket.ErrNotConnected
+	}
+	return p.Fetch(ctx, id)
+}
+
+// Materialize writes a ticket into a worktree through the provider that owns it.
+func Materialize(ctx context.Context, repoPath string, o ticket.Opts) (ticket.Result, error) {
+	p, ok := Owner(repoPath, o.Identifier)
```
hayke102: **Picking a search suggestion outside the repo's keys silently creates a worktree with no ticket.**

`jira.Search` is deliberately unscoped by project (`api.go:395` says so, and Linear's `searchQuery` is the same), so in a repo that names only `BRZ` the `w` dialog will happily offer `PRD-45`. `pickTicket` sets `d.resolved` with no key check, and `ticketForCurrentInput` hands it to `workspaceCreateMsg`. Then `Materialize` re-resolves through `Owner(repoPath, id)` — which gates on the repo's configured keys — and returns `ErrNotConnected`, which `ticketStatusLine` (`internal/ui/ticket.go:96`) deliberately renders as `""`.

Net effect: branch `prd-45-login-crash` gets created, no `.fleet/ticket/` dir, no seeded prompt, and nothing on screen says why. This worked on master — `linear.Materialize` had no repo gate at all — so it's a regression introduced by routing through the router.

Three ways out: scope search results to the bound keys, carry the owning `ticket.Provider` on the picked ticket so materialize doesn't re-resolve, or at minimum stop swallowing `ErrNotConnected` here. Worth noting the comment at `ticket.go:94` ("barely reachable... the dialog's own fetch had to succeed") stopped being true the moment the router became a second source of that sentinel.

internal/jira/provider.go
```diff
@@ -0,0 +1,274 @@
...
+	path := "/rest/api/3/issue/" + url.PathEscape(key) + "?fields=" + fullFields
+	if err := doWith(ctx, cred, ticket.FullTimeout, http.MethodGet, path, nil, &iss); err != nil {
+		return nil, err
+	}
+	if iss.Key == "" {
+		return nil, ticket.ErrNotFound
+	}
+
+	comments := iss.Fields.Comment
+	if comments != nil && comments.Total > len(comments.Comments) {
```
hayke102, line 85: **This re-fetch can lose comments, and it fetches the oldest ones — the opposite of the stated intent.**

Two problems in the same three lines:

1. `orderBy=created` is ascending; descending is `-created`. So the replacement page is the *oldest* 50, while the doc comment above argues the extra round trip is worth it because "the newest comment is usually what changes what a ticket means".
2. `comments = &page` replaces the whole set unconditionally. With `Total=120` and 100 embedded in the issue response, 100 comments are discarded and replaced by 50.

Suggest `orderBy=-created` plus only replacing when the new page actually carries more than what was embedded.

hayke102 (commented):
Findings from `/code-review high` on this branch. All six verified against the code; the `ticketing.go` one is the behavioural regression vs master.

internal/jira/provider.go
```diff
@@ -0,0 +1,274 @@
...
+	// drop the screenshot on exactly the bug reports that are mostly screenshot.
+	// Downloading everything and listing it under ## Attachments costs at worst
+	// an unreferenced image inside caps fleet already enforces.
+	var images []ticket.Image
+	byName := map[string]int{}
+	for _, a := range iss.Fields.Attachment {
+		if !a.isImage() || a.Content == "" {
+			continue
+		}
+		byName[strings.ToLower(a.Filename)] = len(images)
```
hayke102: **Duplicate attachment filenames collapse here, which drops one image and renders the other twice.**

Jira allows two attachments on one issue both named `image.png`. `byName[lower(filename)]` overwrites, so both names resolve to index 1 and index 0 becomes unreachable.

Downstream: `renderAttachments` (:248) iterates `atts`, so it emits `` twice and never emits `fleet-image:0`. `collectImages` then skips index 0 entirely (`!strings.Contains(body, ref)` → `continue`), so the first screenshot is never downloaded and the second appears twice.

Given the comment right above says everything is downloaded precisely so screenshot-heavy bug reports don't lose images, this is the case it was meant to cover. Keying on attachment id, or letting `byName` hold a slice and consuming entries as media nodes match, would fix it.

internal/workspace/repo_config.go
```diff
@@ -124,11 +151,31 @@ func CopyFilesPatterns(repoPath string) []string {
...
+// JiraProjectKeys returns the merged, upper-cased Jira project keys for a repo,
+// or nil when the repo names none. Nil means "this repo is not Jira-tracked".
+func JiraProjectKeys(repoPath string) []string {
+	cfg := loadMergedRepoConfig(repoPath).Jira
+	return trackerKeys(cfg.Project, cfg.Projects)
+}
+
+// JiraSite returns the repo's Jira site override, normalized to lower case, or
+// "" when it names none — in which case the credential's own site is used.
+func JiraSite(repoPath string) string {
```
hayke102: **`JiraSite` is never called outside its own test — the documented per-repo site override does nothing.**

CLAUDE.md documents `{"jira": {"site": "acme.atlassian.net"}}` as "optionally overrides the credential's own site, for a repo belonging to a different Jira". But `jira.credential()` always returns the stored/env site, `Document.Host` closes over `cred.Site`, and `baseURL()` is built from it. A user who follows that line gets their requests sent to the *other* site with no error anywhere — the key is silently accepted and ignored.

Either wire it through `credential()`/`Document`, or drop the field and the doc line.

internal/ticket/images.go
```diff
@@ -0,0 +1,198 @@
...
+// unnamed and unextensioned — and an agent's file-read tool dispatches on
+// extension, making a perfectly downloaded screenshot unreadable. Recovering the
+// extension is the difference between "we fetched it" and "the agent can see it".
+//
+// http.DetectContentType sniffs magic bytes, so it is the backstop when the
+// response headers say nothing useful. Jira does send a filename, but the sniff
+// still runs as a check rather than a rescue: it is what stops a 401 HTML page
+// landing beside real screenshots under a .png name the server volunteered.
+func detectExt(name string, body []byte) (string, bool) {
+	if ext := strings.ToLower(filepath.Ext(name)); ext != "" {
```
hayke102: **The magic-byte sniff never runs when a filename or header already supplies a known extension, so it isn't the check the comments claim.**

This early return means `http.DetectContentType(body)` is only consulted when `filepath.Ext(name)` is empty or unrecognized. And `fetchImage` (:161) tries `extFromHeaders` before falling back to `detectExt` at all.

For Jira, `alt` is always the real filename *and* `Content-Disposition` is always sent, so both short-circuits fire on every attachment and the body is never inspected. A 200 carrying an HTML interstitial for `screenshot.png` lands on disk as `1-screenshot.png`.

The doc comment here and the CLAUDE.md line both assert the sniff "still runs as a check rather than a rescue". Either sniff unconditionally and let the bytes win, or fix both comments — right now they describe a defence that isn't there.

internal/jira/jira.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// escapeJQL quotes a user's search term for a JQL string literal.
+//
+// JQL string literals take backslash escapes, so a term containing a quote
+// would otherwise end the literal and turn the rest of what someone typed into
+// syntax. Only these two characters need it — everything else inside quotes is
+// data, including the reserved words that would break an unquoted term.
+func escapeJQL(term string) string {
+	r := strings.NewReplacer(`\`, `\\`, `"`, `\"`)
```
hayke102: **Only `\` and `"` are escaped, but the value lands in a Lucene `text ~` clause.**

Atlassian documents `+ - && || ! ( ) { } [ ] ^ ~ * ? :` as reserved in text searches, and this fires on any ≥3 runes of prose typed into the New branch field. So `fix (login` or a trailing `?` produces a 400, which the dialog surfaces as "Jira: unavailable" rather than "no matches" — an error about the tracker for what is really a syntax artifact of the user's own typing.

Escaping (or stripping) the reserved set would make prose search robust.

internal/ticketing/ticketing.go
hayke102: Fixed — confirmed as a regression against master. `ticket.Opts` now carries `Provider` and `ticketing.Materialize` prefers it, falling back to `Owner(repoPath, id)` only for a bare identifier nobody has claimed yet (the `fleet wt --ticket` path). The root cause was the double gate you identified: the provider was already decided when the dialog fetched the ticket, and Materialize re-decided it by a different rule. Took the second of your three options rather than scoping search, since the repo gate's job is deciding whether ticket surfaces appear at all, not policing which issue you may work on. Also stopped swallowing ErrNotConnected in ticketStatusLine — you're right that the 'barely reachable' comment stopped being true, and at that call site it means the credential went away mid-flight, which is the same argument the code already makes for ErrNotFound. Pinned by TestMaterializeHonoursTheTicketsOwnProvider.

internal/jira/provider.go
hayke102: Fixed — both halves. Now `orderBy=-created`, and the page replaces the embedded set only when it is actually larger. Added one thing your comment implied but didn't say outright: the page has to be reversed back to chronological before rendering, or ticket.md would read newest-first for Jira and oldest-first for Linear. The existing test fixture returned ascending comments, so it had to be corrected to match the request too — otherwise the reversal couldn't be tested at all. Covered by TestCommentRefetchNeverShrinksTheSet plus an ordering assertion in TestDocumentFetchesMoreCommentsOnlyWhenTruncated.

internal/jira/provider.go
hayke102: Fixed — and you're right that this was the exact case the comment above it claims to protect. Two indexes now, because they answer two different questions: renderAttachments looks up by attachment id (unique), so each attachment emits its own placeholder exactly once; byName became map[string][]int, a queue, and each media node claims the next unclaimed index for that name. One extra case fell out of it — the same screenshot referenced twice in one description — where the resolver falls back to the first index once they're all claimed, since replacing a present image with 'see Attachments' would be worse than showing it twice. TestDuplicateAttachmentNamesEachGetTheirOwnImage and TestRepeatedReferenceToOneAttachmentStillRenders.

internal/workspace/repo_config.go
hayke102: Fixed by dropping it — took the second of your two options. `Site` is gone from JiraConfig, `JiraSite` and its test case are deleted, and the CLAUDE.md sentence is replaced with a note saying there deliberately isn't one. Wiring it through would need its own credential story rather than just plumbing: an Atlassian API token is scoped to an account rather than a site, so a repo pointing at a second Jira has no guarantee the same token reaches it, and credential() is deliberately repo-unaware today. A key that parses cleanly and is silently ignored is worse than one that doesn't exist. My fault on both sides — I wrote the field and the doc line and never connected them.

internal/ticket/images.go
hayke102: Fixed — and the overclaim was mine, in both the doc comment and CLAUDE.md. The bytes decide now: detectExt takes no filename argument at all, and fetchImage sniffs before it reads any header. One nuance worth recording, since a naive bytes-only rule would have broken valid files: Go's sniffer has no SVG rule (`<?xml` reads as text/xml, a bare `<svg>` as text/plain), so an `image/svg+xml` header is honoured only when the body sniffs as xml or plain text — which still refuses an HTML interstitial, since that sniffs as text/html whatever the headers claim. You were also right that the old TestFetchImageRejectsNonImages passed for the wrong reason: its fixture sent a wrong Content-Type too, so your scenario was genuinely uncovered. TestBytesBeatHeaders covers all four directions now.

internal/jira/jira.go
hayke102: Fixed — two-stage escaping, Lucene then the JQL literal. My comment asserting 'everything else inside quotes is data' was true of an `=` comparison and false of `~`, which is exactly the trap you named. Escaped rather than stripped, so `C++` still searches for C++ rather than C. The existing expectations in TestEscapeJQL and TestSearchQuotesTheTerm were written against the single-stage version and are updated; the search test now also asserts that no character in the reserved set reaches the query unescaped.

internal/linear/api.go
```diff
@@ -293,29 +276,44 @@ type issueLite struct {
...
+		State:    ticket.StateOther,
+		StatePos: 1 << 30,
 	}
-	t := Ticket{Identifier: i.Identifier, Title: i.Title, URL: i.URL, Priority: i.Priority}
 	if i.State != nil {
-		t.StateName, t.StateType = i.State.Name, i.State.Type
+		t.StateName, t.State, t.StatePos = i.State.Name, stateType(i.State.Type), i.State.Position
 	}
 	return t
 }
```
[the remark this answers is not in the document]
hayke102: Fixed — and confirmed rather than reasoned about: a standalone program with the same struct shapes prints `outer: 2, embedded: 0`. Removed the duplicate `Priority` from issueFull so the embedded field is the only target. Worth recording that this was a regression from my refactor: master read `priorityName(i.Priority)` off the *issueFull directly, which is the field that gets the value; moving the read to the projection sent it to the shadowed zero, so every Linear ticket.md has been missing its front-matter priority. Jira is unaffected (no embedding) and so is the tickets tab (that path decodes issueLite directly). TestFullDecodeKeepsPriority asserts through the projection rather than the struct field, since reading the field directly is exactly what made this invisible.

internal/ui/worker_cadence_test.go
```diff
@@ -520,14 +521,23 @@ func TestStatusWorkerFeedsHookChangesIntoPriority(t *testing.T) {
...
+	packages := []string{"linear", "jira", "ticketing", "ticket."}
 	for _, fn := range []string{"refreshAllGitAndPR", "gitWorkerCycle", "statusWorkerCycle"} {
-		if mentions(t, fn, "linear") {
-			t.Errorf("%s reaches into internal/linear — a network call there blows the "+
-				"~70s per-repo budget that workerStallThreshold (90s) is sized against", fn)
+		for _, pkg := range packages {
+			if mentions(t, fn, pkg) {
+				t.Errorf("%s reaches into internal/%s — a network call there blows the "+
+					"~70s per-repo budget that workerStallThreshold (90s) is sized against",
+					fn, strings.TrimSuffix(pkg, "."))
```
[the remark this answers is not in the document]
hayke102: Fixed — dot dropped. You're right about the AST: mentions() walks for *ast.Ident and compares Name, so "ticket." could never match while the other three entries matched fine as receiver idents. Doubly worth catching because I widened this guard in the previous review round and wrote a comment claiming it names every ticket package. Verified the fix properly rather than just re-running it green: temporarily added a `ticket.TicketDir` call to refreshAllGitAndPR and watched the guard fail with the right message, then reverted. A local variable named `ticket` will also trip it now, which is the correct answer here and is why the create path's local is named `tkt`.

cmd/fleet/worktree.go
```diff
@@ -615,9 +641,24 @@ func accountAllowed(email string, allowed []string) bool {
...
+	key, _, _ := strings.Cut(strings.ToUpper(strings.TrimSpace(id)), "-")
+
+	if p, ok := ticketing.ClaimedBy(repoPath, id); ok {
+		// The repo names this key; the tracker just is not connected.
+		return fmt.Sprintf("This repo tracks %s in %s, but %s isn't connected.\n"+
+			"Connect it in the TUI (Ctrl+K → Connect %s), or set its environment variables.",
+			key, p.Name(), p.Name(), p.Name())
+	}
+	return fmt.Sprintf("No tracker in this repo claims %s.\n"+
+		"Name its team or project in .fleet.local.json, e.g. {\"jira\": {\"project\": %q}}", id, key)
```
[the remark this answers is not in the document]
hayke102: Fixed — both forms now. The point that made it worth doing: this branch fires when *no* provider claims the key, so the reader is as likely to be a Linear user as a Jira one, and this line is the only thing telling them what to write. Following a Jira-only example puts a Linear team key under a `jira` block, which then fails a second time and less legibly than the first.

internal/ui/statusreport.go
```diff
@@ -398,6 +400,20 @@ var emailPattern = regexp.MustCompile(`[\w.+-]+@[\w-]+\.[\w.-]+`)
...
+var jiraTokenPattern = regexp.MustCompile(`ATATT[A-Za-z0-9_.\-=]+`)
+
+// basicAuthPattern is the backstop the prefix rules cannot be.
+//
+// Jira's credential travels as base64(email:token) in an Authorization header,
+// where the ATATT prefix is encoded away — so a header that reached a log line
+// or a pane excerpt would survive every other rule here intact. The floor of 20
+// characters keeps it off the words "Basic" and "Basic auth" in prose, since
+// the capture must look like base64 to the end of the run.
+var basicAuthPattern = regexp.MustCompile(`Basic\s+[A-Za-z0-9+/]{20,}={0,2}`)
```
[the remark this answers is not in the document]
hayke102: Fixed — `(?i:Basic)`, plus a lowercase fixture in TestBuildStatusReportBody_NeverLeaksJiraToken. fleet's own code always writes `Basic `, but what this pattern actually scans is a pane excerpt: arbitrary terminal text, including a curl someone typed with a lowercase header. RFC 7235 makes the scheme case-insensitive, and the comment two lines up already states that over-matching is the safe error in a redactor — this is that rule applied.

internal/ui/connect_jira.go
```diff
@@ -0,0 +1,420 @@
...
+	if !ok {
+		return d, nil
+	}
+
+	switch d.stage {
+	case connectWorking:
+		if keyMsg.String() == "esc" {
+			d.Hide()
+		}
+		return d, nil
```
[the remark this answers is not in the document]
hayke102: Fixed — and the same hole existed next door, which is worth flagging since you only looked at this file. The dialog now owns the verification context and cancels it from Hide, and the command re-checks ctx between the round trip and the write, so a verification that completes after an escape stores nothing. Cancellation is filtered out of the error path too, since reporting "rejected — check the email and token" for something the user chose is the dialog arguing with them. **Linear's paste path had the identical bug**: abortSignIn only ever covered the loopback listener for the browser flow, so verifyAndStoreLinearKey ran on context.Background() and escaping mid-verify stored the key. Both are fixed, with a test each.

internal/ui/workspace_picker_ticket.go
```diff
@@ -225,18 +295,29 @@ func (d *WorktreeDialog) applyTickets(m worktreeTicketsMsg) {
...
+		if !known || len(acct.Keys) == 0 {
+			return "", false
+		}
+		for _, repoKey := range b.Keys {
+			for _, acctKey := range acct.Keys {
+				if strings.EqualFold(repoKey, acctKey) {
+					return "", false
+				}
 			}
 		}
```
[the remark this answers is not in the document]
hayke102: Fixed — workspaceMismatchNote now takes the owning provider, and the byID path passes it while the fan-out search still passes nil. Concretely what you caught: a repo tracking Linear BRZ and Jira OPS, with the Jira credential on the wrong site, would find BRZ in Linear's account, bail out early, and report "OPS-42 — no such issue" — which is the exact misdiagnosis this note exists to prevent, reintroduced by the multi-provider loop. One thing that fell out of scoping it: the note also had to stop joining `d.ticketKeys`, since an owner-scoped note listing another tracker's keys would be describing the wrong workspace. Covered by TestMismatchNoteForAnIdentifierChecksOnlyItsOwner, which asserts all three directions — owner mismatched, owner fine, and the search staying silent.

changelog/unreleased/jira-tickets.md
```diff
@@ -0,0 +1,6 @@
+---
+type: added
+highlight: true
+---
+
+**Jira, alongside Linear.** `Ctrl+K` → "Connect Jira" with your site, email and an API token, add `{"jira": {"project": "OPS"}}` to a repo's `.fleet.local.json`, and typing `OPS-42` in the `w` dialog now names the branch, writes the ticket and its screenshots into the worktree, moves the issue to In Progress, and opens the agent already told to read it. `t` shows one merged list across both trackers.
```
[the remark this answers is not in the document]
hayke102: Fixed on length — you're right that it opened with setup steps and enumerated five outcomes in one sentence, against the repo's own rule. Took a different wording than suggested, though: `changelog/README.md` asks for "honest specificity, not marketing" and bans the feature-list tone, so "Connect Jira, configure a repo project, and create worktrees from Jira tickets" would trade one guideline violation for another. Now two sentences that lead with the change and keep one concrete detail:

**Jira tickets, alongside Linear.** Type a Jira key in the `w` dialog and fleet names the branch, pulls the ticket and its screenshots into the worktree, and opens the agent already told to read it. `t` lists both trackers at once.

`highlight: true` stays — this one leads the What's New reel.

PR #283, opened 2026-08-26
fix(status): repair hooks pointing at a deleted binary, and stop one frame ending a run

hayke102:
Diagnosed from a `D` snapshot where a session showed **finished** while `✽ Sock-hopping… (54s · ↓ 2.9k tokens)` was on screen. Two independent faults stacked; the first is fleet-wide and much bigger than the one session.

## 1. Hooks were dead across the whole fleet

`snapshot.json` had an empty `hook` block and the log repeated `no hook data, falling back to pane capture` 48 times. The hook file was never written — and neither was anyone else's:

```
live fleet sessions WITH hook file: 17   (newest write ~19h old)
live fleet sessions WITHOUT hook file: 20
```

Every fleet hook in `~/.claude/settings.json` pointed at `…/brizz-code-base-branch/build/fleet`, a worktree that had since been removed. The hook fires, the command doesn't resolve, nothing is written, and every session silently falls back to pane capture with nothing anywhere reporting it. Routing was fine — `FLEET_INSTANCE_ID` was correctly set.

How it happens: hooks record `FleetBinaryPath()`, which in a dev checkout is a worktree's `build/fleet`. A sibling instance launched from another worktree rewrites that path at launch, the worktree is later deleted, and the long-running instance never re-checks. Homebrew hits the same shape across an upgrade (the versioned Cellar path goes away). The auto-updater is *not* implicated — it renames in place, so its path stays valid.

`RepairClaudeHooks` heals only a path that **doesn't resolve**, never one that merely differs from ours. That distinction is why it exists separately from `InjectClaudeHooks`, which also rewrites a differing path — correct once at launch, ruinous on a timer, since two instances from different worktrees would overwrite each other's `settings.json` every cycle. A path that resolves works whoever wrote it (`hook-handler` writes into one global status-file dir), so *missing* is the only condition worth acting on unprompted. Pinned by `TestRepairClaudeHooksLeavesAResolvablePathAlone`.

Runs throttled (60s) on the worker's heavy pass. A `tipRecurring` fires only after an actual repair and names how many sessions still need restarting, since a running agent keeps the settings it launched with.

## 2. One repaint frame ended a run

With hooks gone, a single frame decided everything:

```
12:38:24.150  matched spinner char  char=✳  line="✳ Sock-hopping… (33s · ↓ 1.6k tokens · thinking with xhigh effort)"
12:38:24.683  matched prompt        line="❯ /ren"  linesFromBottom=4
12:38:24.683  status changed (pane) old=running new=finished detected=finished
              ← 18 seconds of silence →
12:38:42.868  spacejump: landed     targetID=9a5508ce-1787736161
```

The user was typing a slash command while the agent worked. On that repaint the activity line wasn't drawn, so `detectRunning` missed and `detectFinished` matched the prompt. Being finished then dropped the session off the ~500ms fast pass onto the round-robin (5 per 2s over 69 sessions), where the wrong status stood for 18s — and `Space` jumped the user onto a session that was busy.

**The idle-prompt reading is deliberate and stays.** It is the only signal that clears a stale `waiting` hook when the user escapes a permission prompt and starts typing in default mode, where there's no `⏵⏵` mode bar to fall back on (#270). I spiked requiring an empty prompt remainder: all 21 golden fixtures pass, but it breaks `TestScenarioNBSPPromptClearsStaleWaiting`. The same pane shape must clear a stale waiting *and* must not demote a live running, and only the prior state separates them — so the guard belongs at the transition. `detectStatus` is untouched.

`updateStatusFromPane` now needs 2 consecutive finished readings before demoting a running session. Costs one extra pass (~500ms) before a genuine finish shows.

## Tests

- `TestScenarioTypedPromptBlinkDoesNotDemoteRunning` — reproduces the captured frame; fails `expected "running", got "finished"` without the guard.
- `TestScenarioSustainedIdlePromptStillFinishes` — the counterweight: the guard delays a demotion by one pass, never prevents one.
- `TestFleetHookBinary`, plus three `RepairClaudeHooks` tests covering heal / leave-alone / no-settings-file.

`make lint` clean, full suite green.

https://claude.ai/code/session_01UbntCtmZonJDvwaQadprEz

## Summary by CodeRabbit

- **Bug Fixes**
  - Claude hooks now automatically repair themselves when the referenced Fleet binary is missing.
  - Fleet periodically checks for broken hooks without overwriting working configurations.
  - Session status no longer changes to finished because of a single blinking or incomplete frame.
  - Finished sessions require multiple spaced readings for more reliable status detection.

- **User Guidance**
  - Added a notification identifying sessions that should be restarted after hook repair.

### DIFF
--- a/changelog/unreleased/hooks-repair-deleted-binary.md
+++ b/changelog/unreleased/hooks-repair-deleted-binary.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**Status survives a deleted binary.** Hooks pointed at the exact `fleet` binary that installed them, so removing that git worktree — or upgrading past that Homebrew version — silently killed status detection for every session while screen scraping quietly stood in. fleet now re-points them within a minute and names the sessions you need to restart.
+**Status stops going dark.** Every session could lose status detection at once, silently, if the `fleet` binary that installed the hooks was later deleted or upgraded away — the sidebar kept showing a status, just a guessed one. fleet now repairs the hooks within a minute and names the sessions to restart.
--- a/changelog/unreleased/hooks-repair-deleted-binary.md
+++ b/changelog/unreleased/hooks-repair-deleted-binary.md
@@ -1,0 +1,5 @@
+---
+type: fixed
+---
+
+**Status survives a deleted binary.** Hooks pointed at the exact `fleet` binary that installed them, so removing that git worktree — or upgrading past that Homebrew version — silently killed status detection for every session while screen scraping quietly stood in. fleet now re-points them within a minute and names the sessions you need to restart.
--- a/changelog/unreleased/typed-prompt-false-finished.md
+++ b/changelog/unreleased/typed-prompt-false-finished.md
@@ -1,0 +1,5 @@
+---
+type: fixed
+---
+
+**Typing no longer marks a session done.** Starting a message while the agent was still working could flip it to finished for ~18s — long enough for `Space` to jump you to a session that was busy. A single frame can no longer end a run.
--- a/changelog/unreleased/typed-prompt-false-finished.md
+++ b/changelog/unreleased/typed-prompt-false-finished.md
@@ -2,4 +2,4 @@
 type: fixed
 ---
 
-**Typing no longer marks a session done.** Starting a message while the agent was still working could flip it to finished for ~18s — long enough for `Space` to jump you to a session that was busy. A single frame can no longer end a run.
+**Busy sessions stay busy.** Starting a message while the agent was still working could mark it finished for ~18s — long enough for `Space` to jump you onto a session that was mid-run.
--- a/internal/hooks/claude_hooks.go
+++ b/internal/hooks/claude_hooks.go
@@ -2,7 +2,9 @@ package hooks
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -446,10 +448,18 @@ func hookBinaryMissing(hooks map[string]json.RawMessage) bool {
 				if bin == "" || !filepath.IsAbs(bin) {
 					continue
 				}
-				if _, err := os.Stat(bin); err != nil {
+				// Strictly "not there", never "could not tell". EACCES, ELOOP, a
+				// stalled network mount or a parent dir that lost +x all report an
+				// error while the binary may be perfectly fine — and treating those
+				// as deleted would rewrite settings.json every cycle forever and
+				// reopen the cross-instance fight this function exists to avoid.
+				if _, err := os.Stat(bin); errors.Is(err, fs.ErrNotExist) {
 					debuglog.Logger.Warn("claude hooks: hook command no longer exists",
 						"event", cfg.Event, "binary", bin)
 					return true
+				} else if err != nil {
+					debuglog.Logger.Warn("claude hooks: could not stat hook command; assuming it is fine",
+						"event", cfg.Event, "binary", bin, "err", err)
 				}
 			}
 		}
@@ -497,10 +507,15 @@ func RepairClaudeHooks(configDir string) (bool, error) {
 
 	debuglog.Logger.Warn("claude hooks: repairing a hook command that no longer exists",
 		"path", settingsPath, "newCommand", GetHookCommand())
-	if _, err := InjectClaudeHooks(configDir); err != nil {
+	// Report what was actually written, not what we set out to write.
+	// InjectClaudeHooks re-reads settings.json, so a sibling instance that
+	// repaired it between our read and its own leaves nothing to do — and
+	// claiming a repair we did not make latches the caller's tip on for good.
+	wrote, err := InjectClaudeHooks(configDir)
+	if err != nil {
 		return false, err
 	}
-	return true, nil
+	return wrote, nil
 }
 
 // hooksNeedUpdate checks if the hook command path has changed (e.g., after rebuild).
--- a/internal/hooks/claude_hooks.go
+++ b/internal/hooks/claude_hooks.go
@@ -183,7 +183,8 @@ func InjectClaudeHooks(configDir string) (bool, error) {
 		existingHooks = make(map[string]json.RawMessage)
 	}
 
-	if hooksAlreadyInstalled(existingHooks) && !hooksNeedUpdate(existingHooks) && !hasStaleHookEvents(existingHooks) {
+	if hooksAlreadyInstalled(existingHooks) && !hooksNeedUpdate(existingHooks) &&
+		!hasStaleHookEvents(existingHooks) && !hookBinaryMissing(existingHooks) {
 		debuglog.Logger.Debug("claude hooks: already installed and up to date")
 		return false, nil
 	}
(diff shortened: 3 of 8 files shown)


internal/session/session.go
```diff
@@ -1727,10 +1739,36 @@ func (s *Session) updateStatusFromPane(oldStatus Status, log *slog.Logger) {
...
+		// applyHookFinished has its own guards for the same between-bursts frame.
+		// Here there is no hook at all, so the pane is the only vote and a single
+		// bad frame decides everything — it demoted a working session to finished,
+		// which also dropped it off the ~500ms fast pass onto the round-robin,
+		// where the wrong status then stood for ~18s and pulled the Space rotation
+		// onto a session that was busy. So the demotion has to be seen twice.
+		//
+		// Costs one extra pass (~500ms, since a running session is on the fast
+		// pass) before a genuine finish shows.
+		s.paneFinishedFrames++
```
hayke102: This counts **worker cycles, not distinct pane frames**, so a single blink can still demote.

`statusWorkerCycle` runs on the 500ms ticker **or** any `statusTrigger` kick, and there are several: `handleTick`, `enqueuePriorityUpdates` (every hook change from *any* session), the attach-exit path, and reload-all. Two cycles landing within 400ms both read the **same** frame:

- with fewer than 2 active sessions, `seedActiveCaptures` early-returns (`len(active) < 2`) and `CapturePane` serves the `captureCacheTTL` (400ms) cache verbatim — guaranteed duplicate;
- with more, it re-captures, but 50ms later the pane hasn't repainted, so the content is identical anyway.

Either way one blink frame is counted twice and the running session is demoted — the exact bug this is fixing, and most likely on a busy fleet where hook-driven kicks are frequent.

The confirmation needs a **time floor** (`paneFinishedFirstAt`, require ≥ ~400ms since the first reading) or a *content changed since the last finished reading* check, rather than a call count.

internal/session/session.go
```diff
@@ -1727,10 +1739,36 @@ func (s *Session) updateStatusFromPane(oldStatus Status, log *slog.Logger) {
 	case StatusRunning:
 		s.Status = StatusRunning
 		s.Acknowledged = false
+		s.paneFinishedFrames = 0
 	case StatusWaiting:
 		s.Status = StatusWaiting
 		s.Acknowledged = false
+		s.paneFinishedFrames = 0
```
hayke102: These are the only two resets, so the frames aren't actually **consecutive** — which is what the field doc promises.

Two paths leave the counter untouched:

- `default:` (below, session.go:1777) — `detectStatus` returning `""` is the *common* case, and it just logs `no pattern matched`;
- the `return` on capture failure at the top of the function.

So `blink(finished)` → minutes of `no-match` → `blink(finished)` demotes a running session on two non-adjacent frames far apart. Suggest `s.paneFinishedFrames = 0` in `default:` and before the capture-failure `return`.

internal/hooks/claude_hooks.go
```diff
@@ -392,6 +393,116 @@ func hooksAlreadyInstalled(hooks map[string]json.RawMessage) bool {
...
+			}
+			for _, h := range m.Hooks {
+				if !isFleetHook(h.Command) {
+					continue
+				}
+				bin := fleetHookBinary(h.Command)
+				if bin == "" || !filepath.IsAbs(bin) {
+					continue
+				}
+				if _, err := os.Stat(bin); err != nil {
```
hayke102: Any `os.Stat` error is treated as "binary deleted" — EACCES, a stalled network mount, ELOOP, a parent dir that lost `+x`, not just `ENOENT`.

In that state `RepairClaudeHooks` rewrites `settings.json` every 60s forever and latches `hooksRepaired` (so the tip is permanent), and it puts two instances launched from different worktrees straight back into the settings.json fight that `RepairClaudeHooks`'s own doc comment says only a *missing* path may justify.

```go
if _, err := os.Stat(bin); errors.Is(err, fs.ErrNotExist) {
```

internal/hooks/claude_hooks.go
```diff
@@ -392,6 +393,116 @@ func hooksAlreadyInstalled(hooks map[string]json.RawMessage) bool {
...
+	if err := json.Unmarshal(raw, &existingHooks); err != nil {
+		return false, fmt.Errorf("parse hooks section: %w", err)
+	}
+	if !hookBinaryMissing(existingHooks) {
+		return false, nil
+	}
+
+	debuglog.Logger.Warn("claude hooks: repairing a hook command that no longer exists",
+		"path", settingsPath, "newCommand", GetHookCommand())
+	if _, err := InjectClaudeHooks(configDir); err != nil {
```
hayke102: The `bool` from `InjectClaudeHooks` is dropped, so this returns `true` even when nothing was written.

`InjectClaudeHooks` re-reads `settings.json` itself; if a sibling instance repaired it between our read and its own, it returns `(false, nil)` — and we still report a repair, latching `hooksRepaired` and showing the tip for something this process never did.

```go
wrote, err := InjectClaudeHooks(configDir)
if err != nil {
	return false, err
}
return wrote, nil
```

internal/ui/app.go
```diff
@@ -4721,6 +4731,48 @@ func (h *Home) maybeSuspendIdleSessions(sessions []*session.Session) {
...
+// Only ~/.claude is repaired. Per-account config dirs get settings.json copied
+// from it by claudeaccount.Provision on each session launch, so the repair
+// reaches them the same way the hooks did.
+func (h *Home) maybeRepairClaudeHooks() {
+	if !h.lastHookRepairAt.IsZero() && time.Since(h.lastHookRepairAt) < hookRepairInterval {
+		return
+	}
+	h.lastHookRepairAt = time.Now()
+
+	repaired, err := hooks.RepairClaudeHooks(hooks.GetClaudeConfigDir())
```
hayke102: The doc comment above says *"Only `~/.claude` is repaired"*, but `GetClaudeConfigDir()` returns `$CLAUDE_CONFIG_DIR` when it's set — and fleet is commonly launched **from inside a fleet session**, where `sessionEnv` has set that to a per-account dir.

In that case the repair fixes only the account dir; `~/.claude` keeps the dead command, and `claudeaccount.Provision` copies it back into every account dir on the next launch — so the blackout survives the fix in exactly the setup that reported it.

Pre-existing for the launch-time injection, but the new comment now makes a claim the code doesn't hold to. Worth either repairing `~/.claude` explicitly, or narrowing the comment.

internal/ui/tips.go
```diff
@@ -175,6 +193,24 @@ func findTip(id string) *Tip {
...
+// on purpose — as are Starting ones, which simply haven't fired SessionStart yet.
+// Drives the hooks-repaired tip, so it must stay cheap: in-memory reads only.
+func (h *Home) countSessionsMissingHooks() int {
+	n := 0
+	for _, s := range h.sessions {
+		switch s.GetStatus() {
+		case session.StatusSuspended, session.StatusStarting:
+			continue
+		}
+		if s.GetHookStatus() == "" {
```
hayke102, line 223: This doesn't filter by agent, but only `~/.claude/settings.json` is repaired — so the tip counts sessions it can't help and tells the user `r` will fix them.

Codex and OpenCode bake the same binary path: `InjectCodexHooks` builds its entries through the shared `mergeHookEvent`/`fleetHook` → `GetHookCommand()`, and `InjectOpenCodePlugin` embeds `FleetBinaryPath()` in the generated `.ts`. Deleting the binary silences all three; `maybeRepairClaudeHooks` fixes one.

A hookless Codex session is pinned to `StatusIdle` by `UpdateStatus` even while it works, so it lands in this count — and the tip text (tips.go:84) promises a restart restores its status. It won't: `~/.codex/hooks.json` still names the deleted path.

Either repair the Codex/OpenCode targets too, or scope this to `agent.Claude`.

internal/session/session.go
hayke102: Fixed — you're right, and this defeated the fix in its own target scenario. Added `paneFinishedFirstAt` + `paneFinishedConfirmDelay` (400ms, matched to `captureCacheTTL`) alongside the count, so the confirming reading is provably a fresh capture.

Took the time floor rather than the content-changed check: a genuinely finished pane is static, so identical content is exactly what a real finish looks like — content equality can't separate the two.

Pinned by `TestScenarioOneFrameReadTwiceDoesNotConfirm`. The replay engine models the duplicate for free, since it runs checks back-to-back in real time.

internal/session/session.go
hayke102: Fixed — both paths now reset (`resetPaneFinishedConfirmLocked`), including the capture-failure `return`, which needed the lock taking since it sits above it. Pinned by `TestScenarioUnclassifiedFrameBreaksTheConfirmation`, which places the second blink far enough out that the new time floor alone would have passed it.

internal/session/session.go
```diff
@@ -1727,10 +1739,36 @@ func (s *Session) updateStatusFromPane(oldStatus Status, log *slog.Logger) {
...
+		// where the wrong status then stood for ~18s and pulled the Space rotation
+		// onto a session that was busy. So the demotion has to be seen twice.
+		//
+		// Costs one extra pass (~500ms, since a running session is on the fast
+		// pass) before a genuine finish shows.
+		s.paneFinishedFrames++
+		if oldStatus == StatusRunning && s.paneFinishedFrames < paneFinishedConfirmFrames {
+			log.Debug("pane read finished while running — holding for confirmation",
+				"frames", s.paneFinishedFrames, "need", paneFinishedConfirmFrames)
+			return
```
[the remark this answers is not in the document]
hayke102: Fixed — the `default:` branch now resets the counter. Also covered the capture-failure `return` above the switch, which has the same effect and wasn't flagged here.

internal/hooks/claude_hooks.go
hayke102: Fixed — narrowed to `errors.Is(err, fs.ErrNotExist)`; any other stat error now logs and is treated as fine. Pinned by `TestRepairClaudeHooksLeavesAnUnstattableCommandAlone` (EACCES via a parent dir with no +x).

internal/hooks/claude_hooks.go
hayke102: Fixed — returns `wrote` from `InjectClaudeHooks` now. `TestRepairClaudeHooksReportsWhatWasWritten` pins that a second call on an already-healthy file reports no repair.

internal/ui/app.go
hayke102: Fixed — repairing both `GetClaudeConfigDir()` and, when it differs, the literal `~/.claude`.

Confirmed your read: `Provision` reads `filepath.Join(home, ".claude")` literally, so repairing only an account dir gets copied back over on the next session launch. Not narrowing to `~/.claude` alone either — someone may set `CLAUDE_CONFIG_DIR` themselves and that dir is what their sessions read. Both calls write nothing when healthy. Doc comment now states this instead of the claim it couldn't hold.

internal/ui/tips.go
hayke102: Fixed — `countSessionsMissingHooks` is scoped to `agent.Claude` (treating an empty `Agent` as legacy Claude, same fallback `agentGlyph` makes), so the tip can't promise a restart it can't deliver.

Repairing Codex/OpenCode is deliberately out of scope here and filed as #284 — OpenCode needs the embedded path parsed back out of the generated `.ts`, and the same only-when-missing rule has to hold there or it reopens the cross-instance rewrite fight.

internal/ui/tips.go
```diff
@@ -67,6 +68,23 @@ const (
...
+		ID:       tipHooksRepairedID,
+		Policy:   tipRecurring,
+		Priority: 130,
+		active: func(h *Home) bool {
+			return h.hooksRepaired.Load() && h.countSessionsMissingHooks() > 0
+		},
+		text: func(h *Home) string {
+			return fmt.Sprintf("Claude's hooks pointed at a deleted fleet binary, so status came from screen scraping alone. "+
+				"Repaired — but %d running session(s) won't report status until you restart them (`r`).",
+				h.countSessionsMissingHooks())
```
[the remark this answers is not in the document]
hayke102: Fixed — the label is now `session(s)`. Kept the helper counting idle/finished/error too: narrowing it to running would undercount the sessions that actually need restarting. It is now scoped by agent instead (see the thread on tips.go:207).

changelog/unreleased/hooks-repair-deleted-binary.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Status survives a deleted binary.** Hooks pointed at the exact `fleet` binary that installed them, so removing that git worktree — or upgrading past that Homebrew version — silently killed status detection for every session while screen scraping quietly stood in. fleet now re-points them within a minute and names the sessions you need to restart.
```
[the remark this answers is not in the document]
hayke102: Fixed the headline — now **Status stops going dark.** (4 words).

Didn't take the suggested body: it leads with the remedy and never names the annoyance, which the changelog guide explicitly asks for. Rewrote it instead, dropping the worktree/Homebrew mechanism and keeping the user-visible failure.

changelog/unreleased/typed-prompt-false-finished.md
```diff
@@ -0,0 +1,5 @@
+---
+type: fixed
+---
+
+**Typing no longer marks a session done.** Starting a message while the agent was still working could flip it to finished for ~18s — long enough for `Space` to jump you to a session that was busy. A single frame can no longer end a run.
```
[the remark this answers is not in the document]
hayke102: Fixed the headline — now **Busy sessions stay busy.** (4 words), one sentence, `~18s` and the `Space` symptom kept. Rewrote the body rather than taking the suggestion, which dropped the cause.

PR #285, opened 2026-08-27
feat(tickets): narrow the tickets tab to work you have not started

hayke102:
`Ctrl+O` on the tickets tab (`t`) hides started issues, leaving todo, backlog and triage.

```
  all 62   actions 41   repos/worktrees 9   [tickets · todo 7]

  Todo (5)
  ▰▰▱  BRZ-3182  Ticket modal state filter
  ▰▱▱  BRZ-3190  Drawer resize jitter        ● running
  Backlog (2)
  ▱▱▱  BRZ-3201  Revisit snooze presets

  ↑↓ nav  tab: switch tab  ctrl+o: all  enter: select  esc: close
```

## Decisions

**The cut is "not started", not an allowlist of todo+backlog.** Excluding means a category neither provider maps (Linear's `StateOther`) stays visible rather than silently vanishing from a list the user reads as complete. It also isn't spelled "backlog only": Jira emits no backlog or triage category at all, so that scope would read as permanently empty there.

**The scope rides the tickets chip** (`tickets · todo 7`) rather than taking a chip of its own — one thing lit means one thing, and the count beside it is already the scoped count. `renderTabs` applies the same predicate as `rebuildFiltered`, so the bar can never state a number the rows contradict.

**It is the tickets tab's scope, not a global one.** A mixed `all` list quietly holding fewer tickets than its own count claims would be a filter with nothing on screen to explain it.

**`Ctrl+O` because the search box owns the obvious chords.** bubbles' textinput binds `ctrl+a/b/d/e/f/h/k/n/p/u/v/w` plus the plain arrows for line editing; `ctrl+c`/`ctrl+k`/`ctrl+q`/`ctrl+s` are quit, close, detach/XOFF and flow control; `ctrl+i/m/j/[` are byte-level aliases of tab/enter/newline/esc; `ctrl+g`/`ctrl+t` belong to the drawer. Off the tickets tab it falls through to the input rather than arming a scope whose chip is elsewhere.

**Deliberately not reset by `Show`.** A queue you pick your next task out of is the state you want most of the time; re-pressing on every open would make it a chore rather than a mode. It stays legible because the chip renders from every tab, and it is not persisted to config, so a restart clears it.

**`ticketListLimit` 50 → 100 per tracker.** The fetch is ordered by `updatedAt`, so the cut falls on recency rather than state — fifty was truncating exactly the untouched backlog this scope exists to surface. One page for both providers (Linear takes `first` up to 250; Jira's `/search/jql` accepts it in one POST), and the value both already fall back to when handed no limit.

## Tests

- `TestTodoScopeHidesStartedWork` — started row drops, unstarted **and** backlog both survive, and it toggles back.
- `TestTodoScopeCountFollowsTheList` — the chip reads `tickets · todo 2`, so the bar can't lie about the list.
- `TestTodoScopeStaysOnItsOwnTab` — the `all` tab keeps every ticket.
- `TestTodoScopeIsInertOffTheTicketsTab` — `ctrl+o` elsewhere doesn't arm a scope you can't see.

`go test -race ./internal/ui` and `make build` green.

https://claude.ai/code/session_01VDxGYT2mzXfmjnD5aRg93f

## Summary by CodeRabbit

- **New Features**
  - Added a todo-only scope for the tickets tab, toggled with `Ctrl+O`.
  - Filtered results show work that has not yet started, including backlog states.
  - The active scope is reflected in the tickets tab label and result count.
  - Ticket lists now include up to 100 recently updated items per tracker.

- **Documentation**
  - Added release documentation for the new ticket filtering shortcut and expanded ticket retrieval.


internal/ui/command_palette.go
```diff
@@ -580,7 +629,18 @@ func (d *CommandPaletteDialog) View() string {
...
+	// columns spare at an 80-column terminal and a wrapped footer is worse
+	// than a terse one.
+	scope := ""
+	if d.activeTab == PaletteTabTickets {
+		scope = "  ctrl+o: todo"
+		if d.ticketTodoOnly {
+			scope = "  ctrl+o: all"
+		}
+	}
+	b.WriteString(DimStyle.Render("↑↓ nav  tab: switch tab" + scope + "  enter: select  esc: close"))
```
hayke102, line 662: The `~14 columns spare at an 80-column terminal` figure is right, but it's the whole budget — the addition spends all but 2 of it, and below 80 it overdraws.

Measured against `DialogStyle.Width(dialogWidth()).Render(...)` (inner = `min(term-8, 96) - 6`):

| footer | width | fits from |
|---|---|---|
| old | 50 | term 64 |
| new (`ctrl+o: todo`) | 64 | term 78 |

So on terminals **64–77 columns** the footer now wraps to a second line where it didn't before, and the palette box grows a row. At exactly 80 there are 2 columns spare; the comment's reasoning holds at 80 and only at 80.

Not necessarily worth fixing — an 80-column floor is a defensible assumption — but the comment currently reads as though there's headroom left, and there isn't. If you do want it back under the old floor, dropping `nav` (`↑↓` alone is unambiguous) buys 4, and `tab: switch tab` → `tab: tabs` buys 8, which lands it at 52.

internal/ui/command_palette.go
```diff
@@ -103,6 +110,16 @@ type CommandPaletteDialog struct {
 	// from one that genuinely has nothing — the difference between "wait" and
 	// "you're done", which an empty list alone cannot say.
 	ticketsLoaded bool
```
hayke102, line 112: This flag's doc comment promises the exact distinction the todo scope now makes routine — "the difference between 'wait' and 'you're done', which an empty list alone cannot say" — but nothing reads it. `TicketsLoaded()` has no callers outside `palette_tickets_test.go`, and `View` renders a flat `No matches` at command_palette.go:448 for every empty case.

Pre-existing, but the scope changes how often you land there: once `ticket_start_state` has moved your active work to In Progress, "everything I'm assigned is started" is the *normal* state of the tab, not an edge case. So the common outcome of pressing `ctrl+o` is a screen that looks identical to a failed fetch, with only the `tickets · todo 0` chip to tell them apart.

A scoped empty line ("nothing unstarted" vs. "no tickets") would land the same way the chip does — cheap, and it's the state this feature will most often produce.

internal/ui/command_palette.go
```diff
@@ -349,6 +390,14 @@ func (d *CommandPaletteDialog) Update(msg tea.Msg) (*CommandPaletteDialog, tea.C
 	case "shift+tab":
 		d.cycleTab(-1)
 		return d, nil
+	case "ctrl+o":
```
hayke102, line 401: The argument for not resetting on `Show` is that the scope "stays legible because the chip renders from every tab" — but the *control* doesn't. Arm it, press `tab` to `all`, and you're looking at `tickets · todo 2` with no way to clear it from where you're standing: `ctrl+o` falls through to the input, the footer hint is gone, and nothing says the key only works one tab over.

The visibility half of the argument carries; the reachability half doesn't. Two cheap options that keep the deliberate persistence:

- let `ctrl+o` work from any tab while the scope is **on** (turning it off is never ambiguous — there's exactly one scope), or
- keep the footer hint whenever `ticketTodoOnly` is set, not only on the tickets tab.

Second one is one line and doesn't change any key's meaning.

CLAUDE.md
```diff
@@ -191,7 +191,7 @@ Styles in `styles.go` are declared **bare** and constructed **only** in `ApplyPa
...
.... **Deliberately not reset by `Show`**: a queue you pick your next task out of is the state you want most of the time, so re-pressing it on every open would make it a chore rather than a mode — it stays legible because the chip renders from every tab, and it is not persisted to config, so a restart clears it. Colour therefore means session status and only that, the same discipline that keeps agent glyphs monochrome. `Haystack` must be exactly `Name + " " + <right column>`: the renderer maps matched indexes back onto those two strings by offset. The palette's name column is budgeted against the widest *right* column rather than a fixed 22-rune cap, which was truncating titles to `Storage opt…` on a terminal with room to spare. Fetched when the palette opens and never polled — `ticketListLimit` is **100 per tracker**, one page for both (Linear takes `first` up to 250, Jira's `/search/jql` accepts it in one POST), and the number matters because the fetch is ordered by `updatedAt`, so the cut falls on **recency rather than state**: the earlier fifty truncated exactly the untouched backlog the todo-only scope exists to surface. `Ctrl+K` costs nothing at all when no tracker is connected.
```
hayke102: Stale within the same bullet: this still says "typing narrows **50 rows**" a few sentences before the rewritten tail states `ticketListLimit` is **100 per tracker**. Should be 100 (or just "the list").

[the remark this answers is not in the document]
hayke102: Fixed both in 5421e57. CLAUDE.md's "typing narrows 50 rows" now reads "narrows the list" — it was contradicting the tail of its own bullet. The changelog said "just your todo and backlog", which is the allowlist framing the implementation deliberately rejects: the cut is `State != StateStarted`, so triage and unmapped categories survive it too. Now reads "leaving everything you haven't — todo, backlog and triage alike."

hayke102: Fixed in 5421e57 — now reads "typing narrows the list", rather than a number that has to be kept in step with ticketListLimit in two places.

internal/ui/command_palette.go
hayke102: You're right, and the numbers reproduce exactly: inner is min(term-8,96)-6, so the old 50-column footer cleared 64 and the new 64-column one clears 78. Went with the comment rather than the shorter labels — an 80-column floor is assumed throughout this dialog, and trimming `nav`/`tab: switch tab` edits copy this PR didn't otherwise touch. The comment now names both floors (78, and 77 for the shorter `ctrl+o: all` other tabs render since the reachability fix) and states plainly that it wraps below.

internal/ui/command_palette.go
hayke102: Fixed the scoped half in 5421e57: an empty scoped tab now says "Nothing unstarted — ctrl+o shows all". Agreed on why it matters — with ticket_start_state moving active work to In Progress, that is the routine outcome of this scope, not an edge case. Claimed only when rows were really scoped out and nothing is typed; a tracker that returned nothing is not "nothing unstarted", and saying so would be a guess dressed as a reason (TestScopedEmptyTabSaysWhy / TestUnscopedEmptyTabStaysGeneric). Left the pre-existing loading-vs-empty gap alone — TicketsLoaded still has no production caller, and wiring it up is its own change.

internal/ui/command_palette.go
hayke102: Agreed — the reachability half doesn't carry, and it undercut the doc claim I wrote. Fixed in 5421e57 with a variant of your first option: arming still requires the tickets tab (the effect is only visible where the rows are), but while the scope is on, ctrl+o clears it from any tab. Didn't take the second on its own — showing the hint everywhere while leaving the key inert advertises a dead key, which is the dead click the context menu's dimmed rows exist to prevent. So the hint follows the KEY, not the tab: it renders exactly where ctrl+o acts. TestTodoScopeClearsFromAnyTab covers both halves, and TestTodoScopeCannotBeArmedOffItsOwnTab pins the asymmetry.

PR #287, opened 2026-09-02
feat(accounts): pick the Claude account when creating a session with `A`

hayke102:
`Shift+A` grows an **`Account` row beside `Agent`**, so a session starts on the subscription you choose instead of whichever one `account_strategy` ranks first.

The only manual control before this was `.` → **Move to Account…**, which pins *and relaunches* an existing session — you had to create it wrong first, then restart it.

```
│  Agent     ◂ Claude ▸                                        │
│  Account   ◂ Auto — yuval@brizz.ai           12% used ▸      │
│                                                              │
│  ──────────────────────────────────                          │
│  ↑↓ field   ←→ change   ⏎ create   esc cancel                │
```

`↑↓` move between rows, `←→` cycle the focused one — the Settings detail pane's idiom. **`tab` therefore means "next row"** rather than the "next agent" it meant while there was one row.

## Decisions

**`Auto` leads the cycle and is preselected**, riding `sessionCreateMsg.account == ""`, which *already* means "let the strategy pick". So `A ⏎` is byte-for-byte unchanged, and no `accountSet` companion bool is needed (unlike `forkSessionMsg`, where `""` is a live identity meaning the ambient login). Its label names what the strategy would land on, resolved through **`selectAccountFor`, which `resolveAccount` now also calls** — two hand-built `SelectOpts` would drift, and a label naming a different account than the one billed is the lie the account labelling exists to stop telling.

**Nil below two accounts.** With one or none every session runs on the same credential, so the row would be a constant — the rule `previewAccountLabel` already applies to the preview footer. `A` is untouched for the majority.

**Inert, not hidden, on a non-Claude agent.** Hiding would change the box height mid-keystroke, so it dims to `Claude only` — with the account name *removed*, because `yuval@brizz.ai  Claude only` reads as "this session runs on yuval@brizz.ai", the opposite of what happens. The selection survives and reappears on cycling back.

**A blocked account is landed on, dimmed with its reason, and refused.** A cycler can only honour *"a disabled option is shown with its reason"* by letting you reach one, so `⏎` is gated by `submitBlocker` — the one function the footer also reads, the `statusReportForm` shape. The footer swaps to `<reason> — ←→ pick another`, and drops its `↑↓ field` half whenever `Agent` is the only movable row, so no dead key is advertised.

**The allowlist is enforced twice.** An explicit account skips `resolveAccount` and with it the per-origin allowlist, so `handleSessionCreate` re-checks it at the chokepoint — the same check `resolveLaunchAccount` makes for an explicit `--account`. A policy holding on one surface and not the other is worse than not having one.

## Scope

`A` only. `a` stays instant, `n` stays a path prompt, `w` stays base-branch + new-branch — all keep resolving via `account_strategy`.

## Tests

New `internal/ui/session_create_test.go` (the dialog had none) plus four in `account_move_test.go`. Worth noting: **the height guard was vacuous at first** — it measured `View()`, which `lipgloss.Place`s to the full viewport, so a box that grew a row still came back the same height. It now counts the box's own rows, and a deliberate extra row does fail it.

## Verification

`make build`, `make test` (race, 0 failures), `make lint` (0 issues).

Also driven in tmux against a dev build with a throwaway `$HOME`, confirming: the row appears at two accounts and not at one; `Auto` degrades to a bare label when every account is logged out and the strategy can't rank; cycling onto a logged-out account dims it and `⏎` refuses with `logged out — ←→ pick another`; cycling to Codex leaves the box the same height.

https://claude.ai/code/session_017fg2xry9ARvD3wmRKYKWRE

## Summary by CodeRabbit

* **New Features**
  * Added account selection when creating Claude sessions.
  * Added automatic account selection by default, with account availability and quota information.
  * Added keyboard navigation for switching between agents and accounts.
  * Added clear status messages for unavailable, logged-out, or restricted accounts.
  * Account selection is hidden for non-Claude agents and shown only when multiple accounts are available.

* **Documentation**
  * Updated keybinding and changelog documentation for session account selection.


internal/ui/session_create.go
```diff
@@ -78,43 +185,139 @@ func (d *SessionCreateDialog) cycleAgent(delta int) {
...
 	b.WriteString("\n")
-	b.WriteString(dimStyle.Render("←→ agent   ⏎ create   esc cancel"))
+	// The footer names what ⏎ does now, so a refused account replaces it rather
+	// than sitting beside it claiming ⏎ creates.
+	footer := sessionCreateFooterOnly
+	if d.accountRowActive() {
+		footer = sessionCreateFooter
+	}
+	if blocker := d.submitBlocker(); blocker != "" {
+		footer = blocker + " — ←→ pick another"
```
hayke102: `submitBlocker()` is focus-independent, so this footer also renders while the highlight sits on the **Agent** row — where `←→` cycles the agent, not the account.

Land the Account row on a dimmed option (`logged out` / `not allowed here`), press `k`, and the dialog reads `not allowed here — ←→ pick another` while the only thing `←→` can pick is a different agent. One press to Codex makes `accountRowActive()` false, which clears the blocker, so `⏎` creates — off a footer that had just named the account as the reason it wouldn't. The swap also drops the `↑↓ field` half, so nothing on screen says how to get back to the row the message is about.

Cheapest fix is to gate the swap on `d.focus == focusCreateAccount`. The alternative is to pull focus back to the Account row whenever a blocker is live, so the footer and the highlight can't disagree in the first place.

internal/ui/session_create.go
```diff
@@ -78,43 +185,139 @@ func (d *SessionCreateDialog) cycleAgent(delta int) {
...
-		Padding(1, 2).
-		Width(dialogWidth)
+	state := accountStateCell(r)
 
-	box := boxStyle.Render(b.String())
-	return lipgloss.Place(d.width, d.height, lipgloss.Center, lipgloss.Center, box)
+	// The value cell is what's left after the label, the two arrows and their
+	// spaces. Budget off raw text and pad before styling — padding a styled
+	// string counts the ANSI bytes and the columns come out ragged.
+	width := max(inner-sessionCreateLabelW-len("   ")-len("◂ ")-len(" ▸"), 1)
```
hayke102: `len("◂ ")` and `len(" ▸")` are 4 **bytes** each (`◂`/`▸` are 3-byte runes) but 2 **columns**, so the value cell is budgeted 4 columns narrower than it actually is. The comment directly above says the budget is what's left after the label, the two arrows and their spaces — it isn't.

Constant and cosmetic, but it costs the account label ~4 columns of truncation headroom it has, and this row's `▸` stops short of the width the rest of the box uses. It's also the one measurement in this function that doesn't go through `lipgloss.Width`, which `nameW` and `pad` two lines down both do.

internal/ui/session_create.go
```diff
@@ -49,20 +125,51 @@ func (d *SessionCreateDialog) Update(msg tea.Msg) (*SessionCreateDialog, tea.Cmd
 	switch keyMsg.String() {
 	case "esc", "q":
 		d.Hide()
+	case "up", "k", "shift+tab":
+		d.moveFocus(-1)
+	case "down", "j", "tab":
```
hayke102: `tab` was `case "right", "l", "tab": d.cycleAgent(1)` on master. Routing it through `moveFocus` makes it a dead key for anyone with fewer than two accounts: `sessionCreateAccountRows` returns nil below `Len() < 2`, so `accountRowActive()` is false and `moveFocus` returns immediately.

That's the majority of users, and `sessionCreateFooterOnly` advertises neither `tab` nor `↑↓`, so there's no signal the key changed meaning — it just stops doing what it did. (`j`/`k` going inert there is fine; they were never bound.)

Falling through to `cycleAgent(delta)` when the Account row is absent would keep `tab` advancing the only cycler on screen.

internal/ui/session_create.go
hayke102: Fixed in 685fa35 — good catch, and the diagnosis was exact.

One note on the suggested cheap fix: gating the swap on `d.focus == focusCreateAccount` would make the footer read `⏎ create` while `⏎` refuses, since `submitBlocker` is focus-independent for a real reason — the blocked account is still the one that would be used wherever the highlight sits. That's the footer↔enter disagreement `submitBlocker` exists to prevent, so the swap stays focus-independent and the *route* follows the focus instead:

- focus on Account → `logged out — ←→ pick another`
- focus on Agent → `logged out — ↓ then ←→ pick another`

Extracted into `footer()` with the reasoning written down.

On the second half — one press to Codex clearing the blocker so `⏎` creates — I left that as is. A Codex session genuinely ignores the account, so the outcome is correct; it's jarring, not wrong.

`TestCreateDialogFooterAndEnterAgree` only ever ran with focus on the Account row, which is exactly why this shipped. It now covers both focus states, plus `TestCreateDialogBlockedFooterNamesAReachableKey` for the route. Reverting the fix fails both.

internal/ui/session_create.go
hayke102: Fixed in 685fa35 — confirmed, `len("◂ ")` is 4 and `lipgloss.Width("◂ ")` is 2, so the cell reserved 8 columns of chrome where 4 exist.

The arrows and the gap are now named constants shared by both the budget and the line `renderRow` builds, so the two can't drift, and the total goes through `lipgloss.Width`:

```go
var sessionCreateRowChrome = sessionCreateLabelW +
	lipgloss.Width(sessionCreateGap + sessionCreateLead + sessionCreateTrail)
```

`TestCreateDialogRowChromeIsMeasuredInColumns` pins the budget against what `renderRow` actually spends; with `len()` back it reports "a row is 24 columns wide but the budget reserves 28 — off by 4", the exact number you named. Visible in the app too: the Account row's `▸` now lands in the column the rest of the box uses.

internal/ui/session_create.go
hayke102: Fixed in 685fa35 — a real regression, and my own tests missed it: `TestCreateDialogHasNoAccountRowBelowTwoAccounts` asserted `down` doesn't move focus but never that `tab` still did its old job.

`tab` now advances the next thing you can change — the next row when there are two, the agent value when the Account row is absent — so master's behaviour is preserved for everyone below two accounts. Took your carve-out exactly: `↑↓`/`j`/`k` stay pure row navigation and stay inert there, since they were never bound and nothing regresses.

`TestCreateDialogTabStillCyclesTheAgentBelowTwoAccounts` and `TestCreateDialogTabMovesRowsWithTwoAccounts` pin both halves; reverting to `moveFocus` fails the first. Verified in a dev build too — `tab` on a one-account fleet takes Claude → Codex again.

internal/ui/session_create.go
```diff
@@ -78,43 +185,139 @@ func (d *SessionCreateDialog) cycleAgent(delta int) {
...
-	dimStyle := lipgloss.NewStyle().Foreground(ColorTextDim)
-	arrowStyle := lipgloss.NewStyle().Foreground(ColorAccent)
-	valueStyle := lipgloss.NewStyle().Foreground(ColorText).Bold(true)
+	boxW := sessionCreateWidth
+	if maxW := d.width - 4; boxW > maxW {
+		boxW = maxW
+	}
+	if boxW < 30 {
+		boxW = 30
+	}
```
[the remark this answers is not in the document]
hayke102: Correct that the box exceeds the viewport below 34 columns, but not something I want to change in this PR.

This clamp is pre-existing — `git show master:internal/ui/session_create.go` has it byte for byte (`dialogWidth := 54` → clamp to `d.width-4` → floor at 30). I renamed the variable and changed the constant; the logic is untouched.

It's also the package convention rather than a local slip. The same floor-after-clamp is in `branch_checkout.go:323`, `dialogs.go:327`, `dialogs.go:447`, `workspace_create.go:348`, `workspace_picker.go:762` and `command_palette.go:966`. Fixing it here alone would make this the one dialog that behaves differently at 33 columns, which is worse than the current uniform behaviour — and fleet has no minimum terminal-size guard at all, so the right fix is one change across all six plus a guard, not a one-off here.

Tracking it as a follow-up. The rest of this review is addressed in 685fa35.

PR #293, opened 2026-09-08
feat(worktree): fetch the base ref before branching from it

hayke102:
## Summary

`git worktree add -b <new> origin/master` resolves `origin/master` out of the local clone and **never asks the remote**, so a worktree could start days behind and need a merge as its first act. Both create paths now refresh that one ref first — the `w` dialog's async closure and `fleet worktree`.

This started as "should fleet do a periodic background `git fetch`?" The answer, after measuring, is no. Across 10 origin groups on one machine the repos worktrees are actually cut from were already current, because agent sessions fetch as part of the work (brizz-code's `git reflog origin/master` shows `fetch -q origin master` on three separate days). A periodic sweep would spend ~2s per group forever refreshing what's already fresh, and would hand users auto-gc repacks competing with their running sessions plus a credential prompt that hangs on `/dev/tty`. The staleness only bites at one instant, so that's where the fetch goes.

### Design notes

- Refspec-less `origin <branch>` is deliberate — it's what opportunistically updates `refs/remotes/origin/<branch>`, the ref `worktree add` then resolves.
- `--no-write-fetch-head` because FETCH_HEAD belongs to the user and a manual `git merge FETCH_HEAD` must not merge fleet's fetch. It costs the feature on git <2.29 — the safer of the two failures.
- `-c gc.auto=0 -c maintenance.auto=false` because a fetch normally signs off by kicking automatic housekeeping, which on a large repo is a repack competing with the sessions already running.
- Only an `origin/`-prefixed base is fetched, the only form fleet produces: splitting a bare name on `/` would read a local `feature/foo` as remote `feature` and burn the whole budget on a fetch that cannot work.
- **Advisory**, exactly like the file copies beside it: `GIT_TERMINAL_PROMPT=0` so an expired credential fails instead of sitting on a prompt, a 5s budget deliberately under the 8s `gitTimeout` since offline it can only ever fail, and any failure branches from the refs already there.
- Placed **after** the dialog closes, inside the `Creating…` phantom's window, where a ~2s round trip is invisible. At dialog-open time it would either hold the box shut for it or move the base-branch suggestion rows under the cursor when it landed.
- `F` (fork) is untouched — it branches from the session's HEAD, not from origin. The CLI skips the fetch when the branch already exists, since `Create`'s no-`-b` retry drops the base anyway.

## Type of Change

- [x] New feature

## Checklist

- [x] Documentation updated (`CLAUDE.md` conventions)
- [x] Changelog fragment added in `changelog/unreleased/`

## Test plan

Two tests, asserting both directions:

- `TestFetchBaseRefAdvancesTrackingRef` — a bare remote, a teammate clone that pushes, and the user's clone: `origin/master` advances to the pushed tip, and `HEAD` does **not** move (the fetch must be invisible to the checkout it runs in).
- `TestFetchBaseRefSkipsLocalBase` — `master`, `feature/foo` and `""` trigger no fetch, behind a broken origin URL as a positive control. Without that control a nil error would prove nothing.

https://claude.ai/code/session_01FRXgjKJNEYNyWYQ8z2kLLf

## Summary by CodeRabbit

- **New Features**
  - Worktree creation now refreshes the selected remote base branch before creating a new worktree.
  - Uses the latest available remote branch tip when creating branches.
  - Refreshing occurs in the background with a five-second timeout and does not interrupt worktree creation if unavailable.
  - Existing branches, local bases, and custom providers remain unchanged.

- **Documentation**
  - Added documentation describing the updated worktree creation behavior.


internal/git/git.go
```diff
@@ -215,6 +216,59 @@ func GetDefaultBranch(repoPath string) string {
...
+	// sessions already running. fleet is asking for one ref, not volunteering the
+	// user's CPU. GIT_TERMINAL_PROMPT=0: an expired credential would otherwise
+	// sit on a prompt for the whole timeout instead of failing immediately.
+	cmd := exec.CommandContext(ctx, "git",
+		"-C", repoPath,
+		"-c", "gc.auto=0", "-c", "maintenance.auto=false",
+		"fetch", "--quiet", "--no-tags", "--no-write-fetch-head",
+		"origin", branch)
+	cmd.Env = append(os.Environ(), "GIT_TERMINAL_PROMPT=0")
+	if out, err := cmd.CombinedOutput(); err != nil {
```
hayke102, line 309: `cmd.WaitDelay` isn't set, so `fetchTimeout` doesn't actually bound this. `CommandContext` kills `git` at 5s, but `CombinedOutput`'s `Wait` then blocks until every process holding the pipe write-end exits — and `git fetch` over SSH hands those to an `ssh` grandchild. A passphrase-protected key with no agent loaded (or an HTTPS credential helper popping a GUI) leaves `ssh` sitting on the prompt, the pipe open, and `FetchBaseRef` hanging well past the timeout.

That turns the advisory step into a blocking one, which is the one thing the doc comment promises it isn't: in the TUI the `Creating…` phantom spins forever and the worktree is never created; from `fleet wt` the command hangs with no output. `GIT_TERMINAL_PROMPT=0` doesn't cover it — it governs git's own prompts, not `ssh`, which reads `/dev/tty` directly (and will fight the TUI for the terminal while it's at it).

```go
cmd.WaitDelay = time.Second
```

and possibly `GIT_SSH_COMMAND=ssh -o BatchMode=yes` alongside `GIT_TERMINAL_PROMPT=0`, so ssh fails fast instead of prompting at all.

internal/ui/app.go
```diff
@@ -1665,6 +1665,22 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+			// resolves origin/<base> out of this clone and never asks the
+			// remote, so without this a worktree silently starts behind and the
+			// first thing it needs is a merge. Advisory, exactly like the file
+			// copies below — a failed fetch costs freshness, never the worktree.
+			//
+			// Here rather than when `w` opens: the dialog is already closed and
+			// the Creating… phantom is already spinning, so the round trip is
+			// invisible. Fetching at open time would either hold the box shut
+			// for it or move the suggestion rows under the cursor when it landed.
+			if !provider.IsCustom() && baseBranch != "" {
```
hayke102: The CLI path guards the same call with `!reusedBranch` (`cmd/fleet/worktree.go:388`), because `Create`'s no-`-b` retry drops `baseBranch` entirely (`internal/workspace/provider.go:93`) — so when the branch already exists the fetch refreshes a ref `worktree add` never consults. The `w` dialog has no such guard and pays the full round trip (5s, or unbounded per the other comment) for nothing.

Low impact on its own, but the CLI-side comment states it as the rule for "the same act", so the asymmetry reads unintentional rather than decided.

internal/git/git.go
```diff
@@ -215,6 +216,59 @@ func GetDefaultBranch(repoPath string) string {
...
+// "feature/foo" as remote "feature" — spending the whole timeout on a fetch
+// that cannot work.
+//
+// The refspec-less `origin <branch>` form is deliberate: it is what
+// opportunistically updates refs/remotes/origin/<branch>, which is the ref
+// `worktree add` then resolves. The error is returned for logging only — every
+// caller carries on with the refs it already has, so a failure costs freshness,
+// never the worktree.
+func FetchBaseRef(repoPath, baseRef string) error {
+	branch, ok := strings.CutPrefix(baseRef, "origin/")
```
hayke102, line 278: Non-blocking, but worth knowing: this gate means the fetch never fires where `GetDefaultBranch` falls back to a bare local `main`/`master` — repos created with `git init` + `git remote add origin` rather than `git clone`, which have no `refs/remotes/origin/HEAD`. Those are exactly the clones most likely to be stale, since nothing else in fleet fetches them.

Pre-existing `GetDefaultBranch` behaviour and arguably out of scope here, but it decides whether the feature runs at all. Cheap close if you want it: return `origin/<name>` from the fallback when `refs/remotes/origin/<name>` exists.

internal/git/git.go
```diff
@@ -215,6 +216,59 @@ func GetDefaultBranch(repoPath string) string {
...
+	// gc.auto / maintenance.auto: a fetch normally signs off by kicking automatic
+	// housekeeping, which on a large repo is a repack competing with the agent
+	// sessions already running. fleet is asking for one ref, not volunteering the
+	// user's CPU. GIT_TERMINAL_PROMPT=0: an expired credential would otherwise
+	// sit on a prompt for the whole timeout instead of failing immediately.
+	cmd := exec.CommandContext(ctx, "git",
+		"-C", repoPath,
+		"-c", "gc.auto=0", "-c", "maintenance.auto=false",
+		"fetch", "--quiet", "--no-tags", "--no-write-fetch-head",
+		"origin", branch)
```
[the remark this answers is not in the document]
hayke102: Fixed in f797ed1 — `--` now terminates option parsing before `origin`.

Reproduced before fixing: `git fetch origin --upload-pack=<script>` does execute the script, and the terminator turns it into `fatal: couldn't find remote ref --upload-pack=…`. Added `TestFetchBaseRefRefusesAnOptionLookingBase`, verified to fail against the unfixed code.

One correction to the severity for the record: git refuses to create a branch named `--upload-pack=x` (`fatal: not a valid branch name`), so this cannot arrive through the dialog's branch suggestions — it needs the user to type it into the Base branch field or `--base` themselves. Fixed anyway: one token, and it closes the class for callers that pass a less trusted base later.

internal/git/git.go
hayke102: Fixed in f797ed1 — you were right, and it was worse than the doc comment implied.

Measured against a child that leaks a grandchild holding the pipe, on a 1s budget: **30.17s without `WaitDelay`, 1.1s with it**. So the "5s budget, offline it gives up" was the one promise that comment made and the one it did not keep. `cmd.WaitDelay = 1s` now, with the reasoning recorded beside `fetchTimeout`.

Took the ssh half too, in the conservative form: `GIT_SSH_COMMAND=ssh -o BatchMode=yes` is set **only when the caller defines none** — overriding one that carries an identity (`-i ~/.ssh/work_key`) would turn a working fetch into a silently failing one. `core.sshCommand` in git config is the residual gap; reading it costs a subprocess and `WaitDelay` bounds it, so it is documented rather than handled.

`TestFetchBaseRefReturnsWhenAChildLeaksThePipe` pins it. It needs a full second of budget on purpose: at 200ms the fake git is killed before it spawns the grandchild, and the test then passes against the unfixed code — which it did, until I checked.

internal/ui/app.go
hayke102: Fixed in f797ed1 — unintentional, exactly as you read it.

`branchExists` is promoted out of `cmd/fleet/worktree.go` into `internal/git` as `git.BranchExists`, so there is one implementation and both create paths guard on it. The `w` dialog no longer pays a round trip for a ref `worktree add` never reads.

internal/git/git.go
hayke102: Fixed in f797ed1 — took the cheap close.

Confirmed the premise first: `git init` + `git remote add` + `git fetch` leaves `refs/remotes/origin/HEAD` unset while `refs/remotes/origin/master` exists. The fallback now prefers `origin/<name>` when that ref verifies and keeps the bare name only where no remote copy exists — which also settles the contradiction with `GetDefaultBranch`'s own doc comment ("returns the remote-tracking ref … so new worktrees start from the remote tip").

Worth noting for the test: git 2.46+ sets `origin/HEAD` opportunistically on fetch, so `TestGetDefaultBranchPrefersOriginInFallback` deletes it rather than skipping — the fallback is still what every older git reaches here (Ubuntu 24.04 ships 2.43), and what any repo with an unresolvable remote HEAD reaches on every version.

Issue #262, opened 2026-08-23
new session opened on wrong account - maybe it's calculated by weekly limit and not 5h?


1jonahg1:
## Bug Report

<img width="503" height="48" alt="Image" src="https://github.com/user-attachments/assets/e80b6a0d-e790-48c2-84f3-7f0764e52c8a" />

### Description
new session opened on wrong account - maybe it's calculated by weekly limit and not 5h?

### Steps to Reproduce (last 20 actions)
| Time | Action | Detail | Result |
|------|--------|--------|--------|
| 10:47:37 | open bug report |  | ok |
| 10:47:04 | delete session | brizzai-BRZ-3218-quota-split-mcp-from-agent | ok |
| 10:47:00 | create session | ~/brizzai-BRZ-3218-quota-split-mcp-from-agent | ok |
| 10:46:43 | attach session | brizzai-BRZ-3218-quota-split-mcp-from-agent | ok |
| 10:46:30 | create session | ~/brizzai-BRZ-3218-quota-split-mcp-from-agent | ok |
| 10:44:40 | attach session | CR1 | ok |
| 10:44:39 | attach session | CR3 | ok |
| 10:44:37 | attach session | CR2 | ok |
| 10:42:35 | attach session | Pull request explanation | ok |
| 10:41:51 | attach session | Pull request explanation | ok |
| 10:38:44 | attach session | https://github.com/brizzai/brizzai/pull/4851 | ok |
| 10:38:35 | attach session | Explain PR | ok |
| 10:38:27 | attach session | https://github.com/brizzai/brizzai/pull/4851 | ok |
| 10:38:24 | attach session | Explain PR | ok |
| 10:34:12 | attach session | Pull request explanation | ok |
| 10:34:07 | attach session | Explain PR | ok |
| 10:33:08 | attach session | https://github.com/brizzai/brizzai/pull/4849 | ok |
| 10:28:13 | attach session | https://github.com/brizzai/brizzai/pull/4847 | ok |
| 10:28:09 | mark unread | https://github.com/brizzai/brizzai/pull/4849 | ok |
| 10:27:56 | attach session | https://github.com/brizzai/brizzai/pull/4849 | ok |

### Diagnostics
- **Version**: 2.29.0
- **OS**: macOS 26.5.2 (arm64)
- **tmux**: tmux 3.7b
- **Claude CLI**: 2.1.236 (Claude Code)
- **Codex CLI**: codex-cli 0.147.0
[...]

Issue #219, opened 2026-07-27
Wrong status: showed finished, expected finished — New PR takes some time to identify in fleet


eylonronen:
## Wrong Status Detected

**fleet showed `finished` — should have been `finished`**

### What happened
New PR takes some time to identify in fleet

### Status Signals
| Signal | Value |
|--------|-------|
| agent | `claude` |
| captured at | 2026-07-27T13:46:02.809018+03:00 |
| hook status | `finished` |
| hook event | `Stop` |
| hook age | 14.809s |
| pane detected | `finished` |
| TUI showed | `finished` |
| **mismatch** | **false** |
| worker stalled | false |
| worker last cycle | 544ms ago |
| worker cycle in flight | 77ms |
| transcript advanced past hook | false |
| transcript seconds past hook | 0.5 |
| transcript last entry | 14.338s ago |
| transcript recent gaps (s) | `[-2.6 0 2.7 0.2 1.5 0.2 0 0]` |
| Claude own status | `idle` |
| Claude pid alive | true |
| Claude version | 2.1.220 |

<details><summary>Screen (last 40 lines)</summary>

```

⏺ No Terraform/Terragrunt binaries are available in this environment, so I can't run a live plan. Let me review the full diff carefully instead.

  Ran 1 shell command 

⏺ This all looks correct. Let me summarize the changes and what to check before applying.

⏺ User answered Claude's questions:
  ⎿  · No terraform/terragrunt binary is available in this sandbox, so I couldn't run `terragrunt plan` to validate. Want me to do anything else before you review/apply, or is a manual 
     review of the diff sufficient? → Manual review is fine (Recommended)

⏺ Summary

  On new branch add-manual-full-refresh-job (from up-to-date origin/master):

  - modules/scheduled_job/variables.tf / main.tf: cron is now optional. When omitted, the job's schedule trigger is false and schedule_type/schedule_cron are left unset — the job exists
  but only runs via manual trigger (UI/API), never on a schedule. All other triggers (github_webhook, git_provider_webhook, on_merge) stay false as before.
  - dbtcloud.tf: cron = try(each.value.cron, null) so job YAMLs can omit the key.
  - jobs/full-refresh-downstream.yaml (new): projects: [] (all envs, per the existing "empty list = every env" rule), num_threads: 10, and execute_steps: dbt build --select 
  config.full_refresh:true+ — the config.full_refresh:true node-selector method picks models/seeds with that config set, and the trailing + graph operator pulls in everything downstream.
[...]

PR #294, opened 2026-09-08
fix(report): label issues filed from `!`, for reporters without write access

hayke102:
## Problem

The report dialog (`!`) has always passed `--label bug` / `--label enhancement` to `gh issue create`. It has almost never worked.

GitHub accepts labels on issue creation **only from users with write access**, and silently drops them for everyone else — no error, `gh` still exits 0, so nothing on our side could tell. fleet is open source and most reporters are outside contributors, so in practice every report they filed landed unlabelled:

| author | repo permission | labels applied |
|---|---|---|
| `hayke102`, `yairsimantov20` | write / admin | yes |
| `1jonahg1`, `tonytimo`, `eylonronen`, … | `pull` or none | **none** |

42 issues were unlabelled. Every one was filed by someone without triage permission.

## Why not client-side

`gh issue edit --add-label` after creation is refused for exactly the same reason. The label has to be applied by something that holds write access — i.e. the repo itself.

## Fix

`createGitHubIssue` appends the requested label to the issue body as an HTML comment:

```
<!-- fleet-report-label: bug -->
```

`.github/workflows/label-reports.yml` reads it back on `issues: opened` and applies it with `GITHUB_TOKEN`.

- The marker goes in at the **one place all three report kinds pass through**, so it carries whatever label that kind asked for — bug, wrong-status and feature all covered by one line.
- `--label` **stays**. A maintainer's issue is still labelled at creation rather than a workflow run later.
- The workflow **allowlists `bug|enhancement`**. An issue body is written by whoever opened the issue, so the marker is a request, not an instruction — without the allowlist anyone could hand themselves `autorelease`, the label a release PR keys on, by pasting a comment into an issue.

## Test

`TestLabelMarkerMatchesTheWorkflow` pins the one coupling this has: the marker is built in Go, extracted by sed in YAML, and nothing else fails when one of them moves. It asserts the workflow still contains the BRE it extracts with, translates it to a Go regexp, and checks every label fleet emits both matches it and survives the allowlist. A marker the workflow can't read is invisible — the issue is still filed, and still lands unlabelled, which is the state this exists to fix.

## Also done (outside this diff)

- **39 existing issues backfilled.** Four were filed on the wrong form (body says `## Bug Report`, content is a feature request) and got `enhancement`: #159, #53, #43, #22. Three test issues left unlabelled: #76, #6, #5.
- **6 issues closed with explanations** — #274, #157, #160, #28 shipped long ago; #219 (expected == shown; the real complaint is the 60s `gh` poll) and #262 (weekly ranking is the deliberate default since #251) answered rather than fixed.

https://claude.ai/code/session_01FZ2cqYhq9DiCoDoTdX9Rf5

## Summary by CodeRabbit

- **New Features**
  - Bug reports submitted through the app now include classification information that automatically applies either a **bug** or **enhancement** label to the resulting issue.
  - Only supported classifications are applied; unrecognized classifications are safely ignored.
  - When multiple classification markers are present, the latest one determines the issue label.

### DIFF
--- a/.github/workflows/label-reports.yml
+++ b/.github/workflows/label-reports.yml
@@ -22,7 +22,14 @@ jobs:
           # honoured. Without the allowlist anyone could hand themselves
           # `autorelease` — the label merging a release PR keys on — by pasting
           # a comment into an issue.
-          LABEL=$(printf '%s' "$BODY" | sed -n 's/.*<!-- fleet-report-label: \([a-z-]*\) -->.*/\1/p' | head -1)
+          #
+          # tail, not head: fleet appends its marker after the body is fully
+          # built, so it is always the last one. Everything before it is written
+          # by the reporter — a feature request's body is their description
+          # verbatim, and a bug or wrong-status body carries the debug log,
+          # error history and pane excerpt. Any of those quoting the marker
+          # would otherwise outrank the kind they picked in the dialog.
+          LABEL=$(printf '%s' "$BODY" | sed -n 's/.*<!-- fleet-report-label: \([a-z-]*\) -->.*/\1/p' | tail -1)
           case "$LABEL" in
             bug|enhancement) ;;
             *) echo "::notice::ignoring unknown report label '$LABEL'"; exit 0 ;;
--- a/.github/workflows/label-reports.yml
+++ b/.github/workflows/label-reports.yml
@@ -1,0 +1,32 @@
+name: Label Reports
+
+on:
+  issues:
+    types: [opened]
+
+permissions:
+  issues: write
+
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    if: contains(github.event.issue.body, '<!-- fleet-report-label:')
+    steps:
+      - name: Apply the label the report asked for
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BODY: ${{ github.event.issue.body }}
+        run: |
+          # An issue body is written by whoever opened the issue, so the marker
+          # is a request, not an instruction: only these two labels are ever
+          # honoured. Without the allowlist anyone could hand themselves
+          # `autorelease` — the label merging a release PR keys on — by pasting
+          # a comment into an issue.
+          LABEL=$(printf '%s' "$BODY" | sed -n 's/.*<!-- fleet-report-label: \([a-z-]*\) -->.*/\1/p' | head -1)
+          case "$LABEL" in
+            bug|enhancement) ;;
+            *) echo "::notice::ignoring unknown report label '$LABEL'"; exit 0 ;;
+          esac
+          gh issue edit "${{ github.event.issue.number }}" \
+            --repo "${{ github.repository }}" \
+            --add-label "$LABEL"
--- a/internal/ui/bugreport_test.go
+++ b/internal/ui/bugreport_test.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"slices"
+	"sort"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


.github/workflows/label-reports.yml
```diff
@@ -0,0 +1,32 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BODY: ${{ github.event.issue.body }}
+        run: |
+          # An issue body is written by whoever opened the issue, so the marker
+          # is a request, not an instruction: only these two labels are ever
+          # honoured. Without the allowlist anyone could hand themselves
+          # `autorelease` — the label merging a release PR keys on — by pasting
+          # a comment into an issue.
+          LABEL=$(printf '%s' "$BODY" | sed -n 's/.*<!-- fleet-report-label: \([a-z-]*\) -->.*/\1/p' | head -1)
```
hayke102: `head -1` takes the *first* line carrying a marker, but fleet appends its authoritative marker at the *end* of the body (`bugreport.go:366` — `body += fmt.Sprintf(labelMarkerFmt, label)`). Any earlier marker in user-supplied content wins.

Concrete: a feature request whose description quotes an earlier fleet issue containing `<!-- fleet-report-label: bug -->`, or a wrong-status report whose pane excerpt / debug-log block carries that string, lands labelled `bug` — silently contradicting the kind the reporter picked in the dialog.

`tail -1` matches where fleet actually writes it. No privilege consequence either way (both outcomes are allowlisted), just a wrong label on exactly the reports this flow exists to triage.

internal/ui/bugreport_test.go
```diff
@@ -232,3 +234,43 @@ func TestAccountEmailsNeverReachAnIssue(t *testing.T) {
...
+	// string nothing uses.
+	bre := `<!-- fleet-report-label: \([a-z-]*\) -->`
+	if !strings.Contains(string(wf), bre) {
+		t.Fatalf("workflow no longer extracts with %q — update the marker or this test", bre)
+	}
+	re := regexp.MustCompile(strings.NewReplacer(`\(`, `(`, `\)`, `)`).Replace(bre))
+
+	// Both labels the workflow allowlists. A label fleet emits that the
+	// workflow refuses is dropped just as silently as GitHub dropped it.
+	for _, label := range []string{"bug", "enhancement"} {
```
hayke102: The comment above says a label fleet emits that the workflow refuses is caught here, but the loop iterates a literal set rather than the labels the code actually emits (the string literals in `openGitHubIssue` and `submitStatusReport`).

Add a fourth report kind with `label = "documentation"`: the workflow's `case bug|enhancement` refuses it, the issue lands unlabelled, and this test still passes — the exact regression it exists to pin. A shared `var reportLabels = []string{"bug", "enhancement"}` that both the call sites and this loop read would close it.

Minor, same test: the allowlist check on line 272 is a `strings.Contains` over the whole file, so it matches the explanatory comment block above the `case` as happily as the `case` arm itself — it isn't really asserting the allowlist.

.github/workflows/label-reports.yml
hayke102: Fixed in a4adc33 — `head -1` → `tail -1`.

You're right that it's where fleet writes it, not a mirror-image guess: `createGitHubIssue` appends the marker after the body is fully built, so nothing can follow it. Verified against the case you describe — a feature request body quoting `<!-- fleet-report-label: bug -->` now resolves to `enhancement`.

The test pins `tail` now, so flipping it back fails rather than silently mislabelling.

internal/ui/bugreport_test.go
hayke102: Fixed in a4adc33 — both parts.

Labels are now `labelBug`/`labelEnhancement` behind a shared `reportLabels`, which the three call sites and the test both read, so your `documentation` case is visible to it.

For the allowlist check I went further than a tighter `Contains`: the test extracts the `case` arm and asserts **set equality** with `reportLabels`. That fails in both directions — a label fleet emits that the workflow refuses (your case), and a label the workflow honours that fleet never emits, which would be a silent widening of what an issue body can ask for.

Mutation-checked all three: adding `documentation` to `reportLabels`, adding `autorelease` to the workflow arm, and flipping `tail` back to `head` each fail.

PR #295, opened 2026-09-08
feat(ui): jump origin to origin with Ctrl+Shift+arrow

hayke102:
`Shift+↑/↓` steps header by header, so crossing a repo with a dozen worktrees costs a dozen presses. `Ctrl+Shift+↑/↓` is the same motion one level up — **origin headers only**, checkout headers scanned past.

From a session row, `Ctrl+Shift+↑` surfaces your own origin header first and then climbs, exactly as `Shift+↑` surfaces your own checkout header before climbing out. Clamps to the top/bottom row when no origin remains, so it doubles as a jump to the ends of the list.

## Why one scan, not two

`NextHeaderItem` and the new `NextOriginItem` both delegate to `nextRowMatching`, differing only in the predicate. The three things that are easy to get subtly wrong — the "own group first" landing, the edge clamp, and the out-of-range cursor guard (`rebuildFlatItems` doesn't fix `h.cursor`, only `syncViewport` does) — therefore can't drift apart between the two motions.

## Tests

- `TestOriginJumpSkipsCheckoutHeaders` — the skip-checkouts semantics plus both edge clamps.
- `TestOriginJumpKeyBinding` — pins the binding *strings*. `handleKey` matches on `msg.String()`, so a different modifier spelling would dead-press the key with nothing failing; this mirrors the existing `TestHeaderJumpKeyBinding`.

## Verified on screen

Driven in the `demo/drive.sh` sandbox, not just compiled. From a session under `api-server/main`: `C-S-Down` → `notes` origin (skipping the `auth` checkout header) → `tools`; `C-S-Up` walks back; from inside `api-server` it returns to that origin's own header. `Shift+↓` still stops on `auth`, unchanged. Help overlay checked at 80 columns — the key column grows 12→16 and still fits.

## Caveat

The chord needs a terminal that sends `CSI 1;6A`. iTerm2, Ghostty, kitty, WezTerm and VS Code do; **Apple Terminal does not**, where it will dead-press and `Shift+↑/↓` stays the fallback. Noted in `CLAUDE.md`.

https://claude.ai/code/session_01P7iDbYEjK7HhEvCN51kDPr

## Summary by CodeRabbit

* **New Features**
  * Added `Ctrl+Shift+↑/↓` navigation to jump directly between origin headers, skipping checkout headers.
  * Existing `Shift+↑/↓` navigation continues to move between individual headers.
  * Navigation clamps at the list edges without wrapping and handles out-of-range positions safely.
* **Documentation**
  * Updated keybinding documentation and the changelog with the new origin-header navigation.


internal/ui/jump_test.go
```diff
@@ -414,3 +414,61 @@ func TestHeaderJumpCursorOutOfRange(t *testing.T) {
...
+
+	// No origin left in either direction → clamp to the edge, like jumpToHeader.
+	h.jumpToOrigin(1)
+	if want := LastSelectableItem(h.flatItems); h.cursor != want {
+		t.Errorf("ctrl+shift+↓ past the last origin: cursor = %d, want the bottom row (%d)", h.cursor, want)
+	}
+	h.cursor = idxOfOrigin(t, h, "github.com/acme/alpha")
+	h.jumpToOrigin(-1)
+	if want := FirstSelectableItem(h.flatItems); h.cursor != want {
+		t.Errorf("ctrl+shift+↑ above the first origin: cursor = %d, want the top row (%d)", h.cursor, want)
```
hayke102: This assertion is vacuous — it passes whatever the clamp does.

In `headerJumpHome` alpha's origin header **is** index 0, and index 0 isn't a spacer, so `FirstSelectableItem(h.flatItems)` is also 0. Dumped the fixture to be sure:

```
0 origin=true  key=github.com/acme/alpha
1 repo=true    /tmp/hj-a-main
...
idxOfOrigin(alpha)=0  FirstSelectableItem=0
```

So line 441 sets `cursor = 0`, and line 443 compares `0` against `0`. The clamp could return `FirstSelectableItem`, `current`, or a hardcoded `0` and this still goes green — it pins nothing.

The down-direction assertion just above (:438) *is* meaningful: it starts at beta's origin (7) and lands on `LastSelectableItem` (9). The up one needs a fixture with a non-origin row above the first origin to say anything, or should be dropped as untestable here rather than left looking like coverage.

internal/ui/sidebar.go
```diff
@@ -916,6 +916,23 @@ func LastSelectableItem(items []SidebarItem) int {
...
+		return it.IsRepoHeader
+	})
+}
+
+// NextOriginItem is NextHeaderItem narrowed to origin headers: checkout headers
+// are scanned past rather than landed on, so ctrl+shift+↑/↓ moves a whole group
+// at a time however many worktrees sit under it. Everything else — the "own
+// group first" landing, the edge clamp, the out-of-range guard — is the same
+// scan, so the two motions can't disagree about what a stale cursor means.
+func NextOriginItem(items []SidebarItem, current, direction int) int {
```
hayke102, line 936: Worth confirming the inherited edge clamp is what you want at this altitude.

For a user with **one** origin group — the common one-repo case — there is never a next origin, so *every* `Ctrl+Shift+↓` falls through the scan to `LastSelectableItem` and teleports the cursor to a session row deep inside the last worktree. It's not an edge case there, it's the only thing the key ever does.

For `shift+↓` the clamp is a rare outcome that usefully doubles as "go to bottom". For a key whose help text reads "Jump to previous / next origin", landing on a session row is the opposite of the promise. `Ctrl+Shift+↑` doesn't have the symmetric problem — from anywhere below, the first press correctly surfaces the group's own origin header.

A no-op when no further origin exists would read better here. Sharing the clamp with `NextHeaderItem` is the right call for the out-of-range guard and the scan; the *no-match* branch is the one place the two motions arguably shouldn't agree. Currently pinned by the test at jump_test.go:437, so it's a deliberate choice — just flagging that the choice lands differently one level up.

internal/ui/app.go
```diff
@@ -2914,6 +2914,14 @@ func (h *Home) handleKey(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
 		h.jumpToHeader(-1)
 		analytics.Track(analytics.EventHeaderJump, nil)
 		return h, h.fetchPreviewForSelected()
+	case "ctrl+shift+down":
+		h.jumpToOrigin(1)
+		analytics.Track(analytics.EventHeaderJump, nil)
```
hayke102: Both origin cases emit `EventHeaderJump` with `nil` properties — byte-identical to what `shift+↑/↓` sends 5 lines up. The two motions are indistinguishable in telemetry, so after this ships there's no way to answer the one question you'd ask about a new binding: is anyone pressing it?

A property would do it:

```go
analytics.Track(analytics.EventHeaderJump, map[string]any{"level": "origin"})
```

(and `"checkout"` on the existing pair, so the split is readable rather than "origin vs. unlabelled").

internal/ui/jump_test.go
hayke102: Fixed — you're right, it compared 0 against 0. Confirmed by dumping the fixture: alpha's origin header is row 0 and FirstSelectableItem is 0, so the clamp could have returned FirstSelectableItem, current or a hardcoded 0 and it stayed green.

Rather than drop it I split the two claims. That assertion only ever ruled out a *wrap* to the bottom, so it now says exactly that and nothing more. The clamp itself moved to a new TestOriginJumpCursorOutOfRange, mirroring TestHeaderJumpCursorOutOfRange through the new entry point — an out-of-range cursor scanned upward must find the LAST origin (7), which is neither end of the list, so the three candidate answers finally differ.

Mutation-checked: with the clamp removed the new case fails `NextOriginItem(items, 15, -1) = 0, want 7` while the old assertion passed.

internal/ui/sidebar.go
hayke102: Keeping the clamp — moving to the bottom is the wanted answer here too. But the analysis was right that it lands differently one level up, so it's no longer inherited silently: NextOriginItem's doc comment now records that with a single origin group *every* ctrl+shift+down ends in the clamp rather than reaching it occasionally, and that go-to-bottom is chosen on those terms rather than borrowed from NextHeaderItem.

No behaviour change, and the test at jump_test.go:437 still pins it.

internal/ui/app.go
hayke102: Fixed — both motions are labelled now, not just the new one, so the split reads as origin vs. checkout instead of origin vs. unlabelled. Went through a small headerJumpProps(level) helper so the four call sites can't drift into two spellings of the same key.

PR #296, opened 2026-09-08
feat(ui): search and sections in the keybindings sheet

hayke102:
## Why

The `?` sheet holds 51 bindings, and every column was sized to the **single widest binding in the table** (`colW = keyW + 2 + descW` over all of them). One 46-character description therefore made two columns need ~130 terminal columns. Measured on `master`:

| terminal | cols | hidden |
|---|---|---|
| 80×24 | 1 | 37 |
| 100×30 | 1 | 31 |
| 120×40 | 1 | 21 |
| 160×50 | 2 | 0 |

So for anyone not on a 160-column terminal it was a single-column, 51-row scroll list with most of itself off-screen, and `j`/`k` was the only way through. It also threw away the section each binding already knew it was in (`KeyBinding.Section`).

## What

Three coupled changes — they share the layout math, so they ship together.

**Type-to-filter.** Fuzzy over `key + " " + desc` (so `ctrl` finds `Ctrl+T` and `snooze` finds `z`), reusing the palette's `fuzzy.Find` / `highlightMatches` / `filterShiftIndexes`. Arrows, PgUp/PgDn and Home/End scroll; `esc` clears a filter before it closes the sheet, so one press never discards both.

**Sections, with counts.** The four slot bindings move into their own `slots` section — they carried the two longest strings and were setting the width for every other key. Sections **flow across column boundaries rather than pinning them**: packing whole sections into columns would floor the sheet at the height of `session` (23 rows) and make 100×30 *worse* than before. A split section repeats its header as `… (cont.)`, and when the sheet is one column — every terminal under ~120 — the scroll indicator names the section you're in (`⋮ +23 above · sessions`), since `(cont.)` has no boundary to fire at.

**Per-column widths, height-first column count.** Each column is measured against its own contents. The count is the fewest columns that fit both budgets, falling back to the most that fit the width — a width-only search renders a two-hit filter as two one-row columns, which is what the pre-rewrite `colsByHeight` term existed to prevent.

| terminal | before | after |
|---|---|---|
| 80×24 | 1 col, 37 hidden | 1 col, 46 hidden — searchable |
| 100×30 | 1 col, 31 hidden | 1 col, 40 hidden — searchable |
| 120×40 | 1 col, **21 hidden** | **2 cols, nothing hidden** |
| 160×50 | 2 cols, 0 hidden | 2 cols, 0 hidden, narrower box |

Sections add rows, so below ~120 columns the raw hidden count goes **up**. That's the trade: more structure and a filter, against more total rows. At 80×24 the sheet cannot fit under any layout — the `session` block alone is 23 entries — which is exactly where search earns its place.

## Decisions worth reviewing

- **Typing breaks the old "press any key to close" contract**, and with it `j`/`k`/`g`/`G` as scroll keys. Unavoidable once any letter types, and the same trade every text-owning dialog in fleet already makes.
- **Headers drop while filtering; the section folds onto the row instead.** Load-bearing rather than decorative: `` ` ``, `Enter` and `PgUp/PgDn` each appear in two sections meaning different things, so a filtered list without the tag shows one key contradicting itself.
- **`routeToModal` hands this overlay the raw key**, not the `normalizeKey`-remapped one — otherwise a Hebrew user's search box silently types Latin. That moves it out of `TestNormalizedDialogsHoldNoTextInput`'s scan set, so `TestNonLatinTypingReachesTheHelpFilter` guards the other direction. Nothing is lost: every key the sheet still acts on is a named one, which `normalizeKey` returns untouched.
- **Three descriptions were shortened** — they were the direct cause of the width problem, but it is user-visible text: `Bind/unbind slot (re-press same slot clears it)` → `(re-press to clear)`, `Bind slot (fallback if Alt unsupported)` → `(if Alt unsupported)`, and `Snooze session / repo / worktree (again to wake)` → `Snooze / wake session / repo / worktree`. Without that last one, 120 columns misses two columns by **one character** (113 needed against 112 available).
- **Fuzzy, not `strings.Contains`.** Noisier on short queries (`sn` → 17 hits vs 1) but score-ordered, so exact matches lead, and it matches the palette. One line to change if the noise grates in use.

## Tests

12 help tests. The one that matters most is `TestHelpSheetHoldsTwoColumnsAt120`: the 120-column result hangs on ~7 columns of slack (a 105-wide grid against a 112 budget), so a description seven characters longer collapses it back to one column and 30 rows of scrolling, silently. **Verified it fails** on that change, reporting `renders 1 columns, want 2 — a description probably grew`.

Also new: filtered-path overflow at six sizes (a section tag adds up to 17 columns no resting row pays for), filtered results staying in one column, letters-type-arrows-scroll, and the non-Latin routing guard.

`go test ./...`, `golangci-lint` (0 issues) and `make build` all pass.

## Noted, not touched

`HelpBarBindings()` (`keybindings.go`) has zero callers despite its comment claiming the settings dialog uses it. Pre-existing dead code.

https://claude.ai/code/session_014b2bMnXTKHsWnFaE4m2nvM


internal/ui/help.go
```diff
@@ -4,124 +4,307 @@ import (
...
+	var rows []helpRow
+	for _, m := range fuzzy.Find(query, haystacks) {
+		e := all[m.Index]
+		keyLen := runeLen(e.Key)
+		rows = append(rows, helpRow{
+			Key:     e.Key,
+			Desc:    e.Desc,
+			Section: e.Section,
+			Tag:     helpSectionLabel(e.Section),
+			KeyIdx:  filterShiftIndexes(m.MatchedIndexes, 0, keyLen, 0),
```
hayke102: `fuzzy.Find` returns **byte** offsets, but they're split here with `runeLen()`, so every binding whose text contains a non-ASCII rune highlights the wrong characters.

`sahilm/fuzzy@v0.1.2` appends `matchedIndex = j`, where `j` is a byte index into the haystack (`fuzzy.go:170`). The haystack is `Key + " " + Desc`, and `Key` is where the arrows live:

- filtering `group` on `Shift+↑/↓` → `Jump to previous / next group header` highlights `p hea` — off by 4, from two 3-byte arrows;
- filtering `sidebar` on `` ` `` → `Close drawer → sidebar` highlights `debar`, and silently drops two of the seven matched chars because the rune-space `hi` bound clips the tail.

Converting `m.MatchedIndexes` to rune offsets before the split fixes both. `filterShiftIndexes` is shared with `command_palette.go:1020`, which carries the same latent bug on non-ASCII ticket titles — so the conversion probably belongs in the helper rather than at this call site.

internal/ui/help.go
```diff
@@ -4,124 +4,307 @@ import (
...
-	availW := max(colW, h.width-frameH-marginH)
+	// Fewest columns that fit both budgets. Ascending matters: a two-hit filter
+	// laid out as two one-row columns is what a width-only search produces, and
+	// it is what the old code's colsByHeight term prevented. When nothing fits
+	// the height, the loop keeps the widest candidate that fits the width, since
+	// more columns is less scrolling. One column is always allowed — the box has
+	// a MaxWidth net, and refusing to render is not an option.
+	chunks := []helpChunk{newHelpChunk(nil)}
+	best := -1
+	for c := 1; c <= min(helpMaxCols, max(1, len(rows))); c++ {
```
hayke102, line 326: The width budget is gated on `c > 1`, so the one-column layout is never measured against `availW` — and the `MaxWidth` net at the bottom of `View()` then eats the **section tag**, which is the half of the row this PR calls load-bearing.

At 60×24 filtering `s`, the single column is `keyW 14 + 2 + descW 40 + 2 + tag 15 = 73` → box 77 → `MaxWidth(60)` cuts the right border *and* the whole tag column. Rows render as `` ` `` → `Close drawer → sidebar` with no section, which is precisely the "one key contradicting itself" case the tag exists to prevent (`` ` ``, `Enter` and `PgUp/PgDn` each appear in two sections).

Unfiltered, the same terminal fits exactly (column 56, box 60), so this is filter-only, on roughly 60–76 columns. Truncating `Desc` with `truncRunes` when the column exceeds `availW` would drop the part that can be re-read rather than the part that disambiguates.

internal/ui/help_test.go
```diff
@@ -63,15 +64,258 @@ func TestHelpOverlayAllBindingsReachable(t *testing.T) {
...
+			ho.SetSize(s.w, s.h)
+			ho.Show()
+			ho.filter.SetValue(q)
+			view := ho.View()
+			if bh := renderedBoxHeight(view); bh > s.h {
+				t.Errorf("%q at %dx%d: box height %d exceeds %d", q, s.w, s.h, bh, s.h)
+			}
+			for _, ln := range strings.Split(view, "\n") {
+				if w := lipgloss.Width(ln); w > s.w {
+					t.Errorf("%q at %dx%d: line width %d exceeds %d", q, s.w, s.h, w, s.w)
```
hayke102: This test can't fail. `View()` ends with `MaxWidth(h.width).MaxHeight(h.height).Render(box)` and then `lipgloss.Place(h.width, h.height, …)`, so every returned line is ≤ `h.width` and the box ≤ `h.height` whatever `layout()` computed — the clamp is what's being asserted, not the geometry.

It passes at 60×15 while the tag column is being truncated off every row. Asserting pre-clamp is what would catch that:

```go
if w := chunksWidth(lay.chunks, lay.gutter) + DialogStyle.GetHorizontalFrameSize(); w > s.w {
    t.Errorf(...)
}
```

`TestHelpOverlayNeverOverflows` has the same shape — pre-existing, so not this PR's doing, but it means neither overflow test is currently load-bearing.

internal/ui/help.go
```diff
@@ -4,124 +4,307 @@ import (
...
 
+// helpRow is one rendered line of the grid: a binding, a section header, or the
+// blank between two sections.
+type helpRow struct {
+	Key     string
+	Desc    string
+	Section string
+	Tag     string // section label folded onto the row, filtering only
+	Header  bool
+	Spacer  bool
```
hayke102: `Spacer` is never set: `helpRows()`'s at-rest branch emits only headers and bindings, and the filtered branch emits only bindings.

So the leading-spacer strip in `chunkHelpRows`, the `case r.Spacer:` in `newHelpChunk`, the spacer branch in `cell()` and the `Spacer` term in `binding()` are all unreachable, and the `r.Spacer` guard in `help_test.go:98` can never fire. Either drop the field, or emit the blank inter-section rows the surrounding comments describe.

internal/ui/help.go
hayke102: Fixed — confirmed byte offsets in sahilm/fuzzy@v0.1.2 (`matchedIndex = j`, advancing by rune byte size). Reproduced exactly as described: `fuzzy.Find("group", …)` on the 46-rune/50-byte `Shift+↑/↓` haystack returns [38 39 40 41 42], which as rune offsets highlights "p hea".

Added `runeIndexes()` and applied it at both call sites — you were right that `command_palette.go` carried the same latent bug, so it's converted there too rather than only at the help-sheet call site. `TestHelpFilterHighlightsSurviveMultiByteKeys` pins "group" and "sidebar".

internal/ui/help.go
hayke102: Fixed. The `c > 1` gate did leave the one-column layout unmeasured, and `MaxWidth` was cutting the tag — the one element the PR description argued was load-bearing.

Took your suggestion: `fitWidth()` truncates `Desc` with `truncRunes` (floor of 12) when the single column exceeds `availW`, so the part that can be re-read goes and the part that disambiguates stays. Verified at 60×24 filtering `s`: grid+frame is now 58 against a 60 budget, with all 12 rows keeping their section tag. Highlight indexes past the cut are dropped too, or they'd paint off the end.

internal/ui/help_test.go
hayke102: Fixed, and a fair hit — I verified `TestHelpSheetHoldsTwoColumnsAt120` fails when broken and then never applied the same check to the test beside it.

It now asserts pre-clamp with the form you suggested (`chunksWidth + GetHorizontalFrameSize()`). Confirmed load-bearing: with `fitWidth` disabled it fails `box width 79 exceeds 60`, where the old `View()`-based form passed. Height still measures the rendered box, since that clamp isn't what was hiding anything.

You're right that `TestHelpOverlayNeverOverflows` has the same shape — left alone as pre-existing, but worth a follow-up.

internal/ui/help.go
hayke102: Fixed — dropped the field. The blank inter-section rows were removed deliberately (they cost 6 rows, which is the difference between the sheet fitting unscrolled at 120×40 and hiding 4), but I left `Spacer` and its handling behind.

Removed the field, the strip in `chunkHelpRows`, the `case` in `newHelpChunk`, the branch in `cell()`, and the term in `binding()`; the `r.Spacer` guard in the test is now just `r.Header`. Also fixed the three comments that still described the blanks. Layout is unchanged — 120×40 stays 2 columns, nothing hidden.

PR #300, opened 2026-09-09
feat(ui): frost mode

hayke102:
## Summary

Freeze the rendered frame into a styled cell grid — replayed through the `x/vt` emulator so colors, attributes and wide glyphs land where the terminal put them — and run a small cell-level sim over it (`internal/frost`, no Bubble Tea or tmux dependencies). Started from the sidebar by a particular run of keys; ends on resize or `esc`. Ships with `/no-changelog`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist

- [x] Documentation updated (if applicable)
- [ ] Changelog fragment added in `changelog/unreleased/` (if user-facing change; see `changelog/README.md`)

https://claude.ai/code/session_01VrTXgxsLtRGid8ijrxi7nm

## Summary by CodeRabbit

- **New Features**
  - Added a hidden Frost mode that transforms the current screen into an interactive frozen scene.
  - Players can aim, fire, damage the frozen display, navigate debris, and trigger visual effects.
  - Frost mode includes animated rendering, speech bubbles, status information, and a collapse sequence when exiting.
  - The mode preserves the current screen, including overlays, and exits cleanly on completion, quit, or terminal resize.
  - Added a Frost unlock prompt accessible through command palette search.
  - Added analytics tracking for Frost mode sessions.

### DIFF
--- a/internal/analytics/events.go
+++ b/internal/analytics/events.go
@@ -48,10 +48,10 @@ const (
 	// Navigation / filtering.
 	EventFilterUsed = "filter_used"
 	EventSpaceJump  = "space_jump"
+	EventHeaderJump = "header_jump"
 
 	// Frost mode was started from the sidebar.
-	EventFrostMode  = "frost_mode"
-	EventHeaderJump = "header_jump"
+	EventFrostMode = "frost_mode"
 
 	// RTS slot bindings.
 	EventSlotBindingSet = "slot_binding_set"
--- a/internal/analytics/events.go
+++ b/internal/analytics/events.go
@@ -48,6 +48,9 @@ const (
 	// Navigation / filtering.
 	EventFilterUsed = "filter_used"
 	EventSpaceJump  = "space_jump"
+
+	// Frost mode was started from the sidebar.
+	EventFrostMode  = "frost_mode"
 	EventHeaderJump = "header_jump"
 
 	// RTS slot bindings.
--- a/internal/frost/scene.go
+++ b/internal/frost/scene.go
@@ -13,23 +13,28 @@ import (
 // rate below is per tick. Rows are twice as tall as columns are wide, so a
 // vertical velocity is halved to look like the same speed on screen.
 const (
-	probeGravity  = 0.045 // rows per tick²
-	debrisGravity = 0.08
-	driveSpeed    = 0.9 // columns per tick while a drive key is held
-	driveTicks    = 3   // ticks of motion one key event adds (key repeat keeps it topped up)
-	driveMax      = 12  // cap on banked motion, so a burst of presses is not a long slide
-	angleStep     = 5   // degrees per ↑/↓
-	emitCooldown  = 4
-	flashTicks    = 2
-	shakeTicks    = 2
-	burstFrame    = 2 // ticks per impact frame
-	maxDebris     = 600
-	probeSubsteps = 6
-	traceDots     = 40
-	recoilTicks   = 2
-	bubbleTicks   = 40 // how long a line of speech stays up
-	collapseRows  = 3  // rows released per tick when the frame comes down
-	collapseCap   = 80 // ticks before a collapse ends whether or not debris has settled
+	probeGravity   = 0.045 // rows per tick²
+	debrisGravity  = 0.08
+	driveSpeed     = 0.9 // columns per tick while a drive key is held
+	driveTicks     = 3   // ticks of motion one key event adds (key repeat keeps it topped up)
+	driveMax       = 12  // cap on banked motion, so a burst of presses is not a long slide
+	angleStep      = 5   // degrees per ↑/↓
+	emitCooldown   = 4
+	critOdds       = 8  // one emit in critOdds is a crit: twice the footprint, cleared outright
+	critCooldown   = 16 // a crit holds the next shot longer, so there is time to watch it
+	critGap        = 3  // plain shots that must go between one crit and the next
+	flashTicks     = 2
+	shakeTicks     = 2
+	critShake      = 4
+	burstFrame     = 2 // ticks per impact frame
+	maxDebris      = 600
+	probeSubsteps  = 6
+	traceDots      = 40
+	recoilTicks    = 2
+	bubbleTicks    = 40 // how long a line of speech stays up
+	collapseRows   = 3  // rows released per tick when the frame comes down
+	collapseCap    = 80 // ticks before a collapse counts as settled whether or not debris has
+	collapseLinger = 30 // ticks the settled debris stays on screen before the run ends
 )
 
 // Actor is the driveable unit.
@@ -45,13 +50,30 @@ type Actor struct {
 }
 
 // Probe is a projectile in flight. Positions are in cells; y is fractional rows.
-type Probe struct{ x, y, vx, vy float64 }
+// A crit carries its last few positions so it can draw a trail.
+type Probe struct {
+	x, y, vx, vy float64
+	crit         bool
+	past         [3][2]float64
+}
 
-// Burst is an impact animation at a cell.
+// Burst is an impact animation at a cell. A big one is the same animation
+// stamped in a cluster that blooms outward: each outer copy runs a frame or
+// two behind the centre.
 type Burst struct {
 	x, y int
 	age  int
+	big  bool
 }
+
+// bigBurstStamps are a big burst's outer copies around the centre stamp, with
+// the frames each runs behind. Rows are tall, so the cluster is wide.
+var bigBurstStamps = [...]struct{ dx, dy, lag int }{
+	{-3, 0, 1}, {3, 0, 1}, {-2, -1, 1}, {2, -1, 1},
+	{-6, 0, 2}, {6, 0, 2},
+}
+
+const bigBurstLag = 2 // frames the outermost copies run behind
 
 // Debris is a glyph knocked loose and falling.
 type Debris struct {
@@ -68,17 +90,18 @@ type Scene struct {
 	frame    *Grid   // scratch buffer each Render composes into
 	damage   []uint8 // per cell: 0 intact, 1 cracked, 2 gone
 
-	Actor    Actor
-	probes   []Probe
-	bursts   []Burst
-	falling  []Debris
-	heap     [][]Cell // per column, bottom-up: debris that has landed
-	flash    int
-	shake    int
-	tick     int
-	power    float64
-	rng      *rand.Rand
-	showKeys bool
+	Actor     Actor
+	sinceCrit int // shots since the last crit; a crit needs critGap of them
+	probes    []Probe
+	bursts    []Burst
+	falling   []Debris
+	heap      [][]Cell // per column, bottom-up: debris that has landed
+	flash     int
+	shake     int
+	tick      int
+	power     float64
+	rng       *rand.Rand
+	showKeys  bool
 
 	entering bool // driving in from off-screen; input is ignored until parked
 	razed    int  // frame cells cleared by impacts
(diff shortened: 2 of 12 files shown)


internal/frost/world.go
```diff
@@ -0,0 +1,164 @@
...
+// the real terminal would have shown — colors, bold, faint, wide glyphs —
+// lands in the right cell. Lines are joined with CRLF because the emulator
+// treats a bare LF as line-feed only.
+func Freeze(frame string, w, h int) *Grid {
+	g := NewGrid(w, h)
+	if w <= 0 || h <= 0 {
+		return g
+	}
+	emu := vt.NewEmulator(w, h)
+	_, _ = emu.Write([]byte(strings.ReplaceAll(frame, "\n", "\r\n")))
```
hayke102: `Freeze` replays the frame through the emulator with autowrap on and no truncation, so any line wider than `w` (or a frame taller than `h`) wraps, scrolls the emulator and drops row 0 — the frozen picture ends up one row higher than what the user saw, with the header gone.

Repro: `Freeze("row0\nAAAAAAAAAA", 8, 2)` renders `AAAAAAAA` / `AA      ` and `row0` is lost. Reachable in fleet: `renderHelpBar` has no truncation and measures ~93–96 plain columns on a session row / checkout header, and `renderBody` pads up to height but never trims down. Bubble Tea's own renderer clips, so the live view hides it; the replay doesn't.

Suggest `ansi.Truncate` each line to `w` and stop after `h` lines before `emu.Write` (`ansi` is already imported here).

internal/ui/frost.go
```diff
@@ -0,0 +1,78 @@
...
+
+// startFrost freezes what is on screen and starts the sim over it.
+func (h *Home) startFrost() tea.Cmd {
+	if h.width == 0 || !h.booted || h.frost != nil {
+		return nil
+	}
+	debuglog.Logger.Info("frost: start", "size", h.width*h.height)
+	h.actionLog.Add("frost", "start", true)
+	analytics.Track(analytics.EventFrostMode, nil)
+	screen := frost.Freeze(h.renderBody(), h.width, h.height)
```
hayke102: This freezes `h.renderBody()`, but `View()` composites the account-usage readout, the What's New badge, the tip box and toasts on top of `renderBody` *afterwards* (the `overlayAt` calls in `View`). With two accounts configured the readout and badge blink out the moment the run starts and pop back when it ends — and `drawStatus` then writes into the exact corner the readout occupied.

`beginQuit`'s `frozenFrame` takes the same source, but that one is dimmed behind a box so the gap is invisible there. Worth factoring `View`'s pre-chrome composition into one helper and freezing that from both paths.

internal/frost/scene.go
```diff
@@ -0,0 +1,726 @@
...
+		}
+	}
+}
+
+// Key applies one keypress (Bubble Tea's Key.String() spelling). It reports
+// true when the run should end now. The first exit key starts the collapse
+// instead; a second one during it ends the run at once.
+func (s *Scene) Key(k string) (exit bool) {
+	switch k {
+	case "esc", "q", "ctrl+c":
```
hayke102: `ctrl+c` is in the exit set, and `handleKey` hands every key to the run first, so fleet's quit key can't quit while a run is active: the first press starts the collapse, the second aborts it, and only the third reaches `case "ctrl+c": h.beginQuit` — while the frozen help bar still reads `⌃C Quit`.

Either have the app-side intercept treat `ctrl+c` as `endFrost` + `beginQuit`, or drop it from this set and let it fall through.

internal/ui/app.go
```diff
@@ -871,6 +877,9 @@ func (h *Home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		h.width = msg.Width
 		h.height = msg.Height
 		h.sidebarDirty = true
+		if h.frost != nil {
```
hayke102: This ends the run on every `WindowSizeMsg`, including one reporting the same size — and Bubble Tea's `checkResize` sends one on every SIGWINCH without comparing to the previous size. Inside tmux (the supported case) another client attaching or detaching, or a DPI change that only touches the pixel fields, delivers a same-size message and the run vanishes mid-play with no key pressed.

The log line above already gates on an actual dimension change; this needs the same `msg.Width != h.width || msg.Height != h.height` check, computed *before* `h.width`/`h.height` are overwritten two lines up.

internal/ui/app.go
```diff
@@ -2196,6 +2217,9 @@ func (h *Home) View() tea.View {
 	if !h.booted {
 		return h.chrome(RenderSplash(h.width, h.height, h.bootProgress(), h.splashFrame))
 	}
+	if h.frost != nil {
```
hayke102, line 2231: Returning here skips the toast pass, but the run only owns keys — worker and async results still land, and `setError`/`showToast` add toasts with a 5s `ExpiresAt` that `ToastStack.View` prunes on its next call. A run longer than 5s silently discards any error raised during it (a failed workspace creation, a delete error); only `errorHistory` keeps it.

Either overlay the toast stack on the rendered frame, or hold TTLs until first draw.

internal/frost/scene.go
```diff
@@ -0,0 +1,726 @@
...
+	return dots
+}
+
+// Render composes the next frame: the damaged screen, then landed and
+// falling debris, the trace, the actor, probes, bursts and the status line.
+func (s *Scene) Render() string {
+	f := s.frame
+	f.CopyFrom(s.world)
+	for x := 0; x < s.W; x++ {
+		for i, c := range s.heap[x] {
```
hayke102, line 673: The heap pass writes columns left to right with `Grid.Set`. A debris cell carrying a 2-column grapheme keeps its content while `fling` forces `Width = 1`, so `Set` re-upgrades it to width 2 (the `ansi.StringWidth` guard), writes a width-0 spacer at `x+1`, and the very next heap write at `x+1` hits `case 0: g.cells[i-1] = blank` and erases it. Result: a heap cell that is solid but renders blank every frame.

Repro: `heap[5] = {"🔥", Width 1}`, `heap[6] = {"x"}` → the bottom row renders only `x`, while `heapAt(5, …)` is still true. Reachable in play: razing a session title with an emoji lands invisible debris the actor climbs and probes burst against. Same interaction for falling debris drawn side by side.

internal/frost/world.go
```diff
@@ -0,0 +1,164 @@
...
+	if c.Content == "" {
+		c.Content = " "
+	}
+	if c.Width == 1 && ansi.StringWidth(c.Content) == 2 {
+		c.Width = 2 // a caller that guessed wrong would leave the row a column wide
+	}
+	g.cells[i] = c
+	if c.Width == 2 {
+		if x+1 < g.W {
+			g.cells[i+1] = Cell{Width: 0}
```
hayke102, line 119: Writing a width-2 cell unconditionally overwrites `cells[x+1]` with a spacer without checking whether `cells[x+1]` was itself the left half of another wide glyph, so that glyph's spacer at `x+2` is orphaned — `Solid()` reports it as a wall forever and `Render` prints a space.

Repro: `Set(0, 🔥)`, `Set(2, 🔥)`, `Set(1, 🔥)` → `cells[3]` is a spacer with no owner.

Both this and the heap-pass issue are cases `uv.Line.Set` already handles (it walks back to the placeholder's owner), and ultraviolet is promoted to a direct dependency in this PR: `Grid` is close to `uv.Buffer` (`NewBuffer`/`CellAt`/`SetCell`/`Clone`/`Render`), `vt.Emulator` already implements `uv.Screen`, and `Buffer.Render` diffs styles with `Style.Diff` instead of resetting on every change. Delegating would also drop the second direct `x/vt` import, which `internal/vterm`'s package doc says should stay a one-file change.

internal/frost/scene.go
```diff
@@ -0,0 +1,726 @@
...
+		d.vy += debrisGravity
+		d.vx *= 0.9
+		d.x += d.vx
+		d.y += d.vy
+		col := int(math.Round(d.x))
+		if col < 0 || col >= s.W {
+			continue
+		}
+		landing := float64(s.H - 1 - len(s.heap[col]))
+		if d.y >= landing {
```
hayke102: A chip is stacked on whichever column it drifted into, using that column's *current* height, so a chip drifting sideways into a taller neighbouring mound is appended on top of it and jumps several rows up in one frame.

Repro: `heap[10]` six tall, chip at `x=9.4, y=8, vx=0.3` → one `stepDebris()` makes it `heap[10][6]`, drawn at row 3. Routine in play: `fling` gives `vx` up to ~±0.42 with a 0.9 decay, mounds beside the actor are 3+ tall, and `relaxHeap` only moves cells at a ≥2 height difference, so it can't undo the jump before it's drawn.

Only stack once the chip has actually reached the surface (e.g. `d.y >= landing-1`), or resolve against the column it came from.

internal/analytics/events.go
```diff
@@ -48,6 +48,9 @@ const (
 	// Navigation / filtering.
 	EventFilterUsed = "filter_used"
 	EventSpaceJump  = "space_jump"
+
+	// Frost mode was started from the sidebar.
```
hayke102, line 53: Nit: this lands between `EventSpaceJump` and `EventHeaderJump`, so `EventHeaderJump` now sits under the frost comment instead of its `// Navigation / filtering.` group. A separate group after it keeps the existing one intact.

internal/frost/world.go
hayke102: Fixed - Freeze now keeps the first h lines and ansi.Truncates each to w before the replay (e7664a2); TestFreezeClipsAnOversizedFrame pins the row-0 case.

internal/ui/frost.go
hayke102: Fixed - View's pre-chrome composition (body, header overlays, dropdowns, tip/toast stack) is now composeScreen(); the frost freeze and beginQuit's frozenFrame both take it (e7664a2).

internal/frost/scene.go
hayke102: Fixed - ctrl+c is out of the scene's exit set; the app-side intercept ends the run and lets it fall through to beginQuit, so one press quits (e7664a2). TestQuitKeyEndsTheRunInOnePress covers it.

internal/ui/app.go
hayke102: Fixed - the run now ends only on a real dimension change, computed before h.width/h.height are overwritten (e7664a2). TestOnlyARealResizeEndsTheRun sends a same-size message and a real one.

internal/ui/app.go
hayke102: Fixed - the frost branch of View overlays the live toast stack bottom-right on the rendered frame via the shared anchorBottomRight helper (e7664a2). The tip stays suppressed by modalOpen.

internal/frost/scene.go
hayke102: Fixed at the source - fling now sheds a ▪ chip for any glyph wider than one column, since a per-column heap can't carry a 2-column cell (e7664a2). TestWideGlyphShedsAOneColumnChip covers the heap render.

internal/frost/world.go
hayke102: Fixed the orphan - Set blanks the spacer of a width-2 glyph it covers (e7664a2, TestSetBlanksTheSpacerOfTheGlyphItCovers). Left the uv.Buffer swap and the second x/vt import for a follow-up: with the fling change nothing in the game writes a width-2 cell through Set any more, and the swap would reopen Freeze, Solid and the shake shift for a path that no longer has a trigger.

internal/frost/scene.go
hayke102: Fixed - stepDebris remembers the column a chip came from; crossing into a column whose surface is more than a row above it counts as hitting the mound's side, so the chip stays in its own column with vx=0 (e7664a2). TestChipHittingAMoundSideStaysInItsColumn is your repro.

internal/analytics/events.go
hayke102: Fixed - moved into its own group after EventHeaderJump (e7664a2).

PR #302, opened 2026-09-09
feat(ui): sweep a shine across the boot splash wordmark

hayke102:
A narrow vertical glint now travels left→right through the `FLEET` wordmark while fleet boots, brightening the block glyphs in the columns it passes. Spaces pass through uncolored, so the light lands on the letterforms rather than reading as a scanline across the whole row.

```
t=0.00s |.............................................|
t=0.08s |..░▓█▓░......................................|
t=0.16s |........░▒██▓░...............................|
t=0.24s |...............░▒██▒░........................|
t=0.32s |......................░▓██▒░.................|
t=0.40s |.............................░▓█▓▒...........|
t=0.48s |....................................░▓█▓░....|
t=0.56s |..........................................░▒█|
t=0.64s |.............................................|  rest
  ...                                                    (0.52s)
t=1.12s |░............................................|  next pass
```

## Approach

This is the What's New badge's glint applied to the wordmark, not a second dialect — same `hexToRGB`/`lerpRGB`/`rgbToHex` helpers, same "sweep once then rest" shape. Two things differ, both because the target is six stacked rows of solid `█` rather than a line of thin text:

- **One crest, computed per frame in `RenderSplash` and shared by all six rows.** That is what makes it a single vertical bar instead of six independently lit rows.
- **Raised-cosine falloff instead of linear.** Over a band this wide on solid blocks, linear ends on a visible step, and its spiked peak makes brightness jitter as the fractional crest slides between cells. The cosine reaches zero with zero slope, so the glint's outer columns melt into the gradient.

Because the lift is a *fraction* toward white rather than a fixed near-white, each column keeps its hue under the bar — it looks pink over the `F` and violet over the `T`, which is the physically right read for a light crossing a colored surface.

## Two constants are pinned, not free

**`splashShineMax = 0.28` is a hard floor.** Below it the accent end of the gradient downsamples to the *same* xterm-256 index as its own base (`#ff77c6` and its 0.26 lift both land on 212), so the shine disappears outright for anyone not on a truecolor terminal while looking fine to whoever tuned it. Measured against `colorprofile.ANSI256`; the threshold moves with the palette's accent, so re-measure before lowering.

**`splashShineSpeed` and `splashShineWidth` move together.** Their ratio sets how many frames each column spends ramping up and back down; below ~3 the sweep reads as a stepping bar rather than a moving light.

## Why the tick went 80ms → 20ms

That second constraint is the whole reason. At a fixed frame rate, a faster sweep can only be bought by widening the glint — the width *is* the motion blur — until it stops reading as a line and starts reading as a glow. Raising the rate buys speed and narrowness at once: the glint is **5.7 columns** here and was **8** at a third of this speed.

`splashSpinnerDiv` and `splashLabelDwell` divide the counter back down, so the spinner keeps its 80ms cadence and the label its ~1s dwell — only the shine spends the extra frames. The shutdown overlay has its own tick and is untouched.

**50fps was measured, not assumed:** `RenderSplash` benchmarks at ~284µs, so the splash burns ~1.4% of one core for the one to three seconds it is up. The drawer slide already animates at 45fps.

`splashShineRest` is also picked to keep the cycle off-harmonic with the other two animations on this screen — 55.8 frames against the spinner's 40 and the label's 48. Landing near 1.00 would fire the sweep on the same spinner glyph every pass, which reads as a mechanism rather than a flourish.

## FreezeAnim

No new guard needed, and it gets a free win: `FLEET_FREEZE_ANIM` already pins `frame` at 0, where the crest sits exactly `splashShineWidth` from column 0 and the falloff evaluates to zero. Captures stay byte-identical to the un-shined wordmark, so `demo/drive.sh` doesn't churn.

## Tests

`splash.go` had none. The load-bearing addition is **`TestFleetWordmarkColumnsAlign`** — the bar is only vertical because every row is 45 single-width runes, and redrawing the ASCII art would skew it silently with nothing else failing. It checks rune count *and* display width, since a double-width rune would leave layout correct while shearing the shine.

Plus `TestSplashShineCrestSweepsThenRests` (the phase machine: advances, carries past the final column before resting, loops) and `TestRenderSplashShineIsColorOnly` (the shine may change color and never layout — a splash that reflows every frame is the most visible jitter fleet can ship).

https://claude.ai/code/session_019TyDiTSYXo2Q65sH7t3WDh

## Summary by CodeRabbit

- **New Features**
  - Added a sweeping shine effect across the FLEET wordmark during boot startup.
  - Improved splash animation smoothness with faster frame updates while preserving spinner and status-label timing.

- **Documentation**
  - Updated demo and splash-screen documentation to reflect the revised animation timing and capture behavior.

- **Tests**
  - Added coverage for wordmark alignment, shine animation progression, looping, and consistent rendered output.

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -3202,9 +3202,17 @@ func (h *Home) handleKey(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
 	return h, nil
 }
 
-// shutdownTick schedules the next shutdown-overlay spinner advance (~80ms,
-// matching splashTick). Self-rescheduled by Update while `quitting`, until the
-// teardown command emits tea.Quit and the program exits.
+// shutdownTick schedules the next shutdown-overlay spinner advance (~80ms).
+//
+// That is splashTick's interval times splashSpinnerDiv, and the equality is the
+// point rather than a coincidence: the splash runs its tick fast for the shine
+// and divides the counter back down for its spinner, so both spinners advance
+// at the same wall-clock rate off different tick rates. Move either constant
+// and the other has to follow, or the two spinners visibly disagree on a
+// screen the user sees fleet start and stop with.
+//
+// Self-rescheduled by Update while `quitting`, until the teardown command emits
+// tea.Quit and the program exits.
 func (h *Home) shutdownTick() tea.Cmd {
 	if FreezeAnim {
 		return nil
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -6274,12 +6274,27 @@ func (h *Home) bootProgress() float64 {
 	return float64(resolved) / float64(h.bootstrapRepos)
 }
 
-// splashTick schedules the next splash-spinner advance (~80ms cadence).
+// splashTick schedules the next splash frame (~20ms cadence).
+//
+// Four times the rate of the shutdown overlay's otherwise-identical tick, and
+// that asymmetry is the point: the shine sweeping the wordmark is the only
+// thing here that moves continuously, and frame rate is the only currency that
+// buys it speed without cost. At a fixed rate a faster sweep must jump further
+// per frame, which is paid for by widening the glint until it reads as a glow
+// rather than a line — so each speed-up came as a rate bump instead, and the
+// bar got narrower each time rather than broader. The spinner and the label are
+// held to their original cadence by dividing this counter down at the call site
+// (see RenderSplash), so only the shine spends the extra frames.
+//
+// 50fps is affordable here and was measured, not assumed: RenderSplash costs
+// ~284us, so the splash burns ~1.4% of one core while it is up — for one to
+// three seconds, on a screen with nothing else competing for the loop. The
+// drawer slide already animates at 45fps.
 func (h *Home) splashTick() tea.Cmd {
 	if FreezeAnim {
 		return nil
 	}
-	return tea.Tick(80*time.Millisecond, func(t time.Time) tea.Msg {
+	return tea.Tick(20*time.Millisecond, func(t time.Time) tea.Msg {
 		return splashFrameMsg(t)
 	})
 }
--- a/internal/ui/splash_test.go
+++ b/internal/ui/splash_test.go
@@ -57,21 +57,26 @@ func TestSplashShineCrestSweepsThenRests(t *testing.T) {
 }
 
 // The shine may change color and nothing else — a splash that reflows every
-// 80ms is the most visible jitter fleet can ship.
+// 20ms is the most visible jitter fleet can ship.
 func TestRenderSplashShineIsColorOnly(t *testing.T) {
-	// Frames 1 and 3 fall inside one spinner window (splashSpinnerDiv frames
+	// These two frames fall inside one spinner window (splashSpinnerDiv frames
 	// per glyph) and one label bucket (splashLabelDwell), so the shine is the
-	// only thing that can differ. Both are mid-sweep, so both are lit.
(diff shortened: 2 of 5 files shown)


internal/ui/app.go
```diff
@@ -6274,12 +6274,27 @@ func (h *Home) bootProgress() float64 {
 	return float64(resolved) / float64(h.bootstrapRepos)
 }
 
-// splashTick schedules the next splash-spinner advance (~80ms cadence).
+// splashTick schedules the next splash frame (~20ms cadence).
```
hayke102, line 6285: The 80ms → 20ms change strands two comments that still assert the old cadence:

- `shutdownTick` (app.go:3205) reads "~80ms, matching splashTick". It no longer matches, and the relationship it was pointing at survived the change in a different form — the invariant is now `shutdown 80ms == splashTick × splashSpinnerDiv`. Worth stating, since anyone moving one tick will read that comment to decide whether the other has to follow.
- `demo/drive.sh:374` justifies the 120ms `wait_stable` poll by listing fleet's cadences as "100ms (spinners), 80ms (splash) and 60ms (shimmer)" and noting that 120ms "shares no small-integer ratio with any of them". 120ms is now exactly 6× the splash tick. Not reachable today — `wait_stable` only runs after `wait_booted`, by which point the splash is gone — but during the shine's ~520ms rest window two consecutive captures would be byte-identical, so the stated reason for choosing that constant is no longer true.

internal/ui/splash_test.go
```diff
@@ -0,0 +1,77 @@
...
+	b := RenderSplash(80, 24, 0.5, 3)
+
+	if ansi.Strip(a) != ansi.Strip(b) {
+		t.Error("the shine changed the splash's layout; it must only change color")
+	}
+	if lipgloss.Width(a) != lipgloss.Width(b) {
+		t.Errorf("splash width moved between frames: %d vs %d", lipgloss.Width(a), lipgloss.Width(b))
+	}
+	if a == b {
+		t.Error("frames 0 and 10 render identically; the shine is not animating")
```
hayke102: The message names frames 0 and 10, but the test renders 1 and 3 (lines 65-66) — on a regression this sends the reader to inspect frames that were never compared.

Line 60 above is stale in the same direction: "a splash that reflows every 80ms" is 20ms now.

internal/ui/app.go
hayke102: Fixed both, in 2cfe18b.

`shutdownTick` now states the invariant that survived rather than the one that died: 80ms is `splashTick × splashSpinnerDiv`, so both spinners advance at the same wall-clock rate off different tick rates, and moving either constant means moving the other.

`drive.sh` — corrected, with one thing your note didn't have: `fleet_env` exports `FLEET_FREEZE_ANIM=1` unconditionally (line 70), so nothing animates under the harness at all and no cadence aliases against the sampler. The comment now says that, then keeps the 100ms-vs-120ms reasoning as the hedge it actually is, and states the 6× overlap plus why it's unreachable. Fixing only the number would have left a reader checking the arithmetic, finding a false claim, and having nothing explain why it doesn't bite.

internal/ui/splash_test.go
hayke102: Fixed in 2cfe18b, and made it undriftable rather than just correct.

You're right that it's the worst place for staleness to survive — it's the one line a person sees when the test fails. It was also stale twice over: `0, 10` was the pair from two revisions ago, not the `0, 20` it replaced, so it had already drifted by hand once.

So the frames are named constants now and all three messages print them with `%d` instead of restating them. Line 60's 80ms → 20ms too.

PR #306, opened 2026-09-10
feat(site): Google Analytics + working light/dark toggle

adiva6:
Two site-only changes.

## Google Analytics

Loads `gtag.js` on the landing page and every docs page from the root layout. Uses `next/script` with the `afterInteractive` strategy rather than raw script tags, which is what the static export needs to emit them correctly.

**It is gated to production.** The measurement id comes from `NEXT_PUBLIC_GA_ID`, set only by the deploy workflow's build step, alongside the `NEXT_BASE_PATH` already set there. Unset, the layout renders no scripts at all, so `npm run dev` and local production builds send nothing.

The site is a static export with no server, so this is a build-time value that gets inlined into the bundle. It is a variable rather than a secret because a measurement id is visible in the page source either way. The cost is that a local production build can no longer smoke-test the tag without setting the variable by hand.

## The light/dark toggle now actually switches

The toggle rendered and flipped the class on `<html>`, but nothing responded to it, so light mode was dead.

Three things were wrong:

- The `--charm-*` palette was declared only under `:root`, with dark values.
- `html` was pinned to `color-scheme: dark` with a hardcoded background.
- Several components carried literal dark-only colors that no theme could reach, such as `rgba(244,143,177,…)` glows and `#1a0a14` text on pink fills.

The fix splits the palette into a light `:root` block and a dark `.dark` block. The `--color-fd-*` Fumadocs mapping below them stays written once, since every value is a token that now switches on its own.

New token pairs cover what used to be hardcoded: accent glows, the pink tint, the coach banner's translucent slab, and the page background washes. Plus `--charm-on-pink`, the text color that sits on a pink fill, which is near-black in dark and white in light because the light-mode pink is dark enough to need it.

The theme provider also gets `enableSystem: true`. The toggle already offered a System button, but next-themes was told not to handle it, so that third option did nothing. Dark stays the default for a first visit.

The TUI demo stays dark in both modes on purpose, since a terminal does not turn white. Only its coaching banner, which sits outside the terminal frame, follows the theme.

`site/CLAUDE.md` records the rule that caused this, which is that every `--charm-*` token is declared twice and colors never get hardcoded in a component.

## Verification

Clean builds both ways: with `NEXT_PUBLIC_GA_ID` unset the output contains zero references to the tag, and with it set the tag lands in the landing page and every docs page.

I rendered the landing page and the install doc in both themes with headless Chrome. Dark is unchanged, light renders correctly throughout, including feature cards, install tabs, docs sidebar, and code blocks.

Site-only, so `/no-changelog`.

https://claude.ai/code/session_017EPGRBWngTa3Sgb1pc884K

## Summary by CodeRabbit

- **New Features**
  - Added Google Analytics tracking to the deployed site.
  - Added support for following system light and dark theme preferences.

- **Enhancements**
  - Updated the site with coordinated light and dark Charm themes.
  - Extended theme consistency across gradients, buttons, cards, navigation, selection states, landing-page effects, and the TUI demo.
  - Theme changes now apply more broadly across site components and visual effects.

- **Documentation**
  - Expanded guidance for theme tokens, component colors, and the TUI demo’s intentional dark-mode styling.

### DIFF
--- a/site/CLAUDE.md
+++ b/site/CLAUDE.md
@@ -24,6 +24,12 @@ Raw `<a href="/docs/...">` does NOT get the `/fleet` basePath prepended. Only `n
 
 `app/globals.css` defines `--charm-*` and `--tn-*` palette tokens and overrides every `--color-fd-*` Fumadocs token. To restyle docs, edit the token overrides, not the Fumadocs components.
 
+**Every `--charm-*` token is declared twice** — light under `:root`, dark under `.dark` (the class next-themes puts on `<html>`). Source order matters: `:root` and `.dark` have equal specificity, so the dark block must stay *after* the light one. The `--color-fd-*` mapping below them is written once and mode-agnostic, because each value is a `--charm-*` token that already switches.
+
+**Never hardcode a color in a component.** The theme toggle was dead for exactly this reason: the palette lived only in `:root`, so flipping the class changed nothing, and the landing components carried literal `rgba(244,143,177,…)` glows and `#1a0a14` on-pink text that no theme could reach. If you need a new color, add a token pair — that includes glows (`--charm-glow`, `--charm-glow-soft`), tints (`--charm-tint`), translucent slabs (`--charm-panel`, `--charm-panel-ring`), and the page washes (`--charm-wash-1..3`). `--charm-on-pink` is the text color that sits *on* a pink fill; it is near-black in dark and white in light, because the light-mode pink is dark enough to need it.
+
+The **TUI demo is the deliberate exception**: `components/tui-demo/palette.ts` is hardcoded tokyo-night and stays dark in both modes — a terminal doesn't turn white. Only demo chrome *outside* the terminal frame (`CoachBanner`'s panel and inline `kbd`s) reads `--charm-*`.
+
 ## TUI demo (`components/tui-demo/`)
 
 Pure-React recreation of the fleet sidebar+preview. Reads as one feature; touch it carefully.
--- a/site/app/globals.css
+++ b/site/app/globals.css
@@ -7,9 +7,56 @@
 /* ─── Charm palette ────────────────────────────────────── */
 
 @layer base {
+  /* Light is the `:root` default; `.dark` (set by next-themes on <html>)
+     overrides it. Every surface below reads these tokens and nothing else,
+     which is what makes the theme toggle actually switch the site. */
   :root {
+    color-scheme: light;
+
+    --charm-pink: #c93571;
+    --charm-pink-bright: #e85d9a;
+    --charm-pink-hover: #ad2a60;
+    --charm-on-pink: #ffffff;
+    --charm-purple: #6d3ad6;
+    --charm-lime: #2f8f24;
+    --charm-bg: #fbf9fb;
+    --charm-bg-soft: #f4f1f6;
+    --charm-surface: #ffffff;
+    --charm-surface-elev: #f4eff6;
+    --charm-border: #e3dce7;
+    --charm-border-soft: #ede8f0;
+    --charm-text: #191820;
+    --charm-text-dim: #5a5665;
+    --charm-text-faint: #8b8697;
+
+    /* Accent glows and translucent slabs, per-mode so they never sit
+       dark-on-dark in light or wash out in dark. */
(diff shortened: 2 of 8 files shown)


site/CLAUDE.md
```diff
@@ -24,6 +24,12 @@ Raw `<a href="/docs/...">` does NOT get the `/fleet` basePath prepended. Only `n
...
+
+**Never hardcode a color in a component.** The theme toggle was dead for exactly this reason: the palette lived only in `:root`, so flipping the class changed nothing, and the landing components carried literal `rgba(244,143,177,…)` glows and `#1a0a14` on-pink text that no theme could reach. If you need a new color, add a token pair — that includes glows (`--charm-glow`, `--charm-glow-soft`), tints (`--charm-tint`), translucent slabs (`--charm-panel`, `--charm-panel-ring`), and the page washes (`--charm-wash-1..3`). `--charm-on-pink` is the text color that sits *on* a pink fill; it is near-black in dark and white in light, because the light-mode pink is dark enough to need it.
+
+The **TUI demo is the deliberate exception**: `components/tui-demo/palette.ts` is hardcoded tokyo-night and stays dark in both modes — a terminal doesn't turn white. Only demo chrome *outside* the terminal frame (`CoachBanner`'s panel and inline `kbd`s) reads `--charm-*`.
```
hayke102: Small nit, totally optional: `TuiDemo.tsx` still has a few literal pink glows that sit outside the terminal frame, like the focus aura (~L278) and the card's outer shadow (~L307–308). They probably look fine on light, but by this rule they'd read `--charm-glow` / `--charm-wash-*`. Happy for it to be a follow-up.

site/components/landing/CursorAura.tsx
```diff
@@ -48,7 +48,7 @@ export function CursorAura() {
         pointerEvents: "none",
         mixBlendMode: "screen",
```
hayke102: Minor: I think `screen` can only lighten, so over the near-white light background this aura probably ends up invisible, and the new wash tokens don't do much here. Maybe switch the blend mode per theme (e.g. `normal`/`multiply` in light)? Not a blocker, just flagging.

hayke102 (dismissed):
Looks great, thanks for tracking down why the toggle was dead. Two small optional nits inline, nothing blocking.

site/app/globals.css
```diff
@@ -7,9 +7,56 @@
...
+    --tn-red: #c33f57;
+    --tn-gray: #6b7194;
+    --tn-orange: #b45f1e;
+    --tn-purple: #7a52c9;
+  }
+
+  .dark {
+    color-scheme: dark;
+
     --charm-pink: #f48fb1;
```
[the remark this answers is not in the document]
hayke102: `.dark` is only ever set on `<html>` by next-themes, and nothing in the site uses a nested `dark` class, so the bare selector is safe here. Resolving.

site/app/layout.tsx
```diff
@@ -16,11 +19,27 @@ export const metadata: Metadata = {
...
+          src={`https://www.googletagmanager.com/gtag/js?id=${GA_MEASUREMENT_ID}`}
+          strategy="afterInteractive"
+        />
+        <Script id="google-analytics" strategy="afterInteractive">
+          {`window.dataLayer = window.dataLayer || [];
+function gtag(){dataLayer.push(arguments);}
+gtag('js', new Date());
+
+gtag('config', '${GA_MEASUREMENT_ID}');`}
+        </Script>
```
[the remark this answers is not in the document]
hayke102: GA4 enhanced measurement tracks history-based page changes by default ("Page changes based on browser history events"), so `next/link` navigations are already recorded as page views without a manual hook. Resolving.

site/CLAUDE.md
hayke102: Fixed in 83fb01a: the focus aura and outer card glow now mix from `--charm-pink`/`--charm-purple` at the same alphas, so dark renders identically.

site/components/landing/CursorAura.tsx
hayke102: Fixed in 83fb01a: the blend mode now reads `--charm-aura-blend` (`multiply` in light, `screen` in dark).

PR #307, opened 2026-09-10
feat(ui): support the mouse

hayke102:
Click a sidebar row to select it, double-click to activate it, wheel to scroll, and click an entry in the context menu, command palette, snooze picker, account picker or allowed-accounts editor to run it. `FLEET_NO_MOUSE` opts out, matching `FLEET_NO_COPY_COMMAND` and `FLEET_NO_EXTENDED_KEYS`.

### Why this was off, and what changed

Reporting was deliberately disabled (#268) and **the reason has not gone away**: v2's `eventLoop` calls `View()` after *every* message and there is no click-only mouse mode, so one trackpad flick is ~300 `MouseWheelMsg`/sec, each buying a full compose — measured at 220% CPU.

Two things hold that off, and both live in `Update`/`View` rather than in any handler, so a future mouse handler cannot forget them:

- a mouse message is the **only** kind that does not mark `viewDirty`, so an event nothing acted on is served the cached frame (`Home.screen`);
- one that *does* act repaints at most once per `mouseRepaintInterval` (16ms), with a trailing `mouseSettleMsg` so the settled state still lands.

### The decisions worth reviewing

**The wheel scrolls `viewOffset` and never the cursor.** Every cursor move fires `fetchPreviewForSelected`, which forks a `tmux capture-pane`, so a wheel bound to j/k would fork hundreds of processes per flick. Scroll to look, click to select.

**Geometry is recorded at render, never re-derived.** `renderBody` writes the sidebar rect per layout branch (all three put content at `sidebarContentTop`, the constant `contextMenuAnchor` inverts) and `composeScreen` writes the overlay rect where the box's x/y are already known. Row maps are built by each dialog's own `View` as it emits rows, because the palette interleaves section headers, a `recent` label and blank spacers with its results — nothing but that loop knows which line a row got. `sidebarWindow` is now shared with `RenderSidebar` for the same reason: the `… N more below` indicator sits exactly on the first index a naive `viewOffset + row` returns.

**Activation is the dialog's own key**, returned by `ClickRowAt` and replayed through `routeToModal`: enter in a menu, **space** in the allowed-accounts editor (where enter *saves*, so a menu-style click would store whatever happened to be ticked), and nothing on the snooze dialog's custom-duration row (enter there means "snooze for what I typed" and the box is empty when you click into it). A click outside an overlay closes it via its own Esc and never falls through to the sidebar behind, which is still painted because the dropdowns deliberately don't dim.

**Enter's body moves to `activateCursorRow`** so a double-click means literally what the key means, rather than a second copy that forgets the split-mode or suspended-session branch — the copy in `handleSlotJump` already misses the first of them.

**Clicks are inert while the drawer is typing or a split is focused.** `j`/`k` cannot move the sidebar there either, and a click that did would point the highlight at one session while the keys go to another. The wheel stays live, since scrolling is looking.

### Trade-off

Mouse reporting takes drag-to-select away from the terminal, so copying text out of fleet's own window now needs `Shift` (Option in iTerm2). Selection *inside* an attached agent pane is unaffected — that's a full-screen tmux client with its own `mouse on` and `copy-command`.

### Test guard replaced

`TestChromeKeepsMouseReportingOff` is deleted rather than updated: its premise was that nothing in the TUI handles a mouse event. The invariant it protected is now covered from the other side — an ignored mouse event must leave the frame clean while every other message dirties it. `TestAttachRestoresTerminalModes` stays, with its rationale corrected: Bubble Tea only *sets* the modes the view asks for on resume, so tmux's leftover 1000 and 1003 still need clearing.

### Verified

Driven in the sandbox (`demo/drive.sh`), not just compiled — clicks in dual/stacked/single, scroll-indicator rows inert, wheel scrolling without cursor drift, double-click attach, mouse reporting surviving an attach round-trip (`button=1 sgr=1` on return), context-menu and palette clicks, click-away dismiss, and a ticket row under an interleaved section header prefilling exactly that ticket. `make test` (race) green, `golangci-lint` 0 issues.

https://claude.ai/code/session_0173jM9NGPE8DPLtZhismTq9

## Summary by CodeRabbit

* **New Features**
  * Added mouse support for sidebar selection, double-click activation, scrolling, and overlay interactions.
  * Added mouse controls for account pickers, context menus, command palettes, snooze dialogs, and permission selection.
  * Added an opt-out setting through `FLEET_NO_MOUSE`.
  * Improved pointer handling for clicks, drags, releases, focus changes, and unsupported areas.

* **Performance**
  * Reduced unnecessary screen refreshes during mouse interaction for smoother performance.


internal/ui/snooze_dialog.go
```diff
@@ -187,6 +189,16 @@ func (d *SnoozeDialog) View() string {
...
+	// the third line of the box and the presets start two below it. The input is
+	// the row after the presets and a blank, and carries the same highlight, so
+	// it is a click target like any other row.
+	d.clickRows.reset(2)
+	for i := range SnoozeDurations {
+		d.clickRows.set(2+i, i)
+	}
+	d.clickRows.set(2+len(SnoozeDurations)+1, snoozeInputRow())
+
 	b.WriteString(TitleStyle.Render(ansi.Truncate(d.title, boxW, "…")))
```
hayke102: **The title is truncated to `boxW`, but the box only has `boxW-6` content columns** — so a title over ~34 columns wraps to two lines and shifts every preset one row below where `clickRows` says it is.

That was cosmetic before this PR. It isn't now: `clickRows` above assumes a one-line title, and the label is `"Snooze " + session.Title`, where `naming.GenerateTitle` cuts at ~50 runes. So the wrap is the common case, not the edge one.

Rendered at the default width with `Show("Snooze refactor the status detection pipeline")`:

```
│  Snooze refactor the status detection  │   line 2
│  pipe…                                 │   line 3   ← the wrap
│                                        │
│    30 minutes        17:05             │   line 5   map says 4
│    1 hour            17:35             │   line 6   map says 5
│    4 hours           20:35             │   line 7   map says 6
│    Tomorrow      Fri 09:00             │   line 8   map says 7
```

Clicking **30 minutes snoozes for 1 hour**. Clicking `Tomorrow` (line 8) and the custom-duration row (line 10) fall off the end of the map and are silently swallowed.

`account_picker.go:156` and `allowed_accounts.go` both truncate to `inner := max(boxW-chrome, 1)` and are correct — this is the only one of the five that measures against the outer width.

internal/ui/command_palette.go
```diff
@@ -619,6 +627,9 @@ func (d *CommandPaletteDialog) View() string {
 			if leadCol > 0 {
 				lead = renderPriorityLead(it.Priority)
 			}
+			// Everything written so far ends in a newline, so the row lands on
+			// the line whose index is the count of them.
+			d.clickRows.set(strings.Count(b.String(), "\n"), i)
```
hayke102: **This counts newlines *written*, not lines *rendered*.** `boxed()` ends with `DialogStyle.Width(d.dialogWidth()).Render(...)`, and Lip Gloss re-wraps anything wider than that — so once the tab bar or the search input wraps, every recorded row is one or more lines above where it is actually drawn.

The row bodies are budgeted against `innerContentWidth()`, but `renderTabs()` and `"  " + d.filterInput.View()` are not. Measured by rendering the real dialog and locating each row in the output:

| terminal | dialogWidth | inner | drawn | recorded |
|---|---|---|---|---|
| 65 | 57 | 51 | 8 | 8 ✓ |
| 64 | 56 | 50 | 9 | 8 ✗ |
| 50 | 42 | 36 | 10 | 8 ✗ |
| 40 | 32 | 26 | 11 | 8 ✗ |

At 64 columns the tab bar (`all 1   actions 1   repos/worktrees 0   tickets 0`) takes a second line, so **clicking a palette entry fires the entry below it** — the wrong command, the wrong repo, the wrong ticket — and the bottom row maps past `len(d.filtered)` and is swallowed. The threshold moves further up as the counts grow past one digit.

64 columns is a normal split pane, and the palette is otherwise usable there. The four simple dropdowns are unaffected because their content is truncated to the inner width; the palette is the only one that writes content the box can re-wrap.

Cheapest fix that keeps the "recorded as they are emitted" property: count the lines of the wrapped chrome once (`lipgloss.Height` of `renderTabs()` and of the input) and carry that as an offset, or record rows against `lipgloss.Height(b.String())` rather than a raw newline count. There's also an O(n²) in here as it stands — `strings.Count(b.String(), "\n")` rescans the whole buffer once per row, on every keystroke.

internal/ui/mouse.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func (m overlayRowMap) at(dy int) (int, bool) {
+	i, ok := m.rows[dy]
+	return i, ok
+}
+
+// handleMouse routes a mouse message. It returns without marking the frame
+// dirty unless something actually changed, so an ignored event is free.
+func (h *Home) handleMouse(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
```
hayke102, line 140: **No frost guard here, and `handleKey` has one.** `handleKey` opens with `if h.frost != nil { ... return h, nil }` and swallows every key for the run; `handleMouse` doesn't, and `View()` returns `h.frost.Render()` without going through `screen()` — so `h.layout.sidebar` keeps the live geometry that `startFrost`'s own `composeScreen()` call wrote a moment earlier.

Reproduced:

```go
h.frost = frost.New(frost.Freeze("x", h.width, h.height), 1)
h.Update(tea.KeyPressMsg{Code: 'j', Text: "j"})              // cursor 0 — swallowed ✓
h.Update(tea.MouseClickMsg{X: 5, Y: sidebarContentTop + 4})  // cursor 4 ✗
h.Update(tea.MouseWheelMsg{X: 5, Y: sidebarContentTop + 2})  // viewOffset 1 ✗
```

So during a run: a click moves the cursor and forks a `capture-pane` behind the picture, and **two clicks inside `mouseDoubleClickWindow` run `activateCursorRow()`** — a `tea.Exec` attach, or a resume of a suspended session, launched from a screen showing an animation with no repaint in between to say what was picked.

`h.frost != nil` is already in `modalOpen()`, so the guard is one line at the top of `handleMouse`.

internal/ui/mouse.go
```diff
@@ -0,0 +1,337 @@
...
+	// and so does the pointer, rather than parking the cursor somewhere no key
+	// could have put it.
+	if h.flatItems[idx].IsSpacer {
+		return h, nil
+	}
+
+	doubleClick := idx == h.lastClickRow && time.Since(h.lastClickAt) < mouseDoubleClickWindow
+	moved := h.cursor != idx
+	h.cursor = idx
+	h.syncViewport()
```
hayke102: **`syncViewport` anchors on `sidebarMinVisibleRows()` (contentHeight−4) while `h.layout.sidebar.h` is contentHeight−2**, so clicking one of the last two painted rows of a scrolled list scrolls it out from under the pointer — and the double-click can never land.

Pure-geometry sweep (click a cell, run `syncViewport`, re-resolve the same cell):

```
termH=24 (rect h=20, minVisible=18) n=50 off=30 row=19:
  clicked item 48, same cell now resolves to 49 (off→31)
termH=24 ...            n=50 off=31 row=19:
  clicked item 49, same cell now resolves to -1  (off→32)
```

So the second click of a double-click isn't one: it selects the **neighbouring session** and forks another `capture-pane`, and a third click then attaches the wrong session. One offset further along it resolves to the `… N more below` row and does nothing at all.

It only bites when the list is scrolled to within a row or two of the end — which is exactly where the newest session sits, and where you'd double-click to attach to it. Either re-clamp against the rect the frame was drawn from rather than the conservative bound, or skip `syncViewport` on a click, since the click already proves the row is on screen.

internal/ui/mouse.go
```diff
@@ -0,0 +1,337 @@
...
+		return h.handleClick(m)
+	}
+	// Release and motion carry nothing fleet acts on. Drag lands here too:
+	// cell-motion reports it, and swallowing it is what stops a drag over the
+	// sidebar from walking the cursor.
+	return h, nil
+}
+
+// handleWheel scrolls whatever the pointer is over.
+func (h *Home) handleWheel(m tea.Mouse) (tea.Model, tea.Cmd) {
```
hayke102, line 167: **The wheel is live in focus mode but invisible there.** `handleClick` returns early on `h.drawerHasFocus() || h.focusMode`; the wheel deliberately doesn't, and the reasoning ("scrolling is looking") is right — except that in dual layout, which is the default, `renderBody` reuses `h.cachedSidebar` whenever `h.focusMode && !h.sidebarDirty && h.cachedSidebar != ""`. `scrollSidebar` sets neither.

So with `enter_mode: split`: Enter to focus a session, wheel over the sidebar — `viewOffset` moves, `markMouseRepaint` buys a frame, and the panel is served from the cache, so nothing on screen changes. Press `esc` and the sidebar jumps to wherever the wheel left it.

`h.sidebarDirty = true` in `scrollSidebar` is the smaller of the two fixes and keeps the stated intent; gating the wheel like the click is the other.

internal/ui/mouse.go
```diff
@@ -0,0 +1,337 @@
...
+		return h, nil
+	}
+	// A spacer is a blank row between origin groups; the keyboard skips over it
+	// and so does the pointer, rather than parking the cursor somewhere no key
+	// could have put it.
+	if h.flatItems[idx].IsSpacer {
+		return h, nil
+	}
+
+	doubleClick := idx == h.lastClickRow && time.Since(h.lastClickAt) < mouseDoubleClickWindow
```
hayke102: Minor, and the same hazard the comments elsewhere in this PR cite: **`lastClickRow` is a raw `flatItems` index, not a row identity.**

`rebuildFlatItems` runs on the ~2s tick and from `adoptSessionsMsg`/`handleSessionCreate`, so a session appearing above the clicked row inside the 400ms window makes index 5 a different session — and the second click then reads as a double-click on a row that was only ever clicked once, and attaches it.

This is what `contextMenuTarget` + `targetForCursor()` were introduced for on the context menu (`TestContextMenuDispatchFollowsTargetNotCursor`), and the latch could carry the same `target` instead of an int. Low probability, but the failure is an attach nobody asked for.

internal/ui/sidebar.go
```diff
@@ -315,6 +315,49 @@ func CollectGroupInfo(sessions []*session.Session, repoPath string) RepoGroupInf
...
+// viewOffset inside an inner content area `height` rows tall.
+func sidebarWindow(itemCount, viewOffset, height int) sidebarWindowSpec {
+	visibleHeight := height
+	if visibleHeight < 1 {
+		visibleHeight = 1
+	}
+
+	w := sidebarWindowSpec{Start: viewOffset}
+	w.Above = viewOffset > 0
+	w.Below = (viewOffset + visibleHeight) < itemCount
```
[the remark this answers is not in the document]
hayke102: Fixed in 513fe51 — `Below` is now decided after the `Above` row is reserved.

Confirmed the case you gave: 11 items at offset 1 in 10 rows drew items 1..9 and neither drew nor announced item 10. The ordering predates this PR (`sidebarWindow` was extracted verbatim from `RenderSidebar`), but you're right that the wheel is what makes it reachable — `viewOffset` used to be written only by `syncViewport`, which never parks there.

`TestSidebarWindowAnnouncesTheLastItem` pins your case plus offset 0 and 2 (both unchanged), and asserts the general invariant: every item is either drawn or announced.

internal/ui/snooze_dialog.go
hayke102: Fixed in 513fe51 — the title is truncated to `inner` now, with the same `snoozeChrome = 6` the sibling dropdowns use.

You're right that this stopped being cosmetic: `clickRows` assumes a one-line title, and `"Snooze " + sc.label` is the common case, not the edge one. Verified in the sandbox with a renamed session — the title now truncates to one line and clicking `4 hours` writes a deadline of 21:13 against a 17:14 clock, i.e. the row it points at.

`TestSnoozeRowMapSurvivesALongTitle` renders with a 56-char title and asserts every preset's drawn line maps to its own index.

internal/ui/command_palette.go
hayke102: Fixed in 513fe51. The chrome (title → the blank after the search input) is built into its own builder and measured at `innerContentWidth()`, and the rows carry a counter from there instead of rescanning the buffer — so the O(n²) goes with it.

One correction to the suggested fix: `lipgloss.Height(renderTabs())` is 1 by construction, since it measures the string as written rather than as wrapped. It has to be rendered at the content width first. And the measurement needs a `-1`: the chrome ends in a newline, which opens the next line rather than occupying one, while `Height` counts the empty segment after it.

`TestPaletteRowMapMatchesTheDrawnRowsAtEveryWidth` sweeps 40/50/64/65/80/120, locating each row in the rendered box and asserting `ClickRowAt` on that line resolves to it. It reproduces your table exactly on the old code (65 ✓, 64 and below ✗). Also checked live at 64 columns with the tab bar visibly wrapped: clicking `My Tickets` opens the tickets tab.

internal/ui/mouse.go
hayke102: Fixed in 513fe51 — `if h.frost != nil { return h, nil }` at the top of `handleMouse`, with the reason on it (View() returns the frost frame without passing through `screen()`, so `h.layout` keeps live geometry under the picture).

`TestMouseIsInertDuringAFrostRun` reproduces your snippet and fails without it: cursor 4 and viewOffset 1 during a run.

internal/ui/mouse.go
hayke102: Fixed in 513fe51 — `syncViewport` is gone from `handleClick`, which was the second of the two fixes you named. A click is proof the row is on screen, and the index came out of `sidebarWindow`, so there is nothing to bring into view and nothing to clamp.

`TestClickedRowStaysUnderThePointer` runs your sweep over offsets 28–32 and reproduces your numbers on the old code: `offset 30 row 19: clicked item 48, the same cell now resolves to 49`.

internal/ui/mouse.go
hayke102: Fixed in 513fe51 with `h.sidebarDirty = true` in `scrollSidebar` — the smaller of the two you named, and the one that keeps the stated intent rather than deleting it.

`TestWheelInFocusModeRepaintsTheSidebar` asserts the scroll both moves `viewOffset` and marks the sidebar dirty; without the line it reports "the scroll left the sidebar cache standing — the frame would not change".

internal/ui/mouse.go
hayke102: Fixed in 513fe51 — the latch is a `contextMenuTarget` now, compared by identity, which also drops the `-1` sentinel and its constructor seed.

Agreed on the framing: low probability, but the failure mode is an attach nobody asked for, and it is the same hazard `targetForCursor()` already exists for. `TestDoubleClickFollowsTheRowNotTheIndex` prepends a session between the two clicks and asserts the second one does not activate.

PR #320, opened 2026-09-17
feat(ui): P jumps to the next PR that needs attention

hayke102:
## What

`P` (next to `p`, open PR) moves the cursor to the next worktree whose PR badge is **red** — CI failed, changes requested, unresolved threads, or conflicts. When nothing on screen is red, it goes to the next **green** one (approved + CI passed). It cycles in on-screen order and wraps; with neither on screen it does nothing. Also in `Ctrl+K` as "Jump to Next PR Needing Attention".

## How

- **One classifier.** The red/green rules move out of `prBadgeStyle` into `prVerdict`, which both the badge colour and the jump read — so the jump can never land on a badge painted a different colour than the one it seeks.
- **Candidates are checkout headers on screen.** A collapsed origin hides its headers and is muted, the same rule `Space` applies to sessions; a collapsed checkout still shows its header, so its PR stays reachable.
- **Snooze is respected** — a checkout snoozed by its own deadline or its origin's is skipped. Drafts, merged and closed PRs are never targets.

## Why not a Space chord

- `Ctrl+Space` is macOS's "previous input source" shortcut on any Mac with two keyboard layouts, so the terminal never receives it.
- `Shift+Space` arrives as a plain space — measured with a Bubble Tea v2 key probe in Ghostty outside tmux, and through tmux — so it would silently fire `Space`'s session jump instead.
- `Alt+Space` types U+00A0 unless the terminal sends Option as Meta.

Like every uppercase command, `P` is unreachable while a caseless layout (Hebrew/Arabic) is active; `Ctrl+K` is the escape there.

## Help sheet fix

Adding one row made the `?` sheet at 120×40 hide **three** bindings: on overflow it added two `⋮` indicator rows on top of the blank separators it already had. The indicators now ride those separator rows, so overflowing by k rows hides exactly k and the frame doesn't jump between fitting and scrolling. `TestHelpSheetHoldsTwoColumnsAt120` pins both budgets (whole at 120×41, exactly one row over at 120×40).

## Tests

- `TestPRJumpRedBeforeGreen` — red before green, drafts ignored, snoozed checkout skipped until woken, wraps onto the only red badge.
- `TestPRJumpMutedByCollapsedOrigin` — a collapsed origin is a no-op.
- `TestPRJumpKeyBinding` — `P` routes through the fully-wired `handleKey`.
- `go test -race ./...` passes; checked in the `demo/drive.sh` sandbox that `P` reaches the jump and the help filter shows the row under `p`.

https://claude.ai/code/session_015GLu5pPTb2Naw5ftHY7uc9

## Summary by CodeRabbit

- **New Features**
  - Added a `P` shortcut and command-palette action to jump to the next pull request needing attention.
  - Prioritizes pull requests with failing checks, requested changes, unresolved threads, or conflicts, then falls back to approved pull requests with passing checks.
  - Skips drafts and snoozed items and supports wraparound navigation.

- **Improvements**
  - Help content now uses existing spacing for scroll indicators, allowing more content to remain visible.

- **Documentation**
  - Updated keybinding and help documentation for the new navigation behavior.


internal/ui/app.go
```diff
@@ -5931,6 +5936,57 @@ func (h *Home) buildJumpTree() []SidebarItem {
...
+//
+// Red and green are prVerdict's own answers, the same ones prBadgeStyle paints,
+// so the jump can never land on a badge a different colour than the one it
+// claims to seek. Candidates are the checkout headers already on screen: a
+// collapsed origin hides its headers, which mutes it exactly as Space mutes its
+// sessions, while a collapsed checkout still shows its header, so its PR stays
+// reachable without expanding anything. Snoozed checkouts — by their own
+// deadline or their origin's — are skipped, since snooze is the attention mute
+// and this is an attention surface.
+func (h *Home) jumpToNextAttentionPR() {
```
hayke102, line 5957: **`P` still jumps when PR badges are turned off.**

With Settings → Appearance → PR badges off, `sidebar.go:601`/`:631` draw no badge, but nothing here checks `ShowPRBadges`. `P` (and the palette's "Jump to Next PR Needing Attention", which calls this too) still moves the cursor onto red/green checkouts. The cursor lands on a plain branch row with nothing on screen explaining why it stopped there, which breaks the doc comment's promise that the jump only lands on a badge of the colour it seeks.

Simplest fix is an early return, matching the "silent no-op when nothing qualifies" contract:

```suggestion
func (h *Home) jumpToNextAttentionPR() {
	if !ShowPRBadges {
		return
	}
```

(Alternative: dim the palette entry with a `PR badges off` reason so the command doesn't read as a dead key.)

hayke102: Fixed — `jumpToNextAttentionPR` now returns early when `ShowPRBadges` is off, which covers both `P` and the palette entry. Added `TestPRJumpNoopWithPRBadgesOff`. Didn't take the dim-the-palette-entry alternative: palette items have no disabled/reason state, so it would mean new plumbing for one command.

